PR #83, opened 2016-10-03
Archive upload: image upload fields updated.

kjaskiewiczz:
Device type, checksum and yocto id fields removed.
This information will be extracted from an archive by the server.

 We need to wait with merging this until https://github.com/mendersoftware/deployments/pull/148 will be merged.

@marekswiecznik @michaelatmender


michaelatmender:
*compile

PR #155, opened 2017-01-05
Added id to login button

olehermanse:
@michaelatmender what do you think?

Context: I'm writing selenium tests and it's sometimes messy to find these buttons.


olehermanse:
Also: Possible to add ids to the header buttons?

PR #198, opened 2017-02-06
Remove build files and let travis build those.

pasinskim:


maciejmrowiec:
@mendersoftware/rndity can you double check that one?

PR #260, opened 2017-08-30
Fix docker image tagging based on commits.

maciejmrowiec:
Changelog: none

Signed-off-by: Maciej Mrowiec <mrowiec.maciej@gmail.com>


maciejmrowiec:
@GregorioDiStefano @michaelatmender that should deal with red builds in github.

PR #293, opened 2017-11-15
only compress js file if DEMO flag is not true

GregorioDiStefano:
Changelog: none
Signed-off-by: Gregorio Di Stefano <greg.distefano+github@gmail.com>


GregorioDiStefano:
@michaelatmender @maciejmrowiec @mendersoftware/rndity 

when testing integration (and not running production environment) it takes almost 20 seconds before the main.js file is compressed.

this change will only compress it when debug != true

maciejmrowiec:
👍

PR #317, opened 2018-02-16
Help section adjustments for hosted Mender

michaelatmender:


src/js/components/help/connecting-devices/beagleboneblack.js
```diff
@@ -89,10 +96,26 @@ var BeagleBoneBlack =  createReactClass({
 
         <p> If you cannot see any new device after 10 minutes, please verify that the network connection is working, and feel free to reach out for help with diagnostics via <a onClick={this.props.changePage.bind(null, "help/more-help-resources")}>our further Help resources</a>.</p>
 
+        <p> NOTE: If you reimage the device with the disk image or switch storage (e.g. SD cards) between the devices after they have booted once, <b>authenticating the device to the Mender server will fail</b>. This is because the Mender server tracks the identity (MAC address by default) device public key (randomly generated upon first run of the Mender client) binding. If this happens, you need to *decommission* the device from the Mender server and try again. </p>
```
estenberg: Re "tracks the identity (MAC address by default) device public key .. binding". It means the binding between those two (identity and public key), that's why I had <->.

We could rephrase to "This is because the Mender server tracks the binding between the device's identity (MAC address by default) and public key (randomly generated upon first run of the Mender client)."

Also note this note is in one more place in this commit.

PR #337, opened 2018-04-19
simple update to allow use of "friendly_name" field

krazik:
@estenberg 

I can't seem to get gulp/etc working local so I haven't been able to test this, but this is a simple change that would allow an inventory field named "friendly_name" to be used in addition to the mongo ID in the UI.

Changelog: Title


krazik:
Changelog: Title

michaelatmender:
We won't merge this now, as there is already some work in progress to address this feature in a slightly different way - it will allow users to select either an identity attribute or an inventory attribute as a "friendly name" in the UI place of the ID. This will be stored as a setting for the user. Thanks for the input though, @krazik.

PR #408, opened 2019-01-17
Chore: code linting + increased consistency + dependency updates

mzedel:
This is unreasonably big for several reasons:

- the node_modules are gone (as they were deprecated and are rebuilt during the automated tests anyway)
- callbacks should now almost all be promises to gain a better way to structure execution ordering
- there now is a lint step in the regular watch gulp pipeline, that should notify about problems/ inconsistencies within the code - the lint rules are given in the `.eslintrc`
- due to the deprecation warnings I switched to es6 classes for all components (to get rid of the createReactClass) and bumped several packages, including react and consecutively react router - this should mostly work, but please take extra care during review and testing


michaelatmender:
@mzedel I found a few issues when testing this PR:

- Switching tabs is really slow (in Devices & Settings views)
- Deployments page is slow on first loading the page (but then switching between tabs is fast)
- Loading from https://localhost/ui/: dashboard doesn't load and console shows `browser.js:49 Warning: You are attempting to use a basename on a page whose URL path does not begin with the basename. Expected path "/ui/" to begin with "/ui/#"`
- The current selected item on the left navigation (either dashboard, devices, artifacts, deployments) is no longer highlighted
- Drag and drop artifact box styling looks different and drag & drop no longer works

mzedel:
@michaelatmender I just fixed the sign-off, once this is merged I will proceed to merge this back into the material-ui branch

PR #416, opened 2019-02-19
Men 2276 multiple payloads per artifact

mzedel:
this depends on mendersoftware/deployments#328 to be merged and should not be deployed earlier than said PR


mzedel:
yes that was what I was thinking, the `ExpansionPanel`s in the newer material-ui version allow this to be a implemented a lot easier and if it is not urgently needed I would like to postpone this. especially the nested `Collapse`s are a bit hairy right now.

michaelatmender:
@mzedel that makes sense, agree it can wait until then. The only other thing was to put the "remove artifact" button under the artifact contents like in the wireframes [here](https://projects.invisionapp.com/share/HCPYJSM7YJ3#/screens/340735152), that's just a small thing though

mzedel:
@michaelatmender ah sorry I missed that - should be fixed now

PR #443, opened 2019-03-22
long device inventory texts are no longer cut off + visible on hover

mzedel:
Changelog: Title

### DIFF
--- a/src/js/components/devices/expandable-device-attribute.js
+++ b/src/js/components/devices/expandable-device-attribute.js
@@ -9,19 +9,41 @@ export default class ExpandableDeviceAttribute extends React.Component {
   constructor(props, context) {
     super(props, context);
     this.state = {
-      expanded: false
+      expanded: false,
+      overflowActive: false
     };
+  }
+
+  componentDidMount() {
+    const self = this;
+    if (self.textContent && self.textContainer) {
+      const overflowActive = self.textContent.offsetWidth > self.textContainer.clientWidth;
+      self.setState({ overflowActive });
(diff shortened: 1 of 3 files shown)


michaelatmender:
@mzedel, I had already created a task for this before I realised you were working on it already: https://tracker.mender.io/browse/MEN-2415
I think it would be improved a little with allowing to click to expand/hide as described in the ticket, do you think it could be adjusted easily?

mzedel:
yep, I'll change that - the expand/ hide click would definitely be better and should not take long to adjust

michaelatmender:
Testing locally, I can't see the' behaviour proposed the ticket. It shows the full value all the time, and there is no "show more" or "show less" link.

mzedel:
Now I added the "show less" and "show more" links and they are only visible if it would not be possible to show the whole text in one line. However, this depends on the window size at the time of mounting the component. So if you expand the attribute list and change the window size, the link will not be directly visible only once you close & re-expand the list. If the window size remains unchanged everything should work as expected.
For values that span multiple lines, we would need a fixed number of lines to force a cutoff, detect the cutoff and make it expandable. I think it might be valuable to show the full value if it is not cutoff and only provide the expansion as a workaround for the cutoff case. What do you think @michaelatmender ?

michaelatmender:
Previously it showed 2 lines and ellipsis "..." if the value was too long even for 2 lines. I think we keep that the same but add the "show more" link after the ellipsis. I still have problems with this PR, will describe in the ticket.

PR #457, opened 2019-04-23
Men 2451 device list

mzedel:
also fixes MEN-2366 and MEN-2381, however the condition to show the `updated_ts` only from the inventory api is so far only kept in the authorised/ group list, since the change to the auth data (from my understanding this should reflect when a device was rejected or dismissed) would be of interest


michaelatmender:
Found a couple of issues with this:
* Pending devices: table checkboxes don’t work
* Preauthorized devices: link and button don’t work

PR #474, opened 2019-05-29
Include frequently asked questions in device limit notification.

estenberg:
I did not test how this would look (e.g. line breaks, length of box, etc.).
Also please verify that the ?_target=_blank approach is valid here. :)


michaelatmender:
It was failing a js linter test and the `target=_blank` part wasn't quite right, so I adjusted it. I edited the  order of the content just a little too while I was there!

estenberg:
Thanks!

PR #480, opened 2019-06-04
Fix mac support in onboarding

mzedel:
this adjusts the download links to mender-artifact as well as the steps to create a customized artifact to support going through them on a mac as well

### DIFF
--- a/src/js/components/artifacts/releaserepository.js
+++ b/src/js/components/artifacts/releaserepository.js
@@ -94,6 +94,13 @@ export default class ReleaseRepository extends React.Component {
       });
   }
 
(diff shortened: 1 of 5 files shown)


michaelatmender:
I tested quickly - the link to mender-artifact doesn't work for me. I get access denied to the url https://d1b0l86ne08fsf.cloudfront.net/mender-artifact/3.0.0/darwin/mender-artifact

mzedel:
ah yes - the other link relies on master... I talked to @kacf and it should be ok to point to master given the new help structure and onboarding that would align with the current state of mender-artifact (and we don't have a better way of aligning the releases otherwise)

PR #485, opened 2019-06-20
Qa 75 production gitlab pipeline

mzedel:


.gitlab-ci.yml
```diff
@@ -0,0 +1,83 @@
...
+stages:
+  - build
+  - test
+  - publish
+  - integration
+
+before_script:
+  - docker info
+  - git clone git://github.com/mendersoftware/mendertesting
+  - . mendertesting/check_commits.sh
```
lluiscampos: We should execute the script rather than source it. Sourcing it might have strange side effects

lluiscampos: I meant: `./mendertesting/check_commits.sh`

mzedel: thanks for catching that - I guess I meant to write `./mendertesting/check_commits.sh` - it's now `mendertesting/check_commits.sh` as you proposed + moved to a separate step

mzedel:
thanks for looking into this! I added the acceptance tests and made some adjustments to the commit checking...

PR #486, opened 2019-06-25
Men 2571 onboarding improvements

mzedel:


mzedel:
@lluiscampos please have a look at this to see if this solves the onboarding issues on linux - I tried it in a vm - so far only with a virtual device - but there the onboarding seemed to work

lluiscampos:
@mzedel I can create JIRA tickets for them instead of handling it here. Your call.

michaelatmender:
Going through this using the virtual device, looks good! 
Only big issue:  
- link for webserver is _localhost:85_ but I needed to navigate to _3000_ to see the page. Known issue/part of another repo though, correct?

Smaller issues:
- don't show the “download demo artifact” tip if artifact already exists
- final snackbar after the first successful deployment is too small, disappears too quickly
- the 'waiting for device' issue also mentioned by @lluiscampos

PR #488, opened 2019-07-24
fixed a problem that could cause a crash of the device lists

mzedel:
this was due to the device inventory being empty and thus a resulting null device in the device list

Changelog: Title


mzedel:
Thanks for looking into it. What browser and which browser version did you test this on? the only real change should be the call to `.finally()`, which I fear might be the reason - but group creation shouldn't be different then...

michaelatmender:
> Thanks for looking into it. What browser and which browser version did you test this on? 

Chrome, Version 75.0.3770.100 64 bit. I tested again on master and the issue wasn't there, so it seems to be this PR

mzedel:
that should now be working properly again 😊

michaelatmender:
> that should now be working properly again 😊

Looks good!

PR #492, opened 2019-08-06
disabled broken e2e tests for now as they are preventing image updates

mzedel:
Changelog: None


mzedel:
that PR should be obsolete as soon as MEN-2498 is closed

PR #499, opened 2019-08-09
Allow integration e2e tests

mzedel:


.travis.yml
```diff
@@ -33,13 +33,13 @@ jobs:
...
 
-    # - stage: Tests
-    #   name: 'acceptance'
-    #   install: skip
-    #   script:
-    #     - git clone https://github.com/mendersoftware/integration.git ;
-    #     - cp -r integration/tests/* tests
-    #     - GUI_REPOSITORY=$(pwd) INTEGRATION_PATH=$(pwd)/integration ./tests/e2e_tests/run
+    - stage: Tests
+      name: 'acceptance'
```
lluiscampos: Two comments:
- I think we should call in integration
- They should be run also from main Mender QA repo, right? So that they run on changes for other backend components and integration repo itself. --> If this is much work, we can create a JIRA ticket to do it later.

mzedel: Yep, I completely agree.
My goal would be to unify the current way the e2e tests are set up to use the same approach the rest does and just call a filtered run here. I looked into the overall setup, but since this is already just piggy backed on the GUI test introduction I would push this into another JIRA ticket: https://tracker.mender.io/browse/MEN-2704.

lluiscampos: Agreed. Thanks

PR #500, opened 2019-08-12
Fix commit compliance checking

lluiscampos:
It was broken since we modified check_commits.sh few weeks ago

Changelog: None

Signed-off-by: Lluis Campos <lluis.campos@northern.tech>


lluiscampos:
The commits check works now. The failure in acceptance test job is unrelated. Merging

PR #501, opened 2019-08-12
MEN-2611 Material ui 4.1

mzedel:
also closes: MEN-2629/ MEN-2694, MEN-2630, MEN-1945 😉

requires `npm install`

this replaces the existing pagination elements with the ones that come with material ui now, I will happily remove the pagesize select/ first/ last page buttons - left them in just because to align with the mui example
this also aligns the widths and scrolling behaviour of the sidebar/content views in groups/devicelist, help/helpcontent and releaselist/release (mainly to fix MEN-2630 and remain consistent)


michaelatmender:
Hey, just a couple of small issues, both related to overflows I think:

- Dropdowns from autoselects look a bit off, because they have overflow:scroll instead of overflow:auto
- Deployment results report table looks off, when status is over multiple lines

I can share screenshots if needed

PR #504, opened 2019-08-14
Men 1521 gzip gui content

mzedel:
to see the difference the image has to be rebuilt, the compression leads to a considerable reduction in the transmitted size (~75%)


0lmi:
Great improvement! Already thought about it many times and happy that it's finally in place.

PR #516, opened 2019-08-22
moved to use more environment details to detect onboarding progress

mzedel:
previously onboarding progress could be lost after refreshing
if progress was not persisted to localstorage at that moment

Changelog: None


mzedel:
device type and onboarding approach (virtual device or physical onboarding) are however only determined by the first device in the device list

PR #529, opened 2019-09-04
onboarding: Don't require Enter after copy/paste operation.

kacf:
Especially since the other blocks do not require it.

Changelog: Title


kacf:
@michaelatmender: The only error is clearly some timing related unit test which I don't think is related to this change. Ok to merge?

kacf:
Cherry-picked to 2.1.x as well.

PR #537, opened 2019-09-16
MEN-2760 report error on login when there is no netwok connection

merlin-northern:
ChangeLog:title

### DIFF
--- a/src/js/api/users-api.js
+++ b/src/js/api/users-api.js
@@ -3,6 +3,8 @@ require('superagent-auth-bearer')(request);
 var Promise = require('es6-promise').Promise;
 import cookie from 'react-cookie';
 import { unauthorizedRedirect } from '../auth';
+import AppActions from '../actions/app-actions';
+import { preformatWithRequestID } from '../helpers';
 
 request.use(unauthorizedRedirect);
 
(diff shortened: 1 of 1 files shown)


src/js/api/users-api.js
```diff
@@ -34,6 +36,14 @@ const Api = {
         .set('Content-Type', 'application/json')
         .send(body)
         .end((err, res) => {
+          if (err && !res) {
+            var errorResponse2 = {
+              text: err.response ? JSON.parse(err.response.text) : err,
+              code: err.status
+            };
+            AppActions.setSnackbar(preformatWithRequestID(err.res, err.message), null, 'Copy to clipboard');
```
mzedel: The api/ non-view layer of the application should be kept free from GUI side effects to keep them a bit more stable than the ever changing components/ view layer...

src/js/api/users-api.js
```diff
@@ -34,6 +36,14 @@ const Api = {
         .set('Content-Type', 'application/json')
         .send(body)
         .end((err, res) => {
+          if (err && !res) {
+            var errorResponse2 = {
```
mzedel: that duplicates what happens in the other case `(err || !res.ok)` - is there any particular reason not to adjust the existing condition that I might be missing?

merlin-northern: > that duplicates what happens in the other case `(err || !res.ok)` - is there any particular reason not to adjust the existing condition that I might be missing?

I am sure you are not missing anything.

merlin-northern:
> Thanks for the prompt contribution! Now that the sprint is over I finally got around to do a review, just to be complete... Since we have 2 PRs from around almost the same time I suggest closing this one and move forward with #538, but if you would like to see this contribution make it to production you will have to adjust your current implementation. In any case: I sure hope to see more of your excursions to the JS land in the future.

thanks!
feel free to close this one. I am sorry I did not know you already had the fix.

mzedel:
closing for #538

PR #545, opened 2019-09-23
forced tenant token retrieval to dialog lifetime instead of creation

mzedel:
Changelog: None


merlin-northern:
looks good to me :) (just leaving comment so it does not look like "without review")

PR #548, opened 2019-09-26
enabled push to dockerhub from gitlab, deprecated travis

mzedel:
Changelog: None


.gitlab-ci.yml
```diff
@@ -92,8 +92,9 @@ publish:
   script:
     - docker load -i image.tar
     - docker tag $SERVICE_IMAGE $DOCKER_REPOSITORY:$COMMIT_TAG
```
lluiscampos: Do we need to tag it also as master regardless of `$COMMIT_TAG` value?

lluiscampos:
> We need to modify the build badge at README.md to point to https://gitlab.com/Northern.tech/Mender/mender-qa/badges/master/pipeline.svg

Wrong link, of course. Here: https://gitlab.com/Northern.tech/Mender/gui/badges/master/pipeline.svg

.gitlab-ci.yml
mzedel: good point! I added both branch (should be master or <tag>) and latest docker tags

PR #549, opened 2019-10-01
Fix Pipeline publish stage: use only $SERVICE_IMAGE

lluiscampos:
It seems like even though in build stage we tag with $SERVICE_IMAGE and
$DOCKER_REPOSITORY:$CI_COMMIT_REF_SLUG, only the former we save/load.


lluiscampos:
@mzedel I think this shoud fix the Pipeline error. See latest master run: https://gitlab.com/Northern.tech/Mender/gui/-/jobs/307916086

PR #550, opened 2019-10-01
aligned docker image tagging and publishing

mzedel:
moved pr tag deletion to testing stage

Changelog: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -71,6 +71,7 @@ test:acceptance:
     - git clone https://github.com/mendersoftware/integration.git
     - cp -r integration/tests/* tests
     - GUI_REPOSITORY=$(pwd) INTEGRATION_PATH=$(pwd)/integration ./tests/e2e_tests/run
+    - docker rmi $DOCKER_REPOSITORY:pr
   tags:
     - mender-qa-slave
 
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -92,10 +93,9 @@ publish:
     - docker load -i image.tar
     - docker tag $SERVICE_IMAGE $DOCKER_REPOSITORY:$COMMIT_TAG
     - docker tag $SERVICE_IMAGE $DOCKER_REPOSITORY:$CI_COMMIT_REF_SLUG
-    - docker tag $SERVICE_IMAGE $DOCKER_REPOSITORY:latest
-    - docker rmi $SERVICE_IMAGE
     - echo -n $DOCKER_HUB_PASSWORD | docker login -u $DOCKER_HUB_USERNAME --password-stdin
     - docker push $DOCKER_REPOSITORY:$COMMIT_TAG
+    - docker push $SERVICE_IMAGE
```
lluiscampos: I don't understand.

Will this command push `$DOCKER_REPOSITORY:$CI_COMMIT_REF_SLUG`?
- If yes, it will also be pushing `$DOCKER_REPOSITORY:$COMMIT_TAG` and the previous line can be removed.
- If no, we should add it here or remove the tagging

mzedel: in the overall `before_script` is:
```
DOCKER_REPOSITORY="mendersoftware/gui"
DOCKER_TAG=${CI_COMMIT_REF_SLUG:-master}
SERVICE_IMAGE=$DOCKER_REPOSITORY:$DOCKER_TAG
COMMIT_TAG="$CI_COMMIT_REF_SLUG"_"$CI_COMMIT_SHA"
```
so this should lead to `mendersoftware/gui:master` and `mendersoftware/gui:master_<hashValue>` as the tagged values. Both are currently present in the docker hub repo and since the push refers to specific tags, both pushes seem to be necessary.
So yes, `docker push $SERVICE_IMAGE` will push `$DOCKER_REPOSITORY:$CI_COMMIT_REF_SLUG` (unless `CI_COMMIT_REF_SLUG` is not defined, then it will default to `$DOCKER_REPOSITORY:master`), but no: it will not push `$DOCKER_REPOSITORY:$COMMIT_TAG` automatically (see https://gitlab.com/Northern.tech/Mender/gui/-/jobs/308552903 & https://hub.docker.com/r/mendersoftware/gui/tags where master has not been updated).
Does this clarify the steps?

lluiscampos: Ok. I think I understood now how this "tags" work in Docker. Thanks

PR #552, opened 2019-10-01
reused ci pipeline variables to skip need for git in image build

mzedel:
Changelog: None

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,17 @@
 FROM node:11.6.0-alpine AS build
 WORKDIR /usr/src/app
-RUN apk add --no-cache git
 COPY package-lock.json package.json ./
 RUN npm ci
 COPY . ./
 RUN npm run build
 RUN npm run disclaim
-RUN echo "$(git rev-parse --abbrev-ref HEAD)_$(git rev-parse --short HEAD)" >> ./dist/version
 
 FROM nginx:1.17-alpine
 RUN mkdir -p /var/www/mender-gui/dist
 WORKDIR /var/www/mender-gui/dist
+ARG GIT_REF
+ARG GIT_COMMIT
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,16 +1,17 @@
...
 RUN npm run build
 RUN npm run disclaim
-RUN echo "$(git rev-parse --abbrev-ref HEAD)_$(git rev-parse --short HEAD)" >> ./dist/version
 
 FROM nginx:1.17-alpine
 RUN mkdir -p /var/www/mender-gui/dist
 WORKDIR /var/www/mender-gui/dist
+ARG GIT_REF
+ARG GIT_COMMIT
+ENV GIT_REF="${GIT_REF:-local}" GIT_COMMIT="${GIT_COMMIT:-local}"
```
lluiscampos: I think it is better to query git for the real version, as relying on an env variable here can lead to the caller cheating.

Can we do something like the following?
```
ENV GIT_REF="$(git rev-parse --abbrev-ref HEAD)_$(git rev-parse --short HEAD)"
```

This way we don't rely anymore on this `version` file that seems to be problematic, but still keep the git tracking of the build.

mzedel: My intention was to get the git usage (see https://github.com/mendersoftware/gui/pull/552/files#diff-3254677a7917c6c01f55212f86c57fbfL9) out of the image building process, since we can get it injected straight from the ci system (the version file worked fine so far). If someone wants to cheat on this information there are still quite a few ways to manipulate the included version information.

lluiscampos: I understand. Is just that I don't see much of a benefit of removing git in a builder image.

But it looks good to me, I won't insist.

mzedel: yes it's definitely not an urgent change - saves just a small bit of build time and I honestly forgot about the existence of the CI variables when I included the version information injection

PR #554, opened 2019-10-02
New deploy wizard dialog and functionality for phased deployments

michaelatmender:
Should also address comments in https://github.com/mendersoftware/gui/pull/553


src/js/components/deployments/deployment-wizard/phasesettings.js
```diff
@@ -0,0 +1,191 @@
...
+  updateBatchSize(value, index) {
+    let newPhases = this.props.phases;
+
+    if (value < 1) {
+      value = 1;
+    } else if (value >100) {
+      value = 100;
+    }
+
+    newPhases[index].batch_size = value;
```
mzedel: this has to be `newPhases[index].batch_size = Number(value);` otherwise the `batch_size` will end up as a string and fail to be parsed in the backend

src/js/components/deployments/deployments.js
```diff
@@ -276,41 +301,28 @@ export default class Deployments extends React.Component {
...
-    this.setState({
-      reportDialog: false,
-      artifact: null,
-      release: null,
-      releaseArtifacts: null,
-      group: null
-    });
-  }
-
   _retryDeployment(deployment, devices) {
```
mzedel: when I merged the current master into this, the `_retryDeployment` changes were overwritten and I'm not sure why - just to keep it in mind...

src/js/components/deployments/deployment-wizard/review.js
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+
+import pluralize from 'pluralize';
+
+import { Chip, List } from '@material-ui/core';
+import { InfoOutlined as InfoOutlinedIcon } from '@material-ui/icons';
+
+import ExpandableDeviceAttribute from '../../devices/expandable-device-attribute';
+import { getRemainderPercent } from '../../../helpers';
+
```
mzedel: to get the dates to further align with the design you could add this: `const dateOptions = { year: 'numeric', month: 'numeric', day: 'numeric', hour: 'numeric', minute: 'numeric' };` here...

src/js/components/deployments/deployment-wizard/review.js
```diff
@@ -0,0 +1,77 @@
...
+const Review = props => {
+  const { deploymentDeviceIds, device, group, isEnterprise, phases, release } = props;
+
+  // Create 'phases' for view only
+  var deploymentPhases = phases ? phases : [{batch_size: 100}];
+  const start_time = deploymentPhases[0].start_ts || new Date().toISOString();
```
mzedel: ...and replace the `toISOString()` with `.toLocaleString(undefined, dateOptions)` here

src/js/components/deployments/deployment-wizard/review.js
```diff
@@ -0,0 +1,77 @@
...
+          </div>
+          {deploymentPhases.map((row, index) => {
+            row.batch_size = row.batch_size || getRemainderPercent(deploymentPhases);
+            const deviceCount = (index === deploymentPhases.length-1) 
+              ? Math.ceil((deploymentDeviceIds.length / 100) * row.batch_size)
+              : Math.floor((deploymentDeviceIds.length / 100) * row.batch_size);
+            return (
+              <div className="flexbox column" key={row.start_ts || start_time}>
+                <Chip size="small" label={`Phase ${index + 1}`} />
+                <div>{(row.start_ts || start_time).toLocaleString()}</div>
```
mzedel: ...and replace this line with:
```
<div>{new Date(row.start_ts || start_time).toLocaleString(undefined, dateOptions)}</div>
```

src/js/components/deployments/deployment-wizard/review.js
```diff
@@ -0,0 +1,77 @@
...
+              secondary={item.secondary}
+              style={{ flexBasis: 240, margin: '-5px 0' }}
+              dividerDisabled={true}
+            />
+          ))}
+        </List>
+
+        <h4 className="margin-bottom-none">Rollout Schedule</h4>
+
+        <div className="flexbox deployment-phases-list margin-top-small">
```
mzedel: ... and maybe also add `margin-bottom-small` here

src/less/main.less
```diff
@@ -1852,6 +2044,14 @@ span.day {
   }
 }
 
+.deployment-phases-list > .column {
```
mzedel: ... after merging this part was also gone so my addition to the styling would be:
```
.deployment-phases-list {
  overflow: auto;
  max-width: 900px;
  & > .column {
    min-width: 115px;
    margin: 0 5px;
    justify-content: flex-end;
    & > div {
      margin: 5px 0;
    }
    .MuiChip-root {
      width: fit-content;
    }
  }
}
```
or a mix of both...

mzedel:
the 3 onboarding steps in the creation dialog don't seem to work with this change as well, but IMO that should not be a blocker for the beta

src/less/main.less
michaelatmender: Added in b0050eb

src/js/components/deployments/deployment-wizard/review.js
michaelatmender: I forgot we use react-time elsewhere, so I used that to display the time nicely. Also fixed a bug I found with delay calculation: d0afbba

src/js/components/deployments/deployments.js
michaelatmender: I'm not sure why it would happen, I rebased on latest master first

src/js/components/deployments/deployment-wizard/phasesettings.js
michaelatmender: Thanks! Fixed in 0f3daf0

src/js/components/deployments/deployments.js
mzedel: then probably I did something wrong during the merge 🙈

PR #556, opened 2019-10-03
MEN-2732 verify 2fa setup

mzedel:
this requires a `npm install` since I had to change superagent to exclude the auto logout after a failed login attempt (which should be safe to do - as far as I can see)


src/js/components/user-management/twofactorauthsetup.js
```diff
@@ -0,0 +1,90 @@
...
+              or{' '}
+              <a href="https://support.google.com/accounts/answer/1066447" target="_blank">
+                Google Authenticator
+              </a>
+              .
+            </li>
+            <li>Scan the QR code on the right with the authenticator app you just downloaded on your device.</li>
+            <li>
+              <div>
+                Type in your password and the generated code in the input field below and click confirm.
```
michaelatmender: > click confirm
Should be "click 'Verify'" as there is no confirm button!

PR #558, opened 2019-10-07
Integration and mender client version are not equal anymore.

kacf:
Changelog: None

### DIFF
--- a/src/js/stores/app-store.js
+++ b/src/js/stores/app-store.js
@@ -40,6 +40,7 @@ var _groups = [];
 var _releasesRepo = [];
 var _uploadInProgress = false;
 const _hostAddress = mender_environment && mender_environment.hostAddress ? mender_environment.hostAddress : null;
+const _IntegrationVersion = mender_environment && mender_environment.integrationVersion ? mender_environment.integrationVersion : 'master';
 const _MenderVersion = mender_environment && mender_environment.menderVersion ? mender_environment.menderVersion : 'master';
 const _menderArtifactVersion = mender_environment && mender_environment.menderArtifactVersion ? mender_environment.menderArtifactVersion : 'master';
 const _menderDebPackageVersion = mender_environment && mender_environment.menderDebPackageVersion ? mender_environment.menderDebPackageVersion : 'master';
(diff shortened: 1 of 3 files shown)


src/js/stores/app-store.js
```diff
@@ -685,6 +687,16 @@ var AppStore = Object.assign({}, EventEmitter.prototype, {
 
   getShowCreateArtifactDialog: () => _showCreateArtifactDialog,
 
+  getIntegrationVersion: function() {
+    // return version number
+    var version = '';
+    if (_IntegrationVersion) {
+      // if first character NaN, is master branch
+      version = isNaN(_IntegrationVersion.charAt(0)) ? 'master' : _IntegrationVersion;
```
mzedel: that is (at least almost) already accomplished with the `mender_environment.integrationVersion ? mender_environment.integrationVersion : 'master'` - so unless it's the integration version is set, it will default to "master" - so you can probably just reduce it to `getIntegrationVersion: () => _IntegrationVersion,`

src/js/stores/app-store.js
```diff
@@ -40,14 +40,16 @@ var _groups = [];
...
 const _MenderVersion = mender_environment && mender_environment.menderVersion ? mender_environment.menderVersion : 'master';
 const _menderArtifactVersion = mender_environment && mender_environment.menderArtifactVersion ? mender_environment.menderArtifactVersion : 'master';
 const _menderDebPackageVersion = mender_environment && mender_environment.menderDebPackageVersion ? mender_environment.menderDebPackageVersion : 'master';
 var _demoArtifactPort = mender_environment && mender_environment.demoArtifactPort ? mender_environment.demoArtifactPort : 85;
 var _globalSettings = {};
 
 const _versionInformation = {
-  Mender: mender_environment.menderVersion,
+  Integration: mender_environment.integrationVersion,
+  MenderClient: mender_environment.menderVersion,
```
mzedel: the key of the object is directly shown in the GUI (or omitted if no version is set) - to align it with the other keys I would prefer this as 'Mender-Client'

src/js/stores/app-store.js
kacf: Great! Yes, I chose to be a conformist in this unfamiliar code territory. I didn't even look at what the code does! :-P

src/js/stores/app-store.js
kacf: Fixed!

kacf:
This was a fast-forward for 2.2.x, so I did that too.

PR #561, opened 2019-10-11
MEN-2814 added starttime column to deployments list in enterprise

mzedel:
Changelog: Title


src/js/components/deployments/deploymentitem.js
```diff
@@ -79,34 +80,44 @@ export default class DeploymentItem extends React.Component {
         />
       );
     }
+    started = isEnterprise && phases.length >= 1 ? phases[0].start_ts || created : created;
```
michaelatmender: When I created a deployment with 0 phases, this line throws an error:
`Cannot read property 'length' of undefined`

mzedel: should be fixed now as I added an extra check for the existence of phases there

PR #563, opened 2019-10-15
MEN-2816 reuse rollout patterns

mzedel:
previous patterns are stored in globalsettings since they can be shared with other users from the same tenant


mzedel:
Thanks for the reports! The first 2 points should be fixed with the just now pushed commit 👍
The second part is a bit more tricky though, the `undefined%`/ overwriting the recent patterns should be also fixed - but the general alignment of the pattern sizes is still lacking... would this maybe be covered with the phase verification?

michaelatmender:
> The second part is a bit more tricky though, the `undefined%`/ overwriting the recent patterns should be also fixed - but the general alignment of the pattern sizes is still lacking... would this maybe be covered with the phase verification?
I don't think it would be cover it, the validation is really just reacting to user making changes in the inputs: not letting the user click continue if there is a device-less batch.

I think the issue in the second part is also caused by the last phase being editable when you load a previous pattern. The last phase shouldn't be able to be edited, it should always be whatever the remainder is (I think the API also expects the last phase to have no size)

PR #564, opened 2019-10-16
MEN-2819 fix phased deployment retry

mzedel:


src/js/components/deployments/createdeployment.js
```diff
@@ -36,8 +35,11 @@ export default class CreateDialog extends React.Component {
 
   componentDidUpdate(prevProps) {
     // Update state if single device passed from props
-    if ((prevProps.device !== this.props.device) && this.props.device) {
-      this.setState({deploymentDeviceIds: [this.props.device.id]})
+    if (prevProps.device !== this.props.device && this.props.device) {
+      this.setState({ deploymentDeviceIds: [this.props.device.id] });
+    }
+    if (prevProps.deploymentObject !== this.props.deploymentObject && this.props.deploymentObject) {
+      this.setState({ activeStep: this.state.steps.length - 1, ...this.props.deploymentObject });
```
michaelatmender: I would adjust `this.state.steps.length - 1` to simply `0`. So the user can start at the start of the creation wizard and set schedule if needed. I think it's a bit less intuitive to have to click 'back' from the review screen

mzedel: 👌 should now start at `0`

PR #568, opened 2019-10-16
MEN-2820 added frontend validation to disable empty batches in phased…

michaelatmender:
… deployments

Changelog: Title


michaelatmender:
> Looks good so far, just when the form is opened and a custom phase setting is selected (the default 10%, 90%) you can still click next + if you go back from the review section increasing the not-remainder sections causes the rows to collapse or select the single phase option

Nice catch, I think that might be unrelated to the pr and just a bug in general though! I can add a fix here or a separate PR?

mzedel:
Unfortunately the validation is still missing when the form is opened initially - so:
1. the form is opened for the very first time
2. select a group with few devices (e.g. <10)
3. select custom phases (the default 10%, 90%)
4. you can still click next even though the first phase is marked as invalid

And another issue I came across:
1. open the dialog
2. select an invalid configuration
3. cancel
4. open the dialog again
5. leave single phase deployment selected
6. next is disabled

michaelatmender:
@mzedel rewrote it a bit to move the validation up to the dialog component level, should work a lot better now!

mzedel:
@michaelatmender yes, very nice! That works a lot better! The only (tiny) thing now: when the initial size is populated, the values are not rounded (so my Phase 2 is 85.71428571428572 (7 devices))

src/js/components/deployments/deployment-wizard/phasesettings.js
```diff
@@ -87,56 +90,53 @@ export default class PhaseSettings extends React.Component {
...
+          {(deviceCount<1) ? <div className="warning">Phases must have at least 1 device</div> : null }
+          </TableCell>
+          <TableCell><Time value={startTime} format="YYYY-MM-DD HH:mm" /></TableCell>
+          <TableCell>
+            { phase.delay && (index!==props.phases.length-1) ?
+              ( 
                 <div>
                   <Input
-                    value={phase.delay}
+                    value={phase.delayUnit === 'days' ? Math.ceil(phase.delay/24) : phase.delay}
```
mzedel: given that we now also have minutes as an option I think we should just keep the `phase.delay`, otherwise it feels a bit strange that the input suddenly changes, but just when selecting 'days'

michaelatmender: Yep, good catch. Updated now!

PR #573, opened 2019-10-18
cherry-pick of #551

mzedel:
fixed faulty fallback file definition in nginx config
the fallback file has to be defined with a leading / for the file to be found

Changelog: None


mzedel:
the failure in the acceptance tests is due to the changed tests for the new deployment dialog - this should not at all be connected with this change here. so merging...

PR #574, opened 2019-10-19
Modify Pipeline to publish on release branches

lluiscampos:
Follow up from #548 and #549 to align it with the rest of the repos.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -90,7 +90,6 @@ publish:
     - docker:dind
   dependencies:
     - build
-    - test:acceptance
   script:
     - docker load -i image.tar
     - docker tag $SERVICE_IMAGE $DOCKER_REPOSITORY:$COMMIT_TAG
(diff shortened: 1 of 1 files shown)


lluiscampos:
@mzedel Apart from the trivial commit, I wanted to discuss something.

In all Pipelines we define the DOCKER_TAG as `export DOCKER_TAG=${CI_COMMIT_REF_SLUG:-master}`. Shouldn't we default to "unknown" or "dirty" instead of "master"? It is very unlikely, but in case of a GitLab error we would end up overriding "master" tag

What do you think? I have used this exact expression in all repositories.

mzedel:
Good point, since we don't have any `development` or `production` branches, especially none that we publish from, master seemed to me to be the closest to the "unstable" state (and I honestly didn't consider anything other than a possible branchname)... the fallback then is more or less just in case the variable is not set properly and not so much for any other pipeline error...
In that case the `dirty` or `unknown` tags would be a more explicit tag, I agree. Might also be valuable to stop the pipeline right there or not publish at all...

PR #543, opened 2019-09-18
Fix broken URL at end of onboarding to system updates.

estenberg:
NOTE: This will not work for Hosted Mender, so should probably
change it (but at least it is not broken with this change).

Changelog: None

Signed-off-by: Eystein Måløy Stenberg <eystein.maloy.stenberg@northern.tech>

PR #577, opened 2019-10-22
Hosted Mender will be replaced with Mender Professional.

estenberg:
We may also have different plans in future, so do not
mention their names explicitly in case they change.

Should be backported to 2.2 if accepted.


mzedel:
This should be merged only after https://github.com/mendersoftware/gui/pull/543 and https://github.com/estenberg/gui/pull/2 are, in order to fix the failing tests.

estenberg:
estenberg#2 has been merged, please feel free to merge this one once ready.

mzedel:
closing in favour of #582

PR #586, opened 2019-11-08
fix memory leaking store interactions to use redux & inject data through props

mzedel:
this requires a `npm install` to work and the currently failing tests will be addressed (so far they don't have knowledge of the store).

To get a better understanding of the store structure you can look at the `src/js/reducers/*Reducer.js` files, that should have the structure at the top. The actions filling the store are to be found in 
similarly scoped files in `src/js/actions/` and should combined be largely the same as the ones in (the now deprecated) `src/js/action/app-actions.js`.
One difference that is not to be missed is the absence of the single dispatcher and the `onchange` subscriptions to the store state - which leads to the store changes being injected in the connect function of many components (usually found at the very end of the component files).

This should close MEN-2866, MEN-2610, MEN-2698, MEN-2722, MEN-2989, MEN-2988


michaelatmender:
Found a few issues manually testing UI on this PR:

- [ ] When there are 0 devices, it shows empty table instead of placeholder
- [ ] Ungrouped exists when 0 devices
- [ ] Authorizing pending device by checking row + clicking button doesn’t work, no error message (it does work if you expand device details and accept through the auth status dialog)
- [ ] When 1 device connected (not in a group), Ungrouped shows but is empty
- [ ] Device inventory “waiting” box isn’t showing when inventory hasn’t reported yet
- [ ] Loading device by url (like https://localhost/ui/#/devices/id=5ddfca864211fa00010b4685) doesn’t work
- [ ] A group is not deleted when the last device in the group is removed
- [ ] Filter devices - ID not shown, only mac (Mac doesn’t work because of : characters, known issue?)
- [ ] Error when dismissing an accepted device: There was a problem updating the device authorization status: Actions may not have an undefined "type" property. Have you misspelled a constant?…" Not a very user-friendly message in the UI 
- [ ] On “preauthorised devices” tab, sometimes the accepted devices list shows up for a few seconds
- [ ] Rejected devices tab also goes blank every few seconds
- [ ] First Release is not selected by default on loading Releases tab (when coming from another tab). Same if you click direct link like https://localhost/ui/#/releases/mender-demo-artifact-2.2.0
- [ ] Clicking to edit description should focus the input so you can type without clicking again 
- [ ] Uploading artifact doesn’t refresh Releases list when it is successfully uploaded
- [ ] Expanding artifacts a bit buggy when >1 Release exists
- [ ] "Create release with this deployment” button - Release not pre-filled in deployment dialog
- [ ] In progress deployment: “view details” button throws "Cannot read property '0' of undefined” and doesn’t open dialog
- [ ] Editing password in “My account" shows: "There was an error editing the user. failed to decode request body: JSON payload is empty"
- [ ] Changes to user via the user management screen shows 2 “the user has been updated” messages
- [ ] Changing identity from ID to Mac doesn’t show until refreshing page
- [ ] Reload dashboard, then clicking through to finished deployments from dashboard widget shows empty list
- [ ] Pending deployment table content misaligned with headers
- [ ] Filter deployments by group: dropdown contains   *|=ungrouped=|*

michaelatmender:
Some issues found while manually testing

Devices 'refreshing':

- [ ] Pending device tab: list keeps refreshing, even if expanded or dialog open, it closes dialog
  * Not showing correct devices - flashes between 10 and 1 device
- [ ] Preauth devices also does this
    * If Mac is set as ID, this disappears then reappears on each ‘refresh'
- [ ] Device groups: Create a group, then switch to ungrouped/all devices – the devices in group still show for a short time
    * Devices disappear and reappear when using filters too
- [ ] https://localhost/ui/#/devices/group=All%20devices doesn’t work (sometimes loads initially, then devices disappear)

Device filters:

- [ ] Dropdown and remove icons both show in device filters UI
- [ ] Removing filter from middle of list sometimes doesn’t remove the correct filter, or removing first in list leaves a blank one

Settings:

- [ ] Global settings: Device identity has no default selected
  * Heading also shows inside the dropdown alongside options (although it is not clickable)
- [ ] 2FA: verified "check" icon shows up so it looks successful, but so does an error message in the snackbar.  2FA enabled, but not working when trying to log back in.

Onboarding:

- [ ] Clicked 'abort deployment', got a blank screen and error
    * `deployments.js:305 Uncaught (in promise) TypeError: Cannot read property 'body' of undefined at deployments.js:305`
- [ ] 'Connect a device’ button – goes to device groups, when it should open onboarding dialog
- [ ] Helptips seem to disappear on pending device (related to device ‘refresh' issue?)
- [ ] When creating deployment, 'Select a device group’ should have a helptip
- [ ] Deploy your first Application update shows in left menu when deployment has been created but is not yet complete
- [ ] Artifact is expanded by default, so it makes it look like the help tip is pointing at the “remove artifact” button
- [ ] Web app link not loading for me (might be unrelated to this PR)

mzedel:
thanks for the list and I hope that I managed to fix all points, but there is one that I could not reproduce:

- [ ] Deploy your first Application update shows in left menu when deployment has been created but is not yet complete

seemed to work ok as I went through, but I've seen the onboarding tips show up, disappear and show up again on another instance, while competing requests settled... so that should be caught but I'm not so sure about the `Deploy your first Application update`

oleorhagen:
Oh my lord! I thought this was never gonna end. It it actually all done now? :p

PR #591, opened 2019-11-26
MEN-2841 added billing information page

mzedel:
This needs a `npm install` to update material ui, that allows us to use the mui tooltip with an arrow/ indicator 👍 - which aligns better with the underlying invision design 😁 

Changelog: Title


src/js/components/settings/__snapshots__/organization.test.js.snap
```diff
@@ -168,6 +168,44 @@ exports[`renders correctly 1`] = `
...
+      <div
+        className="flexbox space-between"
+      >
+        <div
+          className="MuiTypography-body2 MuiTypography-colorTextSecondary"
+        >
+          Mender Enterprise
+        </div>
+        <a
+          href="https://mender.io/terms/pricing"
```
michaelatmender: Should be https://mender.io/products/pricing

src/js/components/settings/billing.js
```diff
@@ -0,0 +1,227 @@
...
+            </>
+          )}
+          <div className="explanatory-text margin-top">
+            <p className="margin-bottom">
+              To update your billing details or for any other support questions, contact us at <a href="mailto:support@mender.io">support@mender.io</a>.
+            </p>
+            <p>
+              For complete information and full definitions of these fees and how they are calculated, see{' '}
+              <a href="https://mender.io/terms/pricing" target="_blank">
+                https://mender.io/terms/pricing
```
michaelatmender: Should be https://mender.io/products/pricing

src/js/components/settings/organization.js
```diff
@@ -40,6 +42,7 @@ export default class MyOrganization extends React.Component {
     }, 5000);
   }
   render() {
+    const currentPlan = 'Mender Enterprise';
```
michaelatmender: This looks like it will always be set as Mender Enterprise, unless I'm reading it wrong? Tried testing with the following env. variables
```
  hasMultitenancy: "true",
  isEnterprise: "",
  isHosted: "true"
```

src/js/components/settings/billing.js
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  componentDidMount() {
+    this.updateBillingStatement();
+  }
+
+  updateBillingStatement(date = new Date()) {
+    const self = this;
+    const timeframe = { month: date.getMonth(), year: date.getFullYear() };
+    AppActions.getBillingStatus(timeframe).then(information => {
```
michaelatmender: This throws an error - no function called `getBillingStatus`, there is one called `getBillingStatement`

src/js/components/settings/plannotification.js
```diff
@@ -0,0 +1,19 @@
...
+
+import Divider from '@material-ui/core/Divider';
+
+const PlanNotification = ({ className = '', currentPlan, defaultPlan = 'Mender Professional', planClass = '' }) => (
+  <>
+    <div className={`margin-top margin-bottom-small ${className}`}>
+      <div className="explanatory-text billing-subtitle">Current plan:</div>
+      <div className="flexbox space-between">
+        <div className={planClass}>{currentPlan || defaultPlan}</div>
+        <a href="https://mender.io/terms/pricing" target="_blank">
```
michaelatmender: This one should be https://mender.io/products/pricing

src/js/components/settings/__snapshots__/settings.test.js.snap
```diff
@@ -168,6 +168,44 @@ exports[`renders correctly 1`] = `
...
+      <div
+        className="flexbox space-between"
+      >
+        <div
+          className="MuiTypography-body2 MuiTypography-colorTextSecondary"
+        >
+          Mender Enterprise
+        </div>
+        <a
+          href="https://mender.io/terms/pricing"
```
michaelatmender: Should be https://mender.io/products/pricing

src/less/main.less
```diff
@@ -2496,3 +2497,71 @@ div.shake-alert {
...
+  .usage-report {
+    background-color: @lightestgrey;
+    position: relative;
+    .overlay {
+      position: absolute;
+      margin: (-2 * @defaultGutterWidth) (-3 * @defaultGutterWidth);
+      backdrop-filter: blur(3px);
+      z-index: 100;
+      width: 100%;
+      height: 100%;
```
michaelatmender: Could you add
```
padding: 0 30px;
background-color: rgba(247, 247, 247, 0.5)
```
to .overlay? Just to help a bit with readability

src/less/main.less
```diff
@@ -2496,3 +2497,71 @@ div.shake-alert {
     height: 0;
   }
 }
+
+// Billing
+.billing {
```
michaelatmender: The right panel is a bit narrow on smaller screens, could the left panel be reduced in width a little to make room on smaller screens? See screenshot at https://scrn.li/5g9mdd6VkTrsHJ
We could probably reduce the left menu width too to make more space ( by adjusting .leftFixed and .rightFluid)

src/js/components/settings/billing.js
```diff
@@ -0,0 +1,227 @@
...
+      const interactions = interactionList.map(item => {
+        item.billingInformation = item.billingInformation.map(infoItem => ({
+          ...infoItem,
+          unitFee: information[`${infoItem.key}_unit_fee`],
```
michaelatmender: The values should probably be rounded otherwise long values are breaking the layout a bit. Maybe to 4 decimal places? It may occasionally result a penny difference here or there, but I think for usability and readability it makes sense

michaelatmender:
I missed one comment: could the overlay text be moved up a bit? See screenshot: https://scrn.li/xN2RNsdX1UcAoZ

src/js/components/settings/organization.js
mzedel: Thanks for pointing that out and I changed the settings to rely on the Hosted/ Enterprise values.
Those values were more set as a placeholder until we determined where that information might come from, but until then using the environment should be enough! 👍

src/less/main.less
mzedel: done!

src/js/components/settings/billing.js
mzedel: right, I assumed those came ready from the backend - should be limited to 4 decimal places if they are available

src/less/main.less
mzedel: I reduced the width, but only in this class, .rightFluid was a bit too prominent and changes there broke some different things...

mzedel:
Thanks for the review! The issues should all be addressed now.

src/js/components/settings/billing.js
```diff
@@ -0,0 +1,228 @@
...
+        }));
+        return item;
+      });
+
+      self.setState({
+        billingInformation: {
+          interactions,
+          timeframe,
+          timestamp: new Date(information.timestamp),
+          total: information.summary_total_cost
```
michaelatmender: This should be fixed to 2 decimal points

mzedel: argh 🙈 - should be fixed now

mzedel:
during demoing I found another issue: the billing page was only visible on enterprise and not hosted - this is fixed now + the notice about possibly applied discounts is removed

PR #599, opened 2019-12-05
Rename hosted mender setup flag and make setup quiet

alfrunes:
As part of [MEN-2929](https://tracker.mender.io/browse/MEN-2929), we're renaming the `--mender-professional` flag to `--hosted-mender` for the client setup command.

This PR depends on [PR: 457](https://github.com/mendersoftware/mender/pull/457) in the mender repository.

Signed-off-by: Alf-Rune Siqveland <alf.rune@northern.tech>

### DIFF
--- a/src/js/helpers.js
+++ b/src/js/helpers.js
@@ -436,16 +436,16 @@ export const standardizePhases = phases =>
   });
 
 export const getDebConfigurationCode = (ipAddress, isHosted, isEnterprise, token, packageVersion, deviceType = 'generic-armv6') => {
-  let connectionInstructions = `  --demo ${ipAddress ? `--server-ip ${ipAddress}` : ''}`;
+  let connectionInstructions = `  --quiet --demo ${ipAddress ? `--server-ip ${ipAddress}` : ''}`;
   if (isEnterprise || isHosted) {
     const enterpriseSettings = `  --tenant-token $TENANT_TOKEN \\
   --retry-poll 30 \\
   --update-poll 5 \\
   --inventory-poll 5`;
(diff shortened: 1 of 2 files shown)


lluiscampos:
The uni tests need an update, @mzedel can you please guide @alfrunes here?

src/js/helpers.js
```diff
@@ -436,16 +436,16 @@ export const standardizePhases = phases =>
...
   --retry-poll 30 \\
   --update-poll 5 \\
   --inventory-poll 5`;
-    connectionInstructions = `${ipAddress ? `  --server-url ${ipAddress}` : ' '} --server-cert="" \\
-${enterpriseSettings}`;
+    connectionInstructions = `--quiet ${ipAddress ? `  --server-url ${ipAddress}` : ' '} \\
+    --server-cert="" {enterpriseSettings}`;
     if (isHosted) {
-      connectionInstructions = `  --mender-professional \\
+      connectionInstructions = `  --quiet --hosted-mender \\
```
lluiscampos: I don't think you need `--quiet` here, as it will be adding it twice

lluiscampos: @mzedel What about this ^ ? Am I correct?

mzedel: no the `connectionInstructions` will be overwritten - as the (potentially set) `server-url` and `server-cert` would cause problems otherwise...
(initially I thought the "adding it twice" was referring to the internal implementation of the mender setup)

lluiscampos: Ok, thanks. My bad.

PR #601, opened 2019-12-11
Fix for changing page length for pagination of rejected and preauth d…

michaelatmender:
…evices

Changelog: Title


michaelatmender:
Fixed in #603

PR #602, opened 2019-12-24
MEN-2855 Docker HEALTHCHECK added

merlin-northern:
ChangeLog:title

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,4 +17,6 @@ COPY httpd.conf /etc/nginx/nginx.conf
 COPY --from=build /usr/src/app/dist .
 
 ENTRYPOINT ["/entrypoint.sh"]
+HEALTHCHECK --interval=8s --timeout=15s --start-period=120s --retries=128 CMD wget --quiet --tries=1 --spider --output-document=/dev/null 127.0.0.1
 CMD ["nginx"]
+


Dockerfile
```diff
@@ -8,6 +8,7 @@ RUN npm run disclaim
 
 FROM nginx:1.17-alpine
 RUN mkdir -p /var/www/mender-gui/dist
+RUN apk update && apk add curl
```
mzedel: to prevent the `apk` cache from filling the image you might want to use `RUN apk add --no-cache curl` or use the included `wget`

Dockerfile
```diff
@@ -18,4 +19,5 @@ COPY httpd.conf /etc/nginx/nginx.conf
 COPY --from=build /usr/src/app/dist .
 
 ENTRYPOINT ["/entrypoint.sh"]
-CMD ["nginx"]
\ No newline at end of file
+HEALTHCHECK --interval=8s --timeout=15s --start-period=120s --retries=128 CMD curl -f -s -o /dev/null 127.0.0.1
```
mzedel: with the included `wget` this would be `wget --quiet --tries=1 --spider --output-document=/dev/null 127.0.0.1` or `wget -q -t 1 --spider -O /dev/null 127.0.0.1`

merlin-northern: sure! fixed.

Dockerfile
merlin-northern: sure fixed.

PR #603, opened 2020-01-02
Fixed deployment device list page length changing and device ids

mzedel:
this includes the fixes from #601

targeting 2.2.x since this is the closest to the currently deployed version on MP (which runs 2.2.0) and thus should be tested & deployed if it is considered merge worthy.

### DIFF
--- a/src/js/components/common/pagination.js
+++ b/src/js/components/common/pagination.js
@@ -15,9 +15,8 @@ const paginationIndex = 1;
 class TablePaginationActions extends React.Component {
   constructor(props, context) {
     super(props, context);
-    this.state = {
-      pageNo: 1
-    };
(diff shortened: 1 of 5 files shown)


mzedel:
@merlin-northern now that the build is through, please test `mendersoftware/gui:2-2-x@sha256:2e34e1f5f0fdf91c4a49599c76491a670dbe288335318a3c282c2de59f3e5135` on staging!

merlin-northern:
> @merlin-northern now that the build is through, please test `mendersoftware/gui:2-2-x@sha256:2e34e1f5f0fdf91c4a49599c76491a670dbe288335318a3c282c2de59f3e5135` on staging!

deployed.

PR #604, opened 2020-01-08
MEN-3001 exponential backoff deployment refresh

mzedel:


mzedel:
Thank you for the reviews! With the focus on opening the report right after deployment creation, I think I've found the issue - the device list handed to the report was in the wrong format as it was still just the device ids sent to the backend - this should now be fixed. However, I was not able to recreate the crash on the finished deployments... do you have any other pointers for this?

michaelatmender:
> However, I was not able to recreate the crash on the finished deployments... do you have any other pointers for this?

I created a deployment, stayed on the page and let it finish, then clicked to the 'finished' tab and clicked the row to open it. I made sure not to refresh the page or navigate anywhere else while waiting for it to finish

mzedel:
then the previous fix seems to have fixed this as well I think

PR #608, opened 2020-01-13
MEN-2937 small ui fixes

mzedel:


src/js/components/deployments/report.js
```diff
@@ -173,30 +177,10 @@ export class DeploymentReport extends React.Component {
...
-    if (deployment.finished) {
-      finished = <Time value={formatTime(deployment.finished)} format="YYYY-MM-DD HH:mm" />;
-    }
-
     return (
       <div>
-        <div className="report-container">
-          <div className="deploymentInfo" style={{ width: '240px', height: 'auto', margin: '30px 30px 30px 0', display: 'inline-block', verticalAlign: 'top' }}>
+        <div className={`report-container${past ? '' : ' report-progressing'}`}>
+          <div className="deploymentInfo" style={{ marginTop: 30 }}>
```
michaelatmender: Suggest removing marginTop – it seems to cause some misalignment

src/js/components/artifacts/releaserepository.js
```diff
@@ -188,8 +195,9 @@ export class ReleaseRepository extends React.Component {
       );
     }
 
+    const noArtifactsClass = release ? '' : 'muted';
     return (
-      <div className="relative release-repo margin-left" style={{ width: '100%' }}>
+      <div className={`relative release-repo margin-left${wasSelectedRecently ? ' blurred' : ''}`} style={{ width: '100%' }}>
```
michaelatmender: My only problem with this is that it's not consistent with the rest of the UX – it's the only time this blurring effect is used, so it feels a bit unusual to me. Maybe in the short term we can use the standard loading spinner we use elsewhere, and in the medium-term we can find a nice smooth 'loading' effect to use throughout the UI?

src/js/components/deployments/report.js
mzedel: ah I forgot this for the past deployments - this should only be applied for ongoing deployments, as it aligns with the abort button on the right... just waiting for my deployment to be picked up to verify the fix & then the PR is updated :)

src/js/components/artifacts/releaserepository.js
mzedel: agreed, aligning it is better for now and it should now be based on the loader, I thought the blur might be a bit less "intrusive" as the data already exists in the store

src/js/components/deployments/report.js
mzedel: ah I was wrong, the marginTop was there for everything in the report - that should now align properly 👍

PR #612, opened 2020-01-28
moved e2e tests from integration repo to gui

mzedel:
this is to allow aligning ui checks with gui content/ layout changes
adjust ci script accordingly

Changelog: None

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,10 @@
 #directories
 dist/
 node_modules/
+videos
+screenshots
+.vscode
 
 #files
 src/main.css
+.DS_Store
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -56,11 +56,14 @@ test:acceptance:
   script:
     - docker load -i image.tar
     - docker tag $SERVICE_IMAGE $DOCKER_REPOSITORY:pr
-    - apk add git bash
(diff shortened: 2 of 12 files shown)


mzedel:
This is mainly to prevent issues like the failure in https://github.com/mendersoftware/gui/pull/610 in the future and while the integration repo does not yet call the tests in a different pipeline, the test definitions living in this repo seems to be better suited IMHO.

lluiscampos:
> This is mainly to prevent issues like the failure in #610 in the future and while the integration repo does not yet call the tests in a different pipeline, the test definitions living in this repo seems to be better suited IMHO.

I agree with this. From my point of view, conceptually these are "acceptance" tests and they should live in this repo.

mzedel:
The `WARNING: tests/e2e_tests/cypress/screenshots: no matching files` should only have matching files if there was an error during testing (otherwise there should not be screenshots). Do you know how to set this conditionally?

And yes, I'll remove them from the integration repo once this is merged and cherry picked to the 2.2.x branch. Thanks for the heads up!

lluiscampos:
> The `WARNING: tests/e2e_tests/cypress/screenshots: no matching files` should only have matching files if there was an error during testing (otherwise there should not be screenshots). 

Thanks for clarifying.

> Do you know how to set this conditionally?

I don't think it is possible.

PR #614, opened 2020-02-03
MEN-2542 show artifact metadata

mzedel:


src/js/components/artifacts/selectedartifact.js
```diff
@@ -119,7 +134,13 @@ export class SelectedArtifact extends React.Component {
             />
           </FormControl>
         </div>
-
+        {!!metaData.length && (
+          <List className="list-horizontal-flex">
+            {metaData.map(metadata => (
+              <ExpandableAttribute key={metadata.title} primary={metadata.title} secondary={metadata.value || '-'} />
```
kacf: I could be misunderstanding completely, but isn't the `metadata` field for Update Modules' metadata? At least in the spec, the `artifact_provides` and `artifact_depends`, it should appear under `type_info` (https://github.com/mendersoftware/mender-artifact/blob/master/Documentation/artifact-format-v3.md#type-info).

mzedel: So this assumes the metadata sitting as an attribute of the artifact as shown in: https://docs.mender.io/2.0/apis/management-apis/deployments#example-http-response-1 - I don't (yet) know how the backend transforms the data read from the artifact, but I wouldn't be surprised if this has to be adjusted when the exact location is there - so far neither `type_info` nor `metadata` is present when querying the backend with an artifact where a `_depends` is set.
The metadata from the update modules metadata sits in `artifact.updates[0].[meta_data]` - so metadata on a per update basis and is shown further down in the ui tree (when the artifact is expanded and in there the payload is expanded).

PR #615, opened 2020-02-03
MEN-2901 added link to prebuilt mender images to onboarding

mzedel:
Changelog: None

### DIFF
--- a/src/js/components/common/dialogs/physicaldeviceonboarding.js
+++ b/src/js/components/common/dialogs/physicaldeviceonboarding.js
@@ -109,7 +109,7 @@ export class PhysicalDeviceOnboarding extends React.Component {
                 </a>{' '}
                 and once you&apos;re done flashing you can go ahead and proceed to the next step.
               </p>
-              <p>If you already have an image running, you can proceed but some limitations to the kind of updates you can install apply.</p>
+              <p>If you already have an image running, you can proceed with this tutorial but you will not be able to do full system updates later on.</p>
             </div>
           )}
         </div>
--- a/src/js/components/common/dialogs/physicaldeviceonboarding.js
+++ b/src/js/components/common/dialogs/physicaldeviceonboarding.js
@@ -14,6 +14,25 @@ import { findLocalIpAddress } from '../../../actions/appActions';
 import { getDebConfigurationCode } from '../../../helpers';
 import { advanceOnboarding } from '../../../utils/onboardingmanager';
 
+const types = [
+  {
+    title: 'BeagleBone',
+    value: 'beaglebone'
+  },
+  {
+    title: 'Raspberry Pi 3',
+    value: 'raspberrypi3'
+  },
+  {
+    title: 'Raspberry Pi 4',
+    value: 'raspberrypi4'
+  },
+  {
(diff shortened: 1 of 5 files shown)


src/js/components/common/dialogs/physicaldeviceonboarding.js
```diff
@@ -100,6 +100,18 @@ export class PhysicalDeviceOnboarding extends React.Component {
...
+          {hasConvertedImage && (
+            <div className="margin-top">
+              <p>
+                We prepared an image, fully integrated with Mender for you to start with. You can find it on our{' '}
+                <a href={`https://docs.mender.io/${docsVersion}downloads#disk-images`} target="_blank">
+                  downloads page
+                </a>{' '}
+                and once you&apos;re done flashing you can go ahead and proceed to the next step.
+              </p>
+              <p>If you already have an image running, you can proceed but some limitations to the kind of updates you can install apply.</p>
```
lluiscampos: I would rephrase to something that explicitly points to the limitation. Something similar to:
`If you already have an image running, you can proceed with this tutorial but you will not be able to do full system updates later on.`

src/js/components/helptips/onboardingcompletetip.js
```diff
@@ -75,6 +75,7 @@ export class OnboardingCompleteTip extends React.Component {
             might have to refresh the page.
           </p>
           <p>You&apos;ve now got a good foundation in how to use Mender. Look for more help hints in the UI as you go along.</p>
+          <p>If you used one of our pre-built images you can start using full-image and delta updates right away.</p>
```
lluiscampos: The delta updates is an Enterprise feature, is this message only shown on Hosted Mender?

src/js/components/common/dialogs/physicaldeviceonboarding.js
```diff
@@ -133,7 +145,9 @@ export class PhysicalDeviceOnboarding extends React.Component {
 const actionCreators = { findLocalIpAddress, setOnboardingApproach, setOnboardingDeviceType };
 
 const mapStateToProps = state => {
+  const docsVersion = state.app.docsVersion ? `${state.app.docsVersion}/` : 'development/';
```
lluiscampos: Why do you want to version this page? Wouldn't be ok to just point to latest stable release? This way if we add images in the next release, we still show them for an old running server.

mzedel: Mostly because we can't be certain that the latest image versions will always be compatible with the version the user is currently onboarding in (or if they will exist there at all in later versions)...
But I see the benefit of having it just point to the current version... any preference?

src/js/components/helptips/onboardingcompletetip.js
mzedel: 👌 made it conditional and it will only be visible on hosted now - although it might also be an upselling point if it would show up in OS as well

src/js/components/common/dialogs/physicaldeviceonboarding.js
lluiscampos: Ok. Better safe than sorry (i.e. leave it as is)

PR #618, opened 2020-02-04
Add flags for deb package script to handle existing conf

lluiscampos:
In case the device has an existing conf file (which is the case for the
new Raspbian image), dpkg will prompt the user what to do with the
existing conf file (i.e. `/etc/mender/artifact_info`).

These flags will make dpkg keep the existing file.


lluiscampos:
One question: looking a [these lines](https://github.com/mendersoftware/gui/blob/master/src/js/components/help/application-updates/mender-deb-package.js#L39-L40), what is this other `const dpkgCode` used for? Does it need an update with these flags?

lluiscampos:
> One question: looking a [these lines](https://github.com/mendersoftware/gui/blob/master/src/js/components/help/application-updates/mender-deb-package.js#L39-L40), what is this other `const dpkgCode` used for? Does it need an update with these flags?

I guess it does... as the unit test has failed. But can you clarify, please? Maybe is the test snap what I should have not modified?

PR #620, opened 2020-02-04
MEN-2484 added jest coverage reporting during ci runs

mzedel:
should also enforce test threshold around current coverage levels
Changelog: None


mzedel:
@michaelatmender would this be ok to merge? it doesn't really change anything in the functionality but would help greatly for the demo if it was merged before the review 😄

michaelatmender:
@mzedel I missed your comment on this to me, sorry! I blame it on me just coming back from vacation. Yes, looks fine to merge

PR #621, opened 2020-02-05
MEN-2863 friendlier dates

mzedel:


mzedel:
Thanks for looking into this! The times in the deployment device list should now be absolute times again...

PR #624, opened 2020-02-05
Generate the disclaimer via Dockerfile

lluiscampos:
Generate the disclaimer via Dockerfile

So that it can easily be generated externally (i.e. through the release
tool).

The release tool relied on this file being copied all the way to the
nginx server, which does not happen anymore after e2c48bd.


lluiscampos:
We need this to be able to start the release build. Merging without review.

PR #632, opened 2020-02-11
fixed an issue that would let the user list crash after user removal

mzedel:
Changelog: Title


mzedel:
this also has to be cherry-picked to 2.3.x

PR #635, opened 2020-02-20
So that we can workaround a temporary issue with latest docker:dind

mzedel:
See MEN-3200

See https://gitlab.com/gitlab-com/support-forum/issues/5194#note_288438588

Changelog: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -17,7 +17,7 @@ before_script:
 build:
   stage: build
   services:
-    - docker:dind
+    - docker:19.03.5-dind
   script:
     - echo "building gui for ${SERVICE_IMAGE}"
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -109,7 +109,7 @@ publish:image:
 publish:disclaimer:
   stage: publish
   services:
-    - docker:dind
+    - docker:19.03.5-dind
```
lluiscampos: For the publish jobs it is not necessary. Just FYI, it is a good practice anyway ;)

mzedel: Oh right since they are basically just pushing a file instead of any virtualisation - thanks for the learning!

PR #637, opened 2020-02-26
MEN-3245 removed billing information page

mzedel:
Changelog: None


mzedel:
@michaelatmender should this also get in before the next deployment?

PR #644, opened 2020-03-05
Updated device limit notification message

michaelatmender:
Changelog: None

Signed-off-by: Michael <michael.clelland@mender.io>

### DIFF
--- a/src/js/components/header/devicenotifications.js
+++ b/src/js/components/header/devicenotifications.js
@@ -7,6 +7,7 @@ import pluralize from 'pluralize';
 import DeveloperBoardIcon from '@material-ui/icons/DeveloperBoard';
 
 const DeviceNotifications = ({ total, limit, pending }) => {
+  limit = 500;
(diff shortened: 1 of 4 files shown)


src/js/components/header/devicenotifications.js
```diff
@@ -7,6 +7,7 @@ import pluralize from 'pluralize';
 import DeveloperBoardIcon from '@material-ui/icons/DeveloperBoard';
 
 const DeviceNotifications = ({ total, limit, pending }) => {
+  limit = 500;
```
mzedel: To set a default/ fallback value for the `limit` that should go into the parameter list, so: `({ total, limit, pending })` becomes: `({ total, limit = 500, pending })`, otherwise the limit would always be 500 (I think).

michaelatmender: @mzedel I'm a dummy, this was for quick testing and I forgot to delete it! Thanks for spotting it!

PR #645, opened 2020-03-05
MEN-3174 2fa fixes (additional)

mzedel:


src/js/components/user-management/twofactorauthsetup.js
```diff
@@ -18,12 +18,26 @@ export class TwoFactorAuthSetup extends React.Component {
...
+    self.onUnload = self.onUnload.bind(self);
+  }
+
+  componentDidMount() {
+    window.addEventListener('beforeunload', this.onUnload);
   }
 
   componentWillUnmount() {
-    if (!this.state.validated2fa) {
+    if (!this.state.validated2fa && this.props.qrImage) {
```
merlin-northern: isnt that so: `this.props.qrImage` is not null only when qr is loaded? so, it is possible to click the switch to enable 2fa, and close the browser window before qrImage is fully loaded, and fool this? maybe use `qrExpanded` in here?

src/js/components/user-management/login.js
```diff
@@ -50,6 +50,9 @@ export class Login extends React.Component {
       }, 1000);
       self.props.setSnackbar('');
     }
+    if (prevProps.has2FA !== self.props.has2FA && self.props.has2FA) {
+      self.setState({});
+    }
```
merlin-northern: oh yes! that was annoying! (tested just now, it works!)

src/js/actions/userActions.js
```diff
@@ -62,6 +62,8 @@ export const loginUser = userData => (dispatch, getState) =>
       return Promise.all([Promise.reject(err), dispatch(handleLoginError(err, has2FA))]);
```
merlin-northern: finally the tfa token entryfield is not visible on logout! (also tested it, works perfectly)

src/js/helpers.test.js
```diff
@@ -32,3 +29,85 @@ describe('FileSize Component', () => {
     expect(tree).toMatchSnapshot();
   });
 });
+
+describe('getFormattedSize function', () => {
+  it('converts correctly', () => {
+    expect(getFormattedSize()).toEqual('0 Bytes');
+    expect(getFormattedSize(null)).toEqual('0 Bytes');
+    expect(getFormattedSize(0)).toEqual('0 Bytes');
```
merlin-northern: I guess only one scenario left: 
`expect(getFormattedSize(31)).toEqual('31 Bytes');`
just being mean. sorry.
feel free to disregard, since I am not sure getFormattedSize respects this.

mzedel: good point, except it will return `'31.00 Bytes'` - but I won't care for sub byte size 🤷‍♂ 😁

src/js/components/user-management/twofactorauthsetup.js
mzedel: well before the 2fa is fully loaded it won't be in the store - rendering is independent of this... + yes this is only relevant if `qrExpanded`...

PR #646, opened 2020-03-05
restricted 2fa availability to plans above os

mzedel:
Changelog: None

### DIFF
--- a/src/js/components/settings/__snapshots__/plannotification.test.js.snap
+++ b/src/js/components/settings/__snapshots__/plannotification.test.js.snap
@@ -16,7 +16,7 @@ Array [
       <div
         className=""
       >
-        Mender Open Source
+        Mender Starter
       </div>
       <a
         href="https://mender.io/products/pricing"
--- a/src/js/components/user-management/selfusermanagement.js
+++ b/src/js/components/user-management/selfusermanagement.js
@@ -66,7 +66,7 @@ export class SelfUserManagement extends React.Component {
   render() {
     const self = this;
     const { editEmail, editPass, emailFormId, qrExpanded } = self.state;
-    const { currentUser, has2FA, isEnterprise, isHosted } = self.props;
+    const { canHave2FA, currentUser, has2FA, isEnterprise } = self.props;
     const email = currentUser.email;
     return (
       <div style={{ maxWidth: '750px' }} className="margin-top-small">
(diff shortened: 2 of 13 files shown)


src/js/components/settings/__snapshots__/plannotification.test.js.snap
```diff
@@ -16,7 +16,7 @@ Array [
       <div
         className=""
       >
-        Mender Open Source
+        Mender Starter
```
merlin-northern: can we use the const `plan.os` here?

src/js/components/user-management/selfusermanagement.js
```diff
@@ -189,6 +189,7 @@ const mapStateToProps = state => {
   return {
     has2FA: state.users.globalSettings.hasOwnProperty('2fa') && state.users.globalSettings['2fa'] === 'enabled',
     isEnterprise: state.app.features.isEnterprise || (state.app.features.isHosted && plan !== 'os'),
+    isHosted: state.app.features.isEnterprise || state.app.features.isHosted,
```
merlin-northern: isHosted as it is now is misleading, since `isEnterprise` can be true for on-prem installations. and we handle those as normal "hosted" with `isHosted === ture`?

src/js/components/user-management/selfusermanagement.js
```diff
@@ -189,6 +189,7 @@ const mapStateToProps = state => {
   return {
     has2FA: state.users.globalSettings.hasOwnProperty('2fa') && state.users.globalSettings['2fa'] === 'enabled',
     isEnterprise: state.app.features.isEnterprise || (state.app.features.isHosted && plan !== 'os'),
```
merlin-northern: this one I dont get as well, if the plan is not OS  we set isEnterprise to true while on hosted, are we loosing Professional here? which isHosted, is not enterprise, and is not os

src/js/components/settings/__snapshots__/plannotification.test.js.snap
mzedel: that's in a test snapshot - in these files is just the result of the rendering and they should never be touched by an editor... let alone include imports/ other js functionality

src/js/components/user-management/selfusermanagement.js
mzedel: this one and the next comment are very valid points, however the structure of this "feature flag" is a bit special, since it should show a note in commercial, but not in enterprise & professional (so here these two are considered "enterprisey")... but you're right, I'll reword this...

PR #651, opened 2020-03-10
In onboarding, use snapshot link instead of dated link to

estenberg:
dummy system update demo.


estenberg:
Should backport to 2.3 if it looks ok to you. :)

PR #653, opened 2020-03-13
Improve expandable device attribute using text-overflow ellipsis

tranchitella:


tranchitella:
A quick improvement to something which annoys me a lot.

It was:
pre

It is now:
post

@michaelatmender @mzedel what do you think?

mzedel:
if I remember correctly one of the points of the expandable attributes was to not wrap after a single line and offer most of the data on first glance, but have them wrap for long content... unfortunately multiline overflow detection is only supported very poorly across browsers...

tranchitella:
@mzedel @michaelatmender reworked to be applied to the tenant-token only. I've also fixed acceptance tests, as they depended on removed components (conductor and its friends).

PR #657, opened 2020-03-19
MEN-3062 fix group dialogs

mzedel:


mzedel:
Thanks for looking at this and sorry for the oversights - they should now be fixed... I'm not quite sure why this exact error occurs, but the path to get there should no longer be possible 👌

PR #661, opened 2020-03-24
MEN-3345 - Merge Artifact depends i4

mzedel:


kacf:
I will start to test this a bit today.

PR #668, opened 2020-04-01
Fix/pagination after filters

mzedel:


michaelatmender:
Just one issue – doesn't ["ensured device lists never extend max length](https://github.com/mendersoftware/gui/pull/668/commits/b22eca59ae8c5528e3eb69ec8d75297ad5549127) mean that devices get hidden sometimes? To reproduce: If you have >20 qemux86-64 devices, filtering that device type only shows the first 20

mzedel:
hmm that should actually just limit the device list to a maximum of 50 devices, as it would otherwise be possible to end up with a list of 10000 devices showing up, if a filter matches that many... I'll look into it right away!

mzedel:
@michaelatmender checked again and this is in line with the existing filtering as the filtering calls the device inventory endpoint, but so far the pagination is disabled then... that should be fixed with the saved filters though...

michaelatmender:
Although pagination is hidden when filtering, the number of devices displayed in the filtered list _is_ affected by the pagination "per page" setting. 
View all devices, set per page to 10, filter by device type – only the first 10 devices are visible to the user. Pagination is disabled/hidden in the UI but the "per page" is enforced, which is a little unintuitive. I think either there should be no pagination at all (just show the whole list of devices) or pagination is happening and the controls are available.
If pagination must be silently enforced like this, can the per page at least be set very high?

merlin-northern:
this fixes the problem of not being able to accept devices. tested.

PR #670, opened 2020-04-08
MEN-3355 advanced filtering dynamic groups

mzedel:


michaelatmender:
When adding a second rule to a filter, I often get a blank screen and the error:

> react-dom.development.js:25671 Uncaught Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

Looks like it's caused from autocomplete while typing, but I'm not sure

PR #671, opened 2020-04-14
remove reference to local.conf in tenant token tooltip

mirzak:
local.conf is something specific to Yocto and without providing
additional context this line does not make much sense to non-Yocto
users.

Just drop this, as there are many ways to configure the Mender client
nowadays.

Changelog: None

Signed-off-by: Mirza Krak <mirza.krak@northern.tech>


mirzak:
Test fails, did I remove to much @michaelatmender ? :)

PR #685, opened 2020-04-30
MEN-3454 users roles assignment

mzedel:


src/js/components/user-management/__snapshots__/userform.test.js.snap
```diff
@@ -1,109 +1,3 @@
...
...3-2.76 0-5-2.24-5-5 0-.79.2-1.53.53-2.2zm4.31-.78l3.15 3.15.02-.16c0-1.66-1.34-3-3-3l-.17.01z\\"></path></svg></span><span class=\\"MuiTouchRipple-root\\"></span></button></div></div><p class=\\"MuiFormHelperText-root\\" id=\\"component-error-text\\"></p></div><div><div class=\\"help-text\\" id=\\"pass-strength\\">Strength: <meter max=\\"4\\" min=\\"0\\" value=\\"\\" high=\\"3.9\\" optimum=\\"4\\" low=\\"2.5\\"></meter></div><div class=\\"pass-buttons\\"><button class=\\"MuiButtonBase-root MuiButton-root MuiButton-text MuiButton-textPrimary\\" tabindex=\\"0\\" type=\\"button\\"><span class=\\"MuiButton-label\\">Generate</span><span class=\\"MuiTouchRipple-root\\"></span></button></div></div></div><div class=\\"flexbox\\" style=\\"justify-content: flex-end; margin: 24px 0px -16px 0px;\\"><button class=\\"MuiButtonBase-root MuiButton-root MuiButton-contained Mui-disabled Mui-disabled\\" tabindex=\\"-1\\" type=\\"button\\" disabled=\\"\\" id=\\"submit_button\\"><span class=\\"MuiButton-label\\">Create user</span></button></div></form></div><div class=\\"MuiDialogActions-root MuiDialogActions-spacing\\"></div></div></div><div tabindex=\\"0\\" data-test=\\"sentinelEnd\\"></div></div>"`;
```
merlin-northern, line 3: maybe this line can be longer?
ok, can we maybe make it a bit more editor-friendly?

src/js/components/user-management/userform.js
```diff
@@ -1,49 +1,133 @@
...
+  ListItemText,
+  InputLabel,
+  MenuItem,
+  Select
+} from '@material-ui/core';
 
 import Form from '../common/forms/form';
 import TextInput from '../common/forms/textinput';
 import PasswordInput from '../common/forms/passwordinput';
-import FormButton from '../common/forms/formbutton';
```
merlin-northern, line 6: this commit also changes the buttons, but commit message says otherwise.

src/js/components/user-management/userform.js
```diff
@@ -1,49 +1,133 @@
...
+    const isCreation = !Object.keys(props.user).length;
+    this.state = {
+      editPass: isCreation,
+      isCreation,
+      selectedRoles: props.user.roles ? props.user.roles.map(role => props.roles.find(currentRole => currentRole.id === role)) : []
+    };
+  }
+
+  componentDidUpdate(_, prevState) {
+    // TODO: this is needed due to the re-registering of inputs in the form component and should be fixed at some point
```
merlin-northern, line 33: do we have a follow-up ticket?

src/js/components/user-management/userform.js
mzedel: this specific commit just removes the reference to the button, as in the same step the prior usage of the button is removed since the entire dialog was refactored into its own component (where the `FormButton` is replaced with the standard `Button` component)

src/js/components/user-management/__snapshots__/userform.test.js.snap
mzedel: that is an autogenerated snapshot - if that would be modified the corresponding test fails... the format depends on the renderer that is used to generate the snapshot, which in this case has to be a different one than what it was previously, as I encapsulated the userform in its dialog in order to decouple this from the overall usermanagement component...

src/js/components/user-management/userform.js
mzedel: good point, I'll create one...

src/js/components/user-management/userlist.js
```diff
@@ -22,25 +23,34 @@ const UserList = ({ users, currentUser, editUser, removeUser }) => (
...
+              <TableCell>
                 <Time value={user.created_ts} format="YYYY-MM-DD HH:mm" />
               </TableCell>
-              <TableCell align="left">
+              <TableCell>
                 <RelativeTime updateTime={user.updated_ts} />
               </TableCell>
-              <TableCell padding="none">
-                <Button onClick={() => editUser(user)}>Edit</Button>
+              {isEnterprise && <TableCell>{(user.roles || []).map(role => roles.find(currentRole => currentRole.id === role).title).join(', ')}</TableCell>}
```
merlin-northern, line 49: what if
```js
currentRole => currentRole.id === role
```
is null?

mzedel:
Thanks a lot for looking into this! While I agree with the general sentiment of keeping PRs focused on the relevant topic, the button removal is isolated in its own commit https://github.com/mendersoftware/gui/pull/685/commits/9b5b9e3aa85110d3a128dc924508319ef4b6c8b7 and there are no additional references in the codebase (as otherwise the linter would not succeed). Since I came in touch with the `FormButton` component and removed the only consumer of the `buttonHolder` functionality, the only reason for the `FormButton`s existence (the only difference to a regular button) was gone and following: https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html this should be removed...

merlin-northern:
> Thanks a lot for looking into this! While I agree with the general sentiment of keeping PRs focused on the relevant topic, the button removal is isolated in its own commit [9b5b9e3](https://github.com/mendersoftware/gui/commit/9b5b9e3aa85110d3a128dc924508319ef4b6c8b7) and there are no additional references in the codebase (as otherwise the linter would not succeed). Since I came in touch with the `FormButton` component and removed the only consumer of the `buttonHolder` functionality, the only reason for the `FormButton`s existence (the only difference to a regular button) was gone and following: https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html this should be removed...

no worries. it was just nitpicking, and being mean.

src/js/components/user-management/userlist.js
mzedel: that's imported from https://github.com/mendersoftware/gui/pull/685/files#diff-3e329d83c7c7d5c0b9a0a57bca4db744R26 so unless we insert some ill defined value there, this should be defined and if the map call before is executed on an empty list so `[].map(x => x).join(', ')` the join is just ignored (so will just return `''`) 🙂

merlin-northern: it is all assuming that `userConstants.USER_ROLES` contains all roles that are possible. what happens when user has a role that is not defined in `USER_ROLES`? (it can happen, and we have no control over it here.)

mzedel: argh you're right! + now that you mention it, the constants alone won't be sufficient, since the group based roles e.g. would already break it... => so I'll have to get the roles from the store instead and there I have full control about the structure... 👍 thanks for spotting this I'll rework it!

src/js/components/user-management/userlist.js
```diff
@@ -46,7 +46,9 @@ const UserList = ({ currentUser, editUser, isEnterprise, removeUser, roles, user
               <TableCell>
                 <RelativeTime updateTime={user.updated_ts} />
               </TableCell>
-              {isEnterprise && <TableCell>{(user.roles || []).map(role => roles.find(currentRole => currentRole.id === role).title).join(', ')}</TableCell>}
+              {isEnterprise && (
+                <TableCell>{(user.roles || []).map(role => (roles.find(currentRole => currentRole.id === role) || {}).title).join(', ')}</TableCell>
```
merlin-northern, line 50: lovely!
that will print an empty string followed by `,` in case there are unrecognised groups.

merlin-northern:
> Testing this now, I created a 'read only' user and tried to log in with that user. I can't log in (is this expected?), but there is no error message shown in the UI either.

read only user should be able to login. what useradm are you running?

merlin-northern:
> > Testing this now, I created a 'read only' user and tried to log in with that user. I can't log in (is this expected?), but there is no error message shown in the UI either.
> 
> read only user should be able to login. what useradm are you running?

I have just verified that user with observer role can login with this PR, and current useradm-enterprise master. I do have enterprise mode on in GUI as well as enterprise role.

merlin-northern:
> > > Testing this now, I created a 'read only' user and tried to log in with that user. I can't log in (is this expected?), but there is no error message shown in the UI either.
> > 
> > 
> > read only user should be able to login. what useradm are you running?
> 
> I have just verified that user with observer role can login with this PR, and current useradm-enterprise master. I do have enterprise mode on in GUI as well as enterprise role.

but you have to have OBSERVER role in the db. I think we missed that step.

merlin-northern:
> > > > Testing this now, I created a 'read only' user and tried to log in with that user. I can't log in (is this expected?), but there is no error message shown in the UI either.
> > > 
> > > 
> > > read only user should be able to login. what useradm are you running?
> > 
> > 
> > I have just verified that user with observer role can login with this PR, and current useradm-enterprise master. I do have enterprise mode on in GUI as well as enterprise role.
> 
> but you have to have OBSERVER role in the db. I think we missed that step.

https://github.com/mendersoftware/useradm-enterprise/pull/63

PR #686, opened 2020-05-04
Use deployments count header & fix plan based deployment retrieval

mzedel:


src/js/actions/deploymentActions.js
```diff
@@ -60,7 +59,15 @@ export const getDeploymentsByStatus = (status, page = default_page, per_page = d
...
         return accu;
       },
-      [dispatch({ type: DeploymentConstants[`RECEIVE_${status.toUpperCase()}_DEPLOYMENTS`], deployments, deploymentIds, status })]
+      [
+        dispatch({
+          type: DeploymentConstants[`RECEIVE_${status.toUpperCase()}_DEPLOYMENTS`],
+          deployments,
+          deploymentIds,
+          status,
+          total: Number(res.headers['x-total-count'])
```
merlin-northern, line 71: nitpicking: x-total-count should be a const

src/js/components/deployments/inprogressdeployments.js
```diff
@@ -70,18 +70,14 @@ export class Progress extends React.Component {
...
-        tasks.push(self.props.getDeploymentsByStatus(deploymentStatus, page, perPage));
-      }
-      return Promise.all(tasks)
-        .then(([countAction, deploymentsAction]) => {
+    return self.setState({ [`${mappedDeploymentStatus}Page`]: page, [`${mappedDeploymentStatus}PerPage`]: perPage }, () =>
+      Promise.resolve(self.props.getDeploymentsByStatus(deploymentStatus, page, perPage))
+        .then(deploymentsAction => {
           clearRetryTimer(deploymentStatus, self.props.setSnackbar);
-          if (countAction.deploymentIds.length && deploymentsAction && !deploymentsAction[0].deploymentIds.length) {
+          if (deploymentsAction && deploymentsAction[0].total && !deploymentsAction[0].deploymentIds.length) {
```
merlin-northern, line 80: I would check length here

src/js/components/deployments/pastdeployments.js
```diff
@@ -97,25 +97,13 @@ export class Past extends React.Component {
...
-      if (fullRefresh) {
-        tasks.push(self.props.getDeploymentsByStatus(type, page, perPage, roundedStartDate, roundedEndDate, group));
-      }
-      return Promise.all(tasks)
-        .then(([countAction, deploymentsAction]) => {
+      return Promise.resolve(self.props.getDeploymentsByStatus(type, page, perPage, roundedStartDate, roundedEndDate, group))
+        .then(deploymentsAction => {
           clearRetryTimer(type, self.props.setSnackbar);
-          if (countAction.deploymentIds.length && deploymentsAction && !deploymentsAction[0].deploymentIds.length) {
+          if (deploymentsAction && deploymentsAction[0].total && !deploymentsAction[0].deploymentIds.length) {
```
merlin-northern, line 106: and here as well I would check length

src/js/actions/deploymentActions.js
```diff
@@ -30,7 +30,10 @@ export const getDeployments = (page = default_page, per_page = default_per_page)
         accu[item.status].push(item);
         return accu;
       },
-      { finished: [], inprogress: [], pending: [] }
+      Object.keys(getState().deployments.byStatus).reduce((accu, item) => {
+        accu[item] = [];
+        return accu;
+      }, {})
```
merlin-northern, line 36: real nitpicking, this is used twice, could be a separate util, `emptyStatusHash`

merlin-northern:
looks good to me.
does not work: when I mark in deployments creator "set default" next to retries count. with the next deployment it  is default there again.
 * tested that on RBAC access denied a nice error message is seen:
```
    Error creating deployment. Access denied (RBAC). [Request ID: 3bf2624b]
```
 * tested that scheduled deployments go into the scheduled tab
    * one thing here: there should be a route to redirect to "Scheduled" tab after scheduled deployment is created.

src/js/actions/deploymentActions.js
mzedel: Good point - I'll extract this to the api modules, but in a different PR ;)

src/js/components/deployments/inprogressdeployments.js
mzedel: I would choose to disregard this, since the actions either return a single object or an array if they are a) returning a promise and b) successful - otherwise the returned value will be undefined... until now the returned object depends on the action called (so one of <promise returning object|promise returning array|plain object of dispatched action>) so I'm confident enough that this is aligned as long as the UI doesn't blow up the first time it's called 🤷‍♂️

src/js/actions/deploymentActions.js
mzedel: argh, yep - that's annoying... and sounds like I'll have 2 small PRs to follow up with 😄!

mzedel: https://github.com/mendersoftware/gui/pull/690

src/js/actions/deploymentActions.js
mzedel: #689

src/js/actions/deploymentActions.js
merlin-northern: thanks. approved.

src/js/actions/deploymentActions.js
merlin-northern: thanks. approved.

PR #689, opened 2020-05-06
extracted response headers used during api parsing

mzedel:
Changelog: None


merlin-northern:
verifed that:
* header `'name'` only exists in two const definition
* `res.header` everywhere is used only with the consts

PR #691, opened 2020-05-06
Men 3398 roles settings

mzedel:
depends on #685


src/js/components/user-management/userform.js
```diff
@@ -1,49 +1,133 @@
...
+  handleCheckboxClick(isSelected, role) {
+    const foundIndex = this.state.selectedRoles.findIndex(currentRole => role.id === currentRole.id);
+    let newlySelectedRoles = [...this.state.selectedRoles];
+    if (isSelected) {
+      newlySelectedRoles.push(role);
+    } else if (foundIndex > -1) {
+      newlySelectedRoles.splice(foundIndex);
+    }
+    this.setState({ selectedRoles: newlySelectedRoles });
+  }
```
merlin-northern, line 48: what happens now is: 
* when you click the first checkbox (atm "Admin") it checks/unchecks
* when you click the second one the list scrolls up, so it does when you check the third one
* when you check them in order from to top bottom, and then un check the first (or second) all the lower ones are unchecked as well
* when you check them from the bottom, the same automatic unchecking happens upwards
* when you start checking with the middle, the automatic unchecking happens when (un)checking the middle one
is that expected behaviour?

src/js/components/settings/__snapshots__/roles.test.js.snap
```diff
@@ -0,0 +1,404 @@
...
+            <span
+              className="MuiButton-label"
+            >
+              Cancel
+            </span>
+          </button>
+          <a
+            aria-disabled={false}
+            className="MuiButtonBase-root MuiButton-root MuiButton-contained MuiButton-containedSecondary"
+            href="mailto:support@mender.io?subject=test: Create group&body=%0A%20%20%20%20Organization%20ID%0A%20%20%20%201%0A%20%20%20%20Organization%20name%0A%20%20%20%20test%0A%20%20%20%20Plan%20name%0A%20%20%20%20Mender%20Enterprise%0A%20%20%20%20I%20would%20like%20to%20create%20a%20group%20with%20the%20following%20settings%3A%0A%20%20%20%20%20%20name%3A%20%2C%0A%20%20%20%20%20%20description%3A%20%2C%0A%20%20%20%20%20%20allow%20managing%20other%20users%3A%20No%2C%0A%20%20%20%20%20%20groups%3A%20"
```
michaelatmender, line 377: Maybe change "Group" to "Role" in the email subject and body text

mzedel:
with https://github.com/mendersoftware/useradm-enterprise/pull/67 we could have this done through the API and could avoid the email step entirely - I would love to put this in here as well if that's ok with you..?

michaelatmender:
> with [mendersoftware/useradm-enterprise#67](https://github.com/mendersoftware/useradm-enterprise/pull/67) we could have this done through the API and could avoid the email step entirely - I would love to put this in here as well if that's ok with you..?

Yeah that would be great, I think the email step is a bit obtrusive for the user

michaelatmender:
> When there are no user-created device groups there is a "0" displayed in place of group selection. 'All devices' is not an option – is this by design? (I'm a little hazy on the workings of rbac at this stage)

I suspect this is a problem with my setup with the test stress client. My devices are "offline" because of some strange timestamp bug, it's affecting the software distribution feature too

mzedel:
Thanks for looking into this again, I took care of the 0 and the submit button... the timestamp issue you came across seems to be caused by a change in the backend - should be fixed with #700, but I'll have to follow up for the software distribution

michaelatmender:
> Thanks for looking into this again, I took care of the 0 and the submit button... the timestamp issue you came across seems to be caused by a change in the backend - should be fixed with #700, but I'll have to follow up for the software distribution

It turned out to be unrelated, I deleted the comment as soon as I wrote it!

PR #693, opened 2020-05-08
Dynamic deployments improvements

mzedel:
this should allow creation of dynamics groups without devices present, also with custom attributes (custom as in: not known to the gui) + dynamic deployments should see their status reflected better in the pending deployments list + the group name should now detail the filter that belongs to the group, if the group is a dynamic one


mzedel:
Thanks for that, once again I'm glad to have you keeping an eye on the design - I would have cluttered many areas just like this otherwise 😅.
While getting the missing group name in I just blindly unified both occurrences - now only the group name should be in the overview 👌

PR #703, opened 2020-05-20
MEN-3397 role based devices display

mzedel:
also closing: MEN-3456, MEN-3457

### DIFF
--- a/src/js/actions/userActions.js
+++ b/src/js/actions/userActions.js
@@ -136,7 +136,6 @@ export const getRoles = () => (dispatch, getState) =>
 
 export const createRole = roleData => dispatch => {
   let permissions = roleData.groups.map(group => ({ action: 'CREATE_DEPLOYMENT', object: { type: 'DEVICE_GROUP', value: group } }));
-  permissions.push({ action: 'http', object: { type: 'any', value: `any` } });
   if (roleData.allowUserManagement) {
     permissions.push({
       action: 'http',
--- a/src/js/actions/userActions.js
+++ b/src/js/actions/userActions.js
@@ -136,6 +136,7 @@ export const getRoles = () => (dispatch, getState) =>
 
 export const createRole = roleData => dispatch => {
   let permissions = roleData.groups.map(group => ({ action: 'CREATE_DEPLOYMENT', object: { type: 'DEVICE_GROUP', value: group } }));
+  permissions.push({ action: 'http', object: { type: 'any', value: `any` } });
   if (roleData.allowUserManagement) {
     permissions.push({
       action: 'http',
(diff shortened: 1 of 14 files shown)


src/js/actions/userActions.js
```diff
@@ -136,6 +136,7 @@ export const getRoles = () => (dispatch, getState) =>
 
 export const createRole = roleData => dispatch => {
   let permissions = roleData.groups.map(group => ({ action: 'CREATE_DEPLOYMENT', object: { type: 'DEVICE_GROUP', value: group } }));
+  permissions.push({ action: 'http', object: { type: 'any', value: `any` } });
```
tranchitella, line 139: @mzedel I believe this is fundamentally wrong; the right way to have a user limited to a set of groups is to combine that permission with another role (e.g. the observer/read-only, the CI, or whatever else). This hard-coded permission makes it impossible to combine the limit-to-group with other RBAC roles that limit features.

src/js/components/header/header.js
```diff
@@ -299,8 +302,22 @@ const actionCreators = {
...
+  const currentUser = state.users.byId[state.users.currentUser];
+  let allowUserManagement = false;
+  if (currentUser?.roles) {
+    // TODO: move these + additional role checks into selectors
+    const isAdmin = currentUser.roles.some(role => role === 'RBAC_ROLE_PERMIT_ALL');
+    allowUserManagement =
+      isAdmin ||
+      currentUser.roles.some(role =>
+        state.users.rolesById[role]?.permissions.some(
+          permission => permission.action === 'http' && permission.object.value === '/api/management/v1/useradm/.*' && ['any'].includes(permission.object.type)
```
tranchitella, line 314: Was this discussed with @merlin-northern? I'm not aware of any built-in role with this permission, shall we define a specific role to manage users? From my understanding, the admins are assigned to the role `RBAC_ROLE_PERMIT_ALL`, and this is the only check needed.

merlin-northern: I am sure we should define a specific role to manage users. we started with what we had, now I think we can move forward.

mzedel: yep, that permission is only given to allow user management, while not giving full admin access...
if it's possible to have the role defined in the backend it would be all the better and I'll happily change it here as well

src/js/actions/userActions.js
mzedel: understood... then I think I'll have to signal adding the observer role in the UI, as the groups and user management options won't be usable on their own

mzedel:
thanks for looking into this @tranchitella! I'll adjust the role creation right away...

src/js/components/header/header.js
tranchitella: @mzedel ack, as soon as we get good default roles we'll change the UI to take advantage of them.

src/js/actions/userActions.js
tranchitella: @mzedel we started discussing with @merlin-northern a few additional "default" roles which can be used to create useful users without the "permit all". Keep you posted.

PR #705, opened 2020-05-21
MEN-3306: add cancel subscription form in organization page

tranchitella:
Changelog: title


src/js/api/organizations-api.js
```diff
@@ -0,0 +1,28 @@
+import Cookies from 'universal-cookie';
+import { unauthorizedRedirect } from '../auth';
+
+const cookies = new Cookies();
+var request = require('superagent');
+
+const Api = {
+  postCancelRequest: (url, userData) => {
```
mzedel, line 8: the rest of the api calls mostly stick to general call structures in their naming, since the entire target url is handed over - it would be fitting to do this here as well or just use the `general-api.js` to get this in.

For context: Since most of the api interaction code is the same, I intend to strip down the different files to reuse most of the code whenever possible...

src/js/actions/organizationActions.js
```diff
@@ -0,0 +1,7 @@
+import OrganizationsApi from '../api/organizations-api';
+
+const apiUrl = '/api/management/v2';
+const tenantadmApiUrl = `${apiUrl}/tenantadm`;
```
mzedel, line 4: 👍 good point in having those separated... I'll make sure to bring in the rest of the tenantadm related interactions!

tranchitella: good :)

src/js/api/organizations-api.js
tranchitella: I force pushed and moved the API call to `general-api.js` introducing the `post` method, let me know if I understood correctly what you asked me. :)

mzedel: Thank you very much, that was spot on 😃!

PR #712, opened 2020-05-28
MEN-3450: edit role

tranchitella:


src/js/components/settings/roles.js
```diff
@@ -91,13 +121,16 @@ export class RoleManagement extends React.Component {
...
+                  {role.editable && <Button onClick={() => self.removeRole(role.id)}>Remove</Button>}
+                </TableCell>
               </TableRow>
             ))}
           </TableBody>
         </Table>
 
-        {!adding && <Chip className="margin-top-small" color="primary" icon={<AddIcon />} label="Add a role" onClick={() => self.setState({ adding: true })} />}
+        {!adding && <Chip className="margin-top-small" color="primary" icon={<AddIcon />} label="Add a role" onClick={() => self.addRole()} />}
         <Collapse in={adding} className="margin-right-small filter-wrapper" classes={{ wrapperInner: 'margin-bottom-small margin-right' }}>
```
mzedel, line 134: could you combine the two Collapses and just render the internal things conditionally?

mzedel:
everything looks good, if possible try to get rid of the 2nd Collapse instance, but if not, it will be accepted as is as well!

tranchitella:
@mzedel done, but I suppose this is going to lower code coverage because we won't render them in the snapshot. Is this what you intended?

src/js/components/settings/roles.js
tranchitella: Done

src/js/components/settings/roles.js
```diff
@@ -91,70 +121,133 @@ export class RoleManagement extends React.Component {
...
+                  variant="contained"
+                  target="_blank"
+                  disabled={!(name && (allowUserManagement || groups.some(group => group.selected)))}
+                  onClick={() => self.onSubmit()}
+                >
+                  Submit
+                </Button>
+              </div>
+            </>
           )}
```
mzedel, line 243: @tranchitella due to the similarity in both cases I was more aiming at something like this:

```javascript
<Collapse in={adding || editing} className="margin-right-small filter-wrapper" classes={{ wrapperInner: 'margin-bottom-small margin-right' }}>
  <>
    <h4 style={{ marginTop: 5 }}>{adding ? 'Add a' : 'Edit the'} role</h4>
    <TextField
      label="Role name"
      id="role-name"
      value={name}
      disabled={editing}
      onChange={e => self.setState({ name: e.target.value })}
      style={{ marginTop: 0, marginRight: 30 }}
    />
    <TextField
      label="Description"
      id="role-description"
      value={description}
      placeholder="-"
      onChange={e => self.setState({ description: e.target.value })}
      style={{ marginTop: 0, marginRight: 30 }}
    />
    <div>
      <FormControlLabel
        control={<Checkbox color="primary" onChange={(e, checked) => self.setState({ allowUserManagement: checked })} />}
        checked={allowUserManagement}
        label="Allow to manage other users"
      />
    </div>
    {!!groups.length && (
      <div className="flexbox column margin-top-small">
        <div>Device group permission</div>
        {groups.map(group => (
          <FormControlLabel
            style={{ marginTop: 0, marginLeft: 0 }}
            key={group.name}
            control={<Checkbox color="primary" checked={group.selected} onChange={(e, checked) => self.handleGroupSelection(checked, group)} />}
            label={group.name}
          />
        ))}
      </div>
    )}
    <div className="flexbox centered" style={{ justifyContent: 'flex-end' }}>
      <Button onClick={() => self.onCancel()} style={{ marginRight: 15 }}>
        Cancel
      </Button>
      <Button
        color="secondary"
        variant="contained"
        target="_blank"
        disabled={!(name && (allowUserManagement || groups.some(group => group.selected)))}
        onClick={() => self.onSubmit()}
      >
        Submit
      </Button>
    </div>
  </>
</Collapse>
```

which should increase the shared parts - or you could also extract the adding/ editing sections into a parameterized component (which could result in a bit more work though)

tranchitella: Yup, you are right, I've updated my PR

tranchitella:
@mzedel new round of review :)

tranchitella:
@mzedel I think codecov is confused by force-pushes, it happens in back-end services as well.

PR #714, opened 2020-05-28
2.4.x: MEN-3600: Fix incorrect installation instructions for Enterprise demo.

kacf:
The demo for Enterprise is not officially supported, but we use it for
internal testing, therefore this is important.

Changelog: None

(cherry picked from commit aa4c06f6b39ccc47544f0c4c5597ef8767f829ee)


kacf:
@mzedel: Hmm, why is the lint stage failing here and not in master?

mzedel:
that is very strange - on master we have a newer version of prettier, however I would expect this to fail in any case, as it is an indentation error that should cause problems... hmm... I deferred the entire formatting to prettier in this case and kept the errors from the linter - that's what might keep the error here from showing up on master...

PR #727, opened 2020-06-02
MEN-3630: enhance the list of deployments for lower screen resolutions

tranchitella:
Changelog: none


tranchitella:
Preview of the change, for lower resolutions (<1600px).
preview

PR #751, opened 2020-06-08
MEN-3556 include optional tracker

mzedel:


src/js/components/header/header.js
```diff
@@ -42,11 +43,15 @@ export class Header extends React.Component {
...
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps) {
     const sessionId = this.cookies.get('JWT');
-    if ((!sessionId || !this.props.user || !this.props.user.id || !this.props.user.email.length) && !this.state.gettingUser && !this.state.loggingOut) {
+    const { hasTrackingEnabled, isEnterprise, trackingCode, user } = this.props;
+    if ((!sessionId || !user || !user.id || !user.email.length) && !this.state.gettingUser && !this.state.loggingOut) {
       this._updateUsername();
     }
+    if (prevProps.hasTrackingEnabled !== hasTrackingEnabled && trackingCode && !isEnterprise && hasTrackingEnabled) {
```
tranchitella, line 61: not sure here, why do you need the `!isEnterprise`? It looks weird, because you can have enterprise hosted users.

src/js/components/header/header.js
```diff
@@ -67,7 +76,11 @@ export class Header extends React.Component {
     this.props.getDevicesByStatus(DEVICE_STATES.accepted);
     this.props.getDevicesByStatus(DEVICE_STATES.pending);
     this.props.getDeviceLimit();
-    this.props.getGlobalSettings();
+    this.props.getGlobalSettings().then(() => {
+      if (this.cookies.get('_ga') && typeof this.props.hasTrackingEnabled === 'undefined') {
+        this.props.saveUserSettings({ trackingConsentGiven: true });
+      }
+    });
```
tranchitella, line 83: Nice approach!

src/js/components/header/header.js
mzedel: With enterprise setups in general I thought it might get more tricky due to the implications it might have for whatever policies they might have regarding the storage and use of their data... I wouldn't expect this to be as strict for smaller companies. On the other hand - if they have the `_ga` cookie set that could be enough to just continue using it 🤷‍♂️

tranchitella: To be honest, I would just skip the distinction and enable GA independently from the plan.
If you want to be sure, we can ask Eystein... anyway, approving the the PR.

mzedel: 👌 - then it is now solely based on the presence of the GA code...

PR #754, opened 2020-06-09
MEN-3557 application tracking integration

mzedel:
depends on #751


src/js/components/app.js
```diff
@@ -18,6 +19,19 @@ import { getOnboardingComponentFor } from '../utils/onboardingmanager';
 const timeout = 900000; // 15 minutes idle time
 
 class AppRoot extends React.PureComponent {
+  componentDidMount() {
+    this.props.history.listen(location => {
+      let page = location.pathname || '';
+      ReactGA.set({ page });
```
tranchitella, line 25: Why do you need this set?

src/js/errorboundary.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
```
tranchitella, line 1: Good stuff!

src/js/components/app.js
```diff
@@ -18,6 +19,19 @@ import { getOnboardingComponentFor } from '../utils/onboardingmanager';
...
+      // unless we're on a help page the path will get cut off to the rough region, to ensure we're not sending any sensitive device/ group/ deployment names etc.
+      if (!location.pathname.startsWith('/help')) {
+        const path = location.pathname.split('/');
+        page = path.length > 1 ? path[1] : path[0];
+      }
```
tranchitella, line 30: I'm not 100% sure this is a good idea: we are losing details about what the user is doing and how he transitions from one page to another, without the full path, don't you think so? If you are concerned about disclosing info, do you think it is possible to track the page view using the name of the route (or the component used to render the route) instead?

src/js/components/app.js
mzedel: that should act as a https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#dimension - I actually forgot to assign this with the stripped down path... should now make more sense I hope

src/js/errorboundary.js
mzedel: Over time this should allow catching & handling errors a bit closer to the erroneous area - once they are placed a bit further around

src/js/components/app.js
tranchitella: @mzedel yup, but why do you need to define the page as custom dimension? It is already tracked as URL by pageView.

src/js/components/app.js
mzedel: I tried to make it a bit more detailed, but wanted to stay away from adding the tracking in the different components themselves... the only other place with a more unified access to the components & route would be the router, however the component access there is limited to the top level containers and doesn't really provide additional insight, compared to just the rough area... dialogs are a bit outside of this flow though... hmm I'll try to find a way to hook into their lifecycle...

src/js/components/app.js
mzedel: hmm that was supposed to track the context area/ tab the user is working with - I thought the pageview tracking was not automatically splitting this up but if it does I'll happily remove it...

tranchitella: @mzedel I confirm you when you call `ReactGA.pageview(page);` the page variable, which contains the URL, is available in Google Analytics. No need to track it as a custom dimension.

mzedel: thanks for the clarification - should now be removed 👌!

PR #762, opened 2020-06-12
MEN-3673: do not apply filters with empty string as value

tranchitella:


src/js/components/devices/filteritem.js
```diff
@@ -86,7 +86,7 @@ export default class FilterItem extends React.Component {
     clearTimeout(self.timer);
     self.timer = setTimeout(
       () =>
-        self.state.key && self.state.value
+        self.state.key
```
mzedel, line 89: to prevent this from happening to the key and get around the need for the filtering in the `onFilterChange` you could:
```suggestion
        !!(self.state.key && self.state.value)
```

tranchitella: @mzedel uhm, not sure I understand. To reproduce the issue, you have to:
1. filter by ID, entering the right value, you see the device.
2. remove a character from the ID, it is now a value cannot be found, you don't see the device anymore
3. you remove the value completely; at this point, if I use your proposed change, the filter change is not detected and the previous value (point 2) is still applied, so the device is not shown in the list.

What my change does, is triggering the filter update also when I empty the value, but effectively ignoring the filter when composing the list to be sent to the backend. Does it make sense?

mzedel: Ah right, there notifying about an empty filter I didn't think about - sorry for the noise then!

PR #773, opened 2020-06-16
MEN-3654 made access to device connection dialog permanent

mzedel:
Changelog: None


mzedel:
Thanks for spotting this, should now be fixed 👌

PR #774, opened 2020-06-16
Remove third signup step

mzedel:


mzedel:
😅 I was reminded of it in the demo and if we can avoid it 🤷‍♂️

PR #775, opened 2020-06-17
MEN-3608: show more friendly message on RBAC forbidden errors

tranchitella:
Changelog: none

### DIFF
--- a/src/js/api/general-api.js
+++ b/src/js/api/general-api.js
@@ -11,7 +11,15 @@ export const headerNames = {
   total: 'x-total-count'
 };
 
-const endHandler = (error, res, reject, resolve) => (error || !res.ok ? reject({ error, res }) : resolve(res));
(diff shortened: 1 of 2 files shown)


src/js/api/general-api.js
```diff
@@ -11,7 +11,16 @@ export const headerNames = {
...
 };
 
-const endHandler = (error, res, reject, resolve) => (error || !res.ok ? reject({ error, res }) : resolve(res));
+const endHandler = (error, res, reject, resolve) => {
+  if (error || !res.ok) {
+    if (res && res.statusCode == 403) {
+      res.body.error = res.body.error.message;
+    }
+    reject({ error, res });
+    return Promise.reject();
```
mzedel, line 20:
```suggestion
    return reject({ error, res });
```

tranchitella: @mzedel done!

PR #799, opened 2020-06-26
MEN-3719 align docs links

mzedel:


src/js/components/help/__snapshots__/system-updates.test.js.snap
```diff
@@ -58,7 +58,7 @@ exports[`SystemUpdates Component renders correctly 1`] = `
     You can learn more about Mender's A/B partition architecture
      
     <a
-      href="https://docs.mender.io/architecture/overview"
+      href="https://docs.mender.io/overview/introduction"
```
mirzak, line 61: This this one need a `${docsVersion}`

src/js/components/help/system-updates.js
```diff
@@ -37,7 +37,7 @@ const SystemUpdates = ({ docsVersion = '' }) => (
 
     <p>
       You can learn more about Mender&#39;s A&#47;B partition architecture{' '}
-      <a href={`https://docs.mender.io/${docsVersion}architecture/overview`} target="_blank">
+      <a href={`https://docs.mender.io/${docsVersion}overview/introduction`} target="_blank">
```
mirzak, line 40: Do not know if it is strictly required, but it looks to me a `/`is missing after `${docsVersion}`?

mirzak: There are several occurrences of this, so this comment applies to all of them.

src/js/components/help/__snapshots__/system-updates.test.js.snap
mzedel: the `.snap` files are the results of the test renderings - and for all the help pages the input data from [here](https://github.com/mendersoftware/gui/blob/master/src/js/components/help/mockData.js#L2) is used, which defines the docsVersion as just `''` - so the version path is omitted 👌

src/js/components/help/system-updates.js
mzedel: the `/` is supplied with the `docsVersion`, since this allows omitting the docsVersion/ supplying no version and automatically getting the link to the default docs version

PR #800, opened 2020-06-26
MEN-3717 onboarding docs alignment

mzedel:


src/js/components/common/dialogs/physicaldeviceonboarding.js
```diff
@@ -12,10 +12,6 @@ import { getDebConfigurationCode } from '../../../helpers';
 import { advanceOnboarding } from '../../../utils/onboardingmanager';
 
 const types = [
-  {
-    title: 'BeagleBone',
-    value: 'beaglebone'
-  },
```
mirzak, line 18: I would also remove:

```
  {
    title: 'Generic ARMv6 or newer',
    value: 'generic-armv6'
  }
```

@estenberg, do you object?

Reasoning being, our onboarding use-cases (application, docker, system update) will only work on Raspberry Pi or QEMU and even if technically the "generic ARMv6" would support the "application" use-case it would become very confusing at the end when we link to the system & docker updates sections which only work on Raspberry Pi/QEMU (docker is an exception on QEMU but we are trying to fix that).

Before I comment on the rest of the rest (as this I think has impact on the rest of the content), lets sync up here :).

If we keep this we have 3 paths:

- Raspberry Pi - > application update -> system update -> docker update
- QEMU -> application update -> system update -> docker update (* not supported today but hopefully we can fix this)
- Generic ARM -> Application update

src/js/components/common/dialogs/deviceconnectiondialog.js
```diff
@@ -73,7 +73,7 @@ export class DeviceConnectionDialog extends React.Component {
               </p>
             </ReactTooltip>
           </div>
-          <p>The steps are optimized for Linux workstations and devices running an OS in the Debian family such as Raspbian, Ubuntu and Debian.</p>
+          <p>The steps are optimized for Linux workstations and Raspberry Pi devices, but can be used on any ARM device running a Debian family OS.</p>
```
mirzak, line 76: If we get rid of the "generic ARM" I would change this whole section to something like this:

----

During the guided evaluation we will walk you through installing Mender on a 
device and deploying:

* a simple application update
* a full system update.
* a Docker container update

The evaluation of hosted Mender is optimized for Raspberry Pi 3 & 4 running 
Raspberry Pi OS. If you have such a device at hand you can proceed to prepare
and connect your device.

    < Update images to only contain Raspbian + maybe a picture of the RPi SBC >

    < PREPARE AND CONNECT MY RASPBERRY PI >

If you don't have a Raspberry Pi, you can use a Docker-run virtual device to try out Mender

    < I don't have my own device - use a virtual device for now >

mirzak: To further elaborate, I think the first pop-up in the UI should me very similar to this:

https://docs.mender.io/development/get-started/preparation

Above is the "landing page" if your start position is in the docs. This was at least what I had in mind.

mzedel: it should now be very close to the way the landing page in the docs is set up - the only concern I have might be that users with other devices feel a bit excluded, but with the changed phrasing for the virtual device and the popularity of Rpi we should be good 👍

src/js/components/helptips/onboardingcompletetip.js
```diff
@@ -75,14 +75,13 @@ export class OnboardingCompleteTip extends React.Component {
...
           What next?
           <div>
-            <a href={`https://docs.mender.io/${docsVersion}artifacts/snapshots`} target="_blank">
+            <a href={`https://docs.mender.io/${docsVersion}get-started/deploy-a-system-update`} target="_blank">
               Learn about full-image updates
             </a>{' '}
             or{' '}
-            <a href="https://hub.mender.io/c/update-modules" target="_blank">
+            <a href={`https://docs.mender.io/${docsVersion}get-started/deploy-a-container-update`} target="_blank">
               how to create other kinds of application updates.
```
mirzak, line 85: I think the text at the end should be more explicit as we are now linking to tutorials and not reference text.

Suggestion:

```
What next?

Proceed to one of the following tutorials (listed in recommended order):

1. Deploy a system update
2. Deploy a container update
```

src/js/components/common/dialogs/deviceconnectiondialog.js
mirzak: > the only concern I have might be that users with other devices feel a bit excluded

I understand this concern, but I think bringing custom (non-optimized) devices in to the on-boarding process is to complicated. Our reference environment is based on RPi and QEMU, this you use to test Mender use-cases (system, application, docker updates), it will work because we ensure it works :).

After you are finished with the on-boarding, then it is time to install on your custom device with the knowledge you acquired during onboarding :)

src/js/components/common/dialogs/physicaldeviceonboarding.js
estenberg: I think you are right, @mirzak 
Let's remove the generic one from here. I will update the task & acceptance criteria.

mzedel:
Thanks for going through this @mirzak! Now all this should be adjusted - please let me know if you see anything else 👌

PR #809, opened 2020-06-29
MEN-3685: more user-friendly filter operators

tranchitella:


tranchitella:
> The help tips are great, UX is fine.
> However I was getting bad request errors when the filters were incomplete - to reproduce, select an attribute and >= operator, but don't enter a value yet. The UI makes a call every few seconds which shows an error message to the user, popping up again every time.

May I ask you which error message, exactly? Is it 403 (forbidden) or 400 (bad request)?

michaelatmender:
> May I ask you which error message, exactly? Is it 403 (forbidden) or 400 (bad request)?
400 bad request (I'm testing a different branch now but I'm quite sure it was 400)

tranchitella:
@michaelatmender you are right, I could reproduce the issue.
This commit https://github.com/mendersoftware/gui/pull/809/commits/7114b9473eac709c974e4e02599eaa468524f123 fixes it, but I would appreciate @mzedel feedback as well.

src/js/actions/deviceActions.js
```diff
@@ -105,13 +105,18 @@ const filterProcessors = {
   $lt: val => Number(val) || val,
   $lte: val => Number(val) || val,
   $in: val => ('' + val).split(','),
-  $nin: val => ('' + val).split(',')
+  $nin: val => ('' + val).split(','),
+  $exists: () => 1,
+  $nexists: () => 0
+};
+const filterAliases = {
+  $nexists: () => '$exists',
```
mzedel, line 113: this might be enough:
```suggestion
  $nexists: '$exists'
```
then the access is just `filterAliases[filter.operator]`

src/js/reducers/deviceReducer.js
```diff
@@ -179,9 +179,10 @@ const deviceReducer = (state = initialState, action) => {
       };
     }
     case DeviceConstants.SET_DEVICE_FILTERS: {
+      const filters = action.filters.filter(filter => filter.key && filter.operator && filter.scope && typeof filter.value !== 'undefined');
```
mzedel, line 182: small sidenote: the `typeof ... !== 'undefined'` check most likely is not necessary as `&&` should treat `undefined` as a falsey value - `... && filter.value`

tranchitella: @mzedel yup, but zero is false, isn't it? the "doesn't" exist is "$exists: 0"

src/js/actions/deviceActions.js
tranchitella: Done!

src/js/components/devices/filteritem.js
```diff
@@ -21,6 +21,17 @@ const filterOptionsByPlan = {
...
+  $gte: 'The "greater than or equal" operator can work both on numbers and strings. In the latter case, the operator applies the lexicographical order to the value strings.',
+  $lt: 'The "lesser than" operator can work both on numbers and strings. In the latter case, the operator applies the lexicographical order to the value strings.',
+  $lte: 'The "lesser than or equal" operator can work both on numbers and strings. In the latter case, the operator applies the lexicographical order to the value strings.',
+  $in: 'The "in" operator accepts a list of comma-separated values. It matches if the selected field is equal to one of the specified values.',
+  $nin: 'The "not in" operator accepts a list of comma-separated values. It matches if the selected field\'s value is not equal to any of the specified options.',
+  $exists: 'The "exists" operator matches if the selected field\'s value has a value. No value needs to be provided for this operator.',
+  $nexists: 'The "doesn\'t exist" operator matches if the selected field\'s value has no value. No value needs to be provided for this operator.'
+};
+
```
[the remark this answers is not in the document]
tranchitella: Done

src/js/reducers/deviceReducer.js
mzedel: ah right, when the filters are selected by the user, the value be 1 and only mapped to 0 before being sent to the backend, but when the filters are retrieved from the backend the value will be 0, but defined - sorry I missed that! 👌

PR #813, opened 2020-06-30
Initialize Google Analytics once, at page load, if tracking is on

tranchitella:
Changelog: none

### DIFF
--- a/src/js/tracking.js
+++ b/src/js/tracking.js
@@ -15,8 +15,12 @@ class Tracker {
         }
     }
     initialize(trackingCode) {
+        if (this.initialized) {
+            return false;
+          }
         ReactGA.initialize(trackingCode);
         this.initialized = true;
+        return true;
     }
     pageview(data) {
         if (this.initialized) {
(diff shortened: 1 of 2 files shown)


src/js/tracking.js
```diff
@@ -15,8 +15,16 @@ class Tracker {
...
+        let self = this;
+        return new Promise(function(resolve, reject) {
+            if (self.initialized) {
+                reject(self);
+            } else {
+                ReactGA.initialize(trackingCode);
+                self.initialized = true;
+                resolve(self);
+            }
+        });
```
mzedel, line 27: since there is nothing async happening, wouldn't:
```suggestion
        if (self.initialized) {
          return;
        }
        ReactGA.initialize(trackingCode);
        self.initialized = true;
```
be enough? or does the initialization take some time to complete?

tranchitella: @mzedel You are right, no need to go async. We still need to run the "set" functions conditionally, because we only need them once. I updated the code to return a boolean if the initialization is performed.

PR #815, opened 2020-06-30
MEN-3683: reuse the "create group" modal content as devices help text

tranchitella:
Changelog: none


src/js/components/help/__snapshots__/devices.test.js.snap
```diff
@@ -1,66 +1,141 @@
...
-      href="https://mender.io"
-      target="_blank"
+    />
+  </div>
+  <div
+    className="one-column lightestgrey"
+    style={
+      Object {
+        "flexGrow": 1,
+        "padding": 0,
```
michaelatmender, line 68: The only thing I'd suggest is to add 30px padding here to the lightestgrey section, just for aesthetic reasons

tranchitella: Done!

PR #832, opened 2020-07-02
Exclude testing jobs from pipeline when running for tag saas-*

tranchitella:
Changelog: none

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -19,20 +19,23 @@ include:
 
 test:lint:
   stage: test
+  except:
+    - master
+    - /^saas-[a-zA-Z0-9.]+$/
   image: node:12.14.0-alpine
   script:
     - npm ci
     - npm run lint
-  except:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -27,12 +29,14 @@ test:lint:
     - master
```
mzedel, line 29: here is an `except` definition in the job already, unfortunately they won't get combined automatically

tranchitella: Ops, sorry, I've missed it. Anyway, this change is not for the "staging" pipeline, it is for the `saas-*` tag which is used to re-tag the staging_SHA docker image to `saas-*`, thus no tests are expected to run because we are not running (again) docker-build.

mzedel: that's what I meant, as we won't tag unless the staging build is successful/ the staging image is available - I should have been more explicit with my remark 😅

PR #843, opened 2020-07-06
MEN-3767: preserve the log field, if set, when updating deployments' devices

tranchitella:
Changelog: none

### DIFF
--- a/src/js/actions/deploymentActions.js
+++ b/src/js/actions/deploymentActions.js
@@ -110,10 +110,15 @@ export const getSingleDeploymentStats = id => dispatch =>
     dispatch({ type: DeploymentConstants.RECEIVE_DEPLOYMENT_STATS, stats: res.body, deploymentId: id })
   );
 
-export const getSingleDeploymentDevices = id => dispatch =>
+export const getSingleDeploymentDevices = id => (dispatch, getState) =>
   GeneralApi.get(`${deploymentsApiUrl}/deployments/${id}/devices`).then(res => {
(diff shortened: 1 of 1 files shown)


src/js/actions/deploymentActions.js
```diff
@@ -110,10 +110,15 @@ export const getSingleDeploymentStats = id => dispatch =>
...
-export const getSingleDeploymentDevices = id => dispatch =>
+export const getSingleDeploymentDevices = id => (dispatch, getState) =>
   GeneralApi.get(`${deploymentsApiUrl}/deployments/${id}/devices`).then(res => {
+    const deploymentDevices = (getState().deployments.byId[id] || {}).devices || {};
     const devices = res.body.reduce((accu, item) => {
       accu[item.id] = item;
+      const log = (deploymentDevices[item.id] || {}).log;
+      if (log) {
+        accu[item.id].log = log;
+      }
```
mzedel, line 121: you could most likely get away with:
```suggestion
      accu[item.id] = {
        ...deploymentDevices[item.id],
        ...item
      }
```
which might prevent the tiny decrease in coverage - but it's good as it is already

tranchitella: @mzedel I tried it and it fails, the logs are empty after the first refresh of the deployment devices list.

mzedel: argh, I assumed the log wouldn't be set initially but it seems to be a boolean...

PR #854, opened 2020-07-09
MEN-3802: display attempts and retries in deployment device list, if set

tranchitella:
Changelog: none

### DIFF
--- a/src/js/components/deployments/deployment-report/devicelist.js
+++ b/src/js/components/deployments/deployment-report/devicelist.js
@@ -13,7 +13,7 @@ const stateTitleMap = {
   'already-installed': 'Already installed'
 };
 
-const ProgressDeviceList = ({ created, devices, globalSettings, viewLog }) => {
+const ProgressDeviceList = ({ created, devices, globalSettings, viewLog, retries }) => {
   const deviceList = devices.map((device, index) => {
     var encodedDevice = `id=${device.id}`;
     var id_attribute = device.id || '-';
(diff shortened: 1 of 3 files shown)


src/js/components/deployments/deployment-report/devicelist.js
```diff
@@ -79,11 +84,13 @@ const ProgressDeviceList = ({ created, devices, globalSettings, viewLog }) => {
...
-                {content}
-              </TableCell>
-            ))}
+            {['Device type', 'Current software', 'Started', 'Finished', 'Retries', 'Deployment status', ''].map((content, index) =>
+              content != 'Retries' || retries ? (
+                <TableCell key={`device-list-header-${index + 1}`} style={headerStyle} tooltip={content}>
+                  {content}
+                </TableCell>
+              ) : null
+            )}
```
mzedel, line 93:
```suggestion
            {['Device type', 'Current software', 'Started', 'Finished', '', 'Deployment status', ''].map((content, index) => (
                <TableCell key={`device-list-header-${index + 1}`} style={headerStyle} tooltip={content}>
                  {content}
                </TableCell>
            ))}
```
might be enough

tranchitella: @mzedel with this change, we lose the Retries header, don't we?
We end up with a 1/3 table cell but we don't say what it means.
I'm not sure I get your point.

mzedel: argh right, I misread the condition and just focused on the `content != 'Retries'` - sorry about that!

PR #855, opened 2020-07-09
2.4.x

mzedel:
bringing in MEN-3803, MEN-3804, MEN-3797 - small additional changes for MEN-3803 due to the absence of the deploy to group endpoint...


mzedel:
@tranchitella thank you for checking this! Unfortunately I need another look: to work around the confusion about the filter capabilities I added https://github.com/mendersoftware/gui/pull/855/commits/707514e372d341e7f6d6eed1b3f848abf4cc3d3d - I would vote to remove the scope limitation in master instead of bringing this to master and picking it here, to allow using e.g. regex with the identity attributes - but I'm not sure how fast that would be possible...

tranchitella:
@mzedel it's doable, but I doubt we can do it by today

PR #863, opened 2020-07-17
MEN-3829: do not require marketing consent during signup

tranchitella:
Changelog: none

### DIFF
--- a/src/js/components/user-management/signup-steps/orgdata-entry.js
+++ b/src/js/components/user-management/signup-steps/orgdata-entry.js
@@ -72,8 +72,7 @@ export class OrgDataEntry extends React.Component {
           />
           <FormCheckbox
             id="marketing"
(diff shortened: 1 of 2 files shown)


src/js/components/user-management/signup-steps/orgdata-entry.js
```diff
@@ -72,8 +72,7 @@ export class OrgDataEntry extends React.Component {
           />
           <FormCheckbox
             id="marketing"
-            label="By checking this you agree that we can send you occasional email updates about Mender. You can unsubscribe from these emails at any time *"
-            required={true}
+            label="By checking this you agree that we can send you occasional email updates about Mender. You can unsubscribe from these emails at any time"
```
merlin-northern, line 75: mean nitpick: should there be a full-stop ('.') at the end?

tranchitella: No idea, the previous checkbox has no full-point at the end and the UX screens from Michael do not include the full stop.

PR #864, opened 2020-07-17
MEN-3812, MEN-3308: improvements to filters and dynamic groups

tranchitella:
MEN-3812: allow mixing attributes from different scopes in filters
MEN-3808: allow creation of dynamic groups filtering on identity attributes

### DIFF
--- a/src/less/main.less
+++ b/src/less/main.less
@@ -1093,6 +1093,13 @@ ul.link-list {
 .filter-wrapper {
   padding: 20px;
 }
+.filter-list {
+  margin-bottom: -8px;
+
+  > .MuiChip-root {
+    margin-bottom: 8px;
+  }
+}
 
 .maxTable {
   max-height: 50vh;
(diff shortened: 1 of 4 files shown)


src/less/main.less
```diff
@@ -1093,6 +1093,13 @@ ul.link-list {
 .filter-wrapper {
   padding: 20px;
 }
+.filter-list {
+  margin-bottom: -8px;
```
merlin-northern, line 1097: I have second thought when using negative margins in px. Maybe it will be all right here, but usually I try to find another way. Isn't that going to break RWD a bit?

merlin-northern:
tests ned to be adjusted for the:
```
    +           <div
    +             className="filter-list"
    +           />
```

src/less/main.less
tranchitella: When the filter items wrap to a new line, they need some space between them, so I added a `margin-bottom: 8px`; this makes the container higher by 8px, though, because each element has a margin-bottom. It doesn't matter if we have one or multiple rows, we can safely remove 8px from the bottom to avoid the container to be 8px higher because of the margin-bottom on the latest row of items.

PR #866, opened 2020-07-17
MEN-3418: improved alignment of the status component in deployments' report

tranchitella:
Changelog: none

### DIFF
--- a/src/less/main.less
+++ b/src/less/main.less
@@ -2275,15 +2275,14 @@ span.day {
     display: inline;
   }
   .statusWrapper {
-    max-width: 160px;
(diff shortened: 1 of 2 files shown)


src/less/main.less
```diff
@@ -2296,6 +2295,11 @@ span.day {
       border-bottom: 10px solid transparent;
       border-right: 20px solid @lightestgrey;
     }
+
+    .statusWrapperMessage {
+      display: table-cell;
+      vertical-align: middle;
```
merlin-northern, line 2301: still the only way do to the proper vertical center, after all this time :)

PR #886, opened 2020-08-06
Google Analytics

tranchitella:


src/js/components/app.js
```diff
@@ -27,8 +28,26 @@ const stripePromise = window.mender_environment.stripeAPIKey ? loadStripe(window
...
-    this.props.history.listen(location => {
+    const { trackingCode } = this.props;
+    if (trackingCode) {
+      if (!this.cookies.get('_ga')) {
+        Tracking.cookieconsent(({ trackingConsentGiven }) => {
+          if (trackingConsentGiven) {
+            Tracking.initialize(trackingCode);
+            Tracking.pageview();
+          }
+        });
```
mzedel, line 45: tiny note just as I was expecting the `saveUserSettings` to be passed - the cookieconsent could return a `Promise` and then the handling of the result would be left to the caller, but since the cookieconsent isn't introduced here this is just a sidenote and please feel free to ignore

tranchitella: Refactored to use a Promise

PR #890, opened 2020-08-07
aligned ci image versions + skipped codecov install

mzedel:
Changelog: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,4 +1,4 @@
-image: docker:git
+image: node:14.7.0-alpine
 
 stages:
   - test
@@ -22,7 +22,6 @@ test:lint:
   except:
     - master
     - /^saas-[a-zA-Z0-9.]+$/
-  image: node:12.14.0-alpine
   script:
     - npm ci
     - npm run lint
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -8,21 +8,22 @@ stages:
 
 variables:
   DOCKER_REPOSITORY: mendersoftware/gui
+  NODE_IMAGE_VERSION: 14.7.0-alpine
```
lluiscampos, line 11: I suggest setting `image: node:14.7.0-alpine` above instead (and adding `image: docker:git` to `publish:disclaimer`). The node image is used in all jobs except acceptance testing and the mentioned disclaimer publish, so it makes more sense to set it as the "standard".

.gitlab-ci.yml
```diff
@@ -70,13 +71,11 @@ publish:tests:
...
     - /^saas-[a-zA-Z0-9.]+$/
-  image: node:12.14.0-alpine
+  image: node:$NODE_IMAGE_VERSION
   dependencies:
     - test:unit
-  before_script:
-    - npm install -g codecov
   script:
-    - codecov
+    - npx codecov
```
lluiscampos, line 78: First time I see this `npx`, thanks!

.gitlab-ci.yml
mzedel: A very good point, thanks! changed it as suggested - I guess I got blinded 🙈

PR #895, opened 2020-08-10
Gitlab job to test docs links

tranchitella:
Changelog: none

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -8,7 +8,7 @@ stages:
 
 variables:
   DOCKER_REPOSITORY: mendersoftware/gui
-  DOCS_VERSION: hosted
+  DOCS_VERSION: development
 
 include:
   - project: 'Northern.tech/Mender/mendertesting'
@@ -68,7 +68,7 @@ test:unit:
 test:docs-links:
   stage: test
   except:
-    - /^saas-[a-zA-Z0-9.]+$/
+    - /^(staging|saas-[a-zA-Z0-9.]+)$/
   before_script:
     - apk add --no-cache curl
   script:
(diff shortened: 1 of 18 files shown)


.gitlab-ci.yml
```diff
@@ -64,6 +65,28 @@ test:unit:
...
+    -     error=1
+    -     echo "[ko] $url"
+    -   else
+    -     echo "[ok] $url"
+    -   fi
+    - done
+    - if [ $error -gt 0 ]; then
+    -   exit 1
+    - fi
+  allow_failure: true
```
lluiscampos, line 88: Why you want to allow failure?

lluiscampos: I mean, I know that right now they are some broken links because we have not yet updated mender-docs, but leaving the `allow_failure: true` in the Pipeline I think defeats its purpose.

tranchitella: that's because right now it would fail, and I couldn't come up with a "smart way" to avoid it.
From my point of view, this is more a "safety net" and I'll take care of having that job green before deploying this version to production. Making it conditional requires additional effort, not sure it makes sense. Any suggestion?

lluiscampos: I was not thinking about making it conditional, but just let it fail and force-merge this PR (and others that might happen in exceptional situations). But now that you say so, I had an idea:

We could have this job to use `DOCS_VERSION=development` (and except on branch staging), and a job cloned from this one to run for staging branch with `DOCS_VERSION=hosted`. That will let the gui repo evolve following master docs, but in staging always double check that links are good before production.

If this is a one time thing, and we are removing the `allow_failure` later, then I don't have objections in leaving it as is.

tranchitella: Yup, done!

.gitlab-ci.yml
```diff
@@ -64,6 +65,34 @@ test:unit:
...
+    -   fi
+    - done
+    - if [ $error -gt 0 ]; then
+    -   exit 1
+    - fi
+
+test:docs-links:hosted:
+  extends: test:docs-links
+  only:
+    - /^(staging|saas-[a-zA-Z0-9.]+)$/
```
lluiscampos, line 92: We need to blank out `except` here, otherwise GitLab will compose the job with only + except (from the base job) and nothing will run.
```suggestion
  only:
    - /^(staging|saas-[a-zA-Z0-9.]+)$/
  except: []
```

tranchitella: GitLab CI/CD is tricky, thanks!

PR #896, opened 2020-08-11
MEN-3571 added cancellation option for artifact upload & generation

mzedel:
Changelog: Title


michaelatmender (changes requested):
I suggest adding a "The upload has been cancelled" confirmation snackbar too, but other than that it looks good

michaelatmender (approved):

PR #901, opened 2020-08-14
Fix/2fa endpoint

mzedel:


src/js/api/users-api.js
```diff
@@ -19,7 +19,7 @@ const Api = {
       body = { token2fa: userData.token2fa };
     }
     return request
-      .put(url, body, { ...commonRequestConfig, headers: { ...commonRequestConfig.headers, Authentication: `Bearer ${getToken()}` } })
+      .put(url, body, { ...commonRequestConfig, headers: { ...commonRequestConfig.headers, Authorization: `Bearer ${getToken()}` } })
```
merlin-northern, line 22: do you think it is possible to make e2e test for enabling 2fa? I can provide the token generation code in a language of your choice.

mzedel: yep, now that we have the signup included, adjusting the e2e tests to run against staging and including the 2fa setup there would be the way to go I think...

PR #908, opened 2020-08-17
MEN-3857: improve the create new group modal

tranchitella:


michaelatmender (changes requested):
I commented on MEN-3857: I think this approach re-introduces some UX issues that caused the change initially. See the ticket for my suggestion!

tranchitella:
@mzedel @michaelatmender I reworked the PR to follow the comments and mockup in the ticket.

michaelatmender:
Great, thanks! Just one nitpick left from me - when typing a new group name that also matches the start of an existing group, the 'create "x" group' option doesn't show (see https://tinyurl.com/y5pgxzgw for an example). It's maybe not likely to happen often, but it would be good to display the option consistently if possible.

tranchitella:
> Great, thanks! Just one nitpick left from me - when typing a new group name that also matches the start of an existing group, the 'create "x" group' option doesn't show (see https://tinyurl.com/y5pgxzgw for an example). It's maybe not likely to happen often, but it would be good to display the option consistently if possible.

Updated!

michaelatmender (approved):

PR #911, opened 2020-08-18
extracted stripe script injection to prevent continued stripe check ins

mzedel:
thus limiting the user exposure to the calls to the stripe api (+ prevent the samesite cookie issues from showing up in the browser console)

Changelog: None

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <meta name="viewport content='width=device-width', initial-scale=1" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
     <title>Mender</title>
     <style id="antiClickjack">
       body {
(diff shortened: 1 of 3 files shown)


tranchitella (approved):
This is a bit too much for my React skills. What I could do was testing it works, and I confirm the stripe component is working fine and I could upgrade an account. Each time the browser renders the component, though, I get this warning:
```
main.js:204636 Unsupported prop change on Elements: You cannot change the `stripe` prop after setting it.
```
Not sure it is a problem, I let you decide whether or not, thus I'm pre-approving the PR.

mzedel:
It _should_ result in the same behaviour as the stripe object can't be changed, but still better to fix it - so thank you for catching that! The app component is just initialized once but here I forgot to take that into account.. should be fixed now.
The change itself is just moving the `<Elements />` wrapper and the "inline" import is the async dependency loading, that is possible thanks to the webpack setup. Webpack detects these and extracts them to separate js bundles, that can be loaded on demand - a bit like plugins for js apps...

src/index.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <meta name="viewport content='width=device-width', initial-scale=1" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
```
tranchitella, line 5: Good catch!

tranchitella (approved):

PR #917, opened 2020-08-26
Adjust login and signup for smaller screens

michaelatmender:
Parts were being cut off at smaller screen sizes, and the link at the bottom was hard to click for some users. Also removed a small typo from the code.


mzedel (approved):
Very good! thank you for the attention to detail!

PR #925, opened 2020-09-01
MEN-3314 included timing notification on deployment scheduling

mzedel:
+ reworked scheduling layout slightly to do so

Changelog: None


michaelatmender (changes requested):
Suggest moving the tooltip beside the date/time selection field below, as it is most relevant when the user is picking the specific time ("start immediately" is more self explanatory and doesn't require the tooltip).

Also suggest changing the text a bit to address the timezone question directly, as that is the main reason for confusion. Something like:
"This time is relative to the server only – each device’s time zone will not be taken into account. Devices across different time zones will receive the update at the same time."

michaelatmender (approved):

PR #927, opened 2020-09-03
MEN-3852: rename column retries to attempts in deployment device list

tranchitella:
This commit also updates the content of the table to make it consistent
with the new label.


mzedel (approved):

tranchitella:
@michaelatmender as this is an UX task, I'm waiting for your approval before merging

michaelatmender (approved):
Thanks, this is a good short term fix. I may follow up with more adjustments later, on the deployment creation view.

PR #937, opened 2020-09-08
MEN-3529 force word breakage in device list

mzedel:
forced word breakage in device list to wrap on smaller screens
this might be prevented by diverging CSS defaults on some OS/ browser combinations

Changelog: None


tranchitella (approved):
Renders okay here.

PR #938, opened 2020-09-08
Fix/admin role listing

mzedel:
admins were listed as unable to manage users + the headers in the roles table looked a bit odd due to the headers spanning 2 lines.. also: a tiny fix for the tenant token, which had a differing heading than the other items in the organization settings page

thanks to @oleorhagen for the pointers!


oleorhagen (approved):
Beautiful! 

I tried having a go myself, but it wasn't that easy :see_no_evil:

PR #941, opened 2020-09-09
MEN-3940: hide reset password link if not running on Hosted

tranchitella:
Changelog: none


mzedel (approved):
👍

PR #958, opened 2020-09-18
MEN-3933: remove the loader on error when updating authsets' status

tranchitella:
Changelog: none


mzedel (approved):
the "promise way" of writing this might be to chain a `.catch` or `.finally` call, but the callback should be more than enough

tranchitella:
@mzedel changed to use catch (and fixed linting, sorry)

mzedel:
perfect! Thank you!

PR #960, opened 2020-09-18
added backend call to logout

mzedel:
this will be done either explicit by the user or because of session timeout
so logout will happen regardless of the request success

Changelog: None


tranchitella (approved):
Tested locally using the useradm from my PR, it works fine, thanks!

PR #969, opened 2020-09-24
QA221: switched codecov to coveralls

mzedel:
with the current state of the dependency this is the best that I could get unfortunately...

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -101,7 +101,14 @@ publish:tests:
   dependencies:
     - test:unit
   script:
-    - npx codecov
+    - export CI_BUILD_REF=${CI_COMMIT_SHA}
+    - export CI_BUILD_REF_NAME=${CI_COMMIT_REF_NAME}
+    - export CI_BUILD_NAME=${CI_JOB_NAME}
+    - export CI_BUILD_ID=${CI_JOB_ID}
+    - export CI_MERGE_REQUEST_IID=${CI_COMMIT_BRANCH#pr_}
+    - export COVERALLS_FLAG_NAME=unittests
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -101,7 +101,14 @@ publish:tests:
   dependencies:
     - test:unit
   script:
-    - npx codecov
+    - export CI_BUILD_REF=${CI_COMMIT_SHA}
+    - export CI_BUILD_REF_NAME=${CI_COMMIT_REF_NAME}
+    - export CI_BUILD_NAME=${CI_JOB_NAME}
+    - export CI_BUILD_ID=${CI_JOB_ID}
+    - export CI_MERGE_REQUEST_IID=${CI_COMMIT_BRANCH#pr_}
```
lluiscampos, line 108: Shouldn't it be:
```suggestion
    - export CI_MERGE_REQUEST_ID=${CI_COMMIT_BRANCH#pr_}
```
?

Anyway, this won't be correctly used as we do GitHub PRs and not GitLab MRs

lluiscampos (approved):

lluiscampos:
@mzedel Thank you for taking care of this!

.gitlab-ci.yml
mzedel: the `IID` is read in the coveralls dependency -> https://github.com/nickmerwin/node-coveralls/blob/master/lib/getOptions.js#L90 - and I _think_ this is used to determine the PR this is coming from, but I'll re-run without and see if the information available on coveralls is the same

lluiscampos: It's okey. We can leave it

PR #976, opened 2020-09-28
MEN-3823 credit card updates & updated organization settings

mzedel:
unfortunately the backend apparently often doesn't provide the card data due to the stripe SCA changes - for now I added a note to incentivise card data re-entry, but I'm not sure if we can't also get the data from the existing payment source information... apart from that upgrading the plan through the UI is only possible for trial tenants so far, thus this is deferred to email for now...


michaelatmender (changes requested):
I think this all looks good - the only thing I noticed is that the payment card settings still show on Enterprise, but Enterprise users pay via invoice rather than card. So it should not be an option for Enterprise plan users. 

(Bear in mind it could be the case I've made an error when testing it locally!)

mzedel:
Thanks for looking into this @michaelatmender! I changed the enterprise handling, but included the device limit progress, as I thought this would be relevant in any case... I hope that's ok!

michaelatmender (approved):

PR #979, opened 2020-09-30
switched count calls to no longer use deviceauth due to inconsistent counts

mzedel:
I hope this change doesn't increase the load, but it should help avoid some of the inconsistencies that are reported for the frontend

workaround for men-3831
Changelog: None


tranchitella (approved):
Looks good to me, and I believe it is a good idea

PR #980, opened 2020-09-30
Feature/device url filter tracking

mzedel:


src/js/components/devices/devices.js
```diff
@@ -54,26 +84,36 @@ export class Devices extends React.Component {
...
+        .some(attribute => attribute)
+    ) {
+      const { filters, groupName } = convertQueryToFilterAndGroup(location.search, filteringAttributes);
       if (groupName) {
-        this.props.selectGroup(groupName, filters);
+        selectGroup(groupName, filters);
       } else {
-        this.props.setDeviceFilters(filters);
+        setDeviceFilters(filters);
       }
```
tranchitella, line 117: I'm not sure I understand why you work on location.search here, instead of location.hash. What I see is that this block is equal to the one in componentDidMount, apart from working on location.search instead of match.params.filters. Maybe we can refactor them and share the implementation across the two different methods? Does it make sense?

mzedel: Good point and I wanted to unify the two, but had to keep the `match.params.filter` one for compatibility with the current link style `/devices/<deviceId>` and the search one for the `/devices?id=<deviceId>` one that is unfortunately unsupported by react-router... I'll refactor to extract more shared code though👌

tranchitella: Ah, I see, now I understand. No need to refactor if you think it's useless, the rest of the PR looks good and works fine here.

tranchitella (approved):

PR #982, opened 2020-10-01
Set up code scanning

mzedel:
Changelog: None


tranchitella (approved):

tranchitella:
@mzedel I trust you on this :)

mzedel:
that should activate checks like they run in the deployments service already (e.g. here https://github.com/mendersoftware/deployments/pull/484/checks) - for some reason the auto setup detects python as a repo language (even though there should not be any python in here for at least 2 years). with this the scanning should be `js` only and the checks on PRs should be activated through the code in the repo :).

PR #1000, opened 2020-10-12
MEN-4041 parse audit log entries

mzedel:


tranchitella (approved):

mzedel:
Thanks for the heads up @alfrunes! I'll adjust this PR accordingly... and won't complain about the lesser amount in parsing work 😄.

Treating the `change` property as an object was more due to the example in the corresponding ticket and thus only results in a "very low effort" object parsing + trying to make the most of the object to format the content... (+ the change in the rendering resulted in slightly easier formatting)...

tranchitella (approved):

PR #1002, opened 2020-10-12
MEN-3922 onboarding refactoring

mzedel:


michaelatmender (commented):
Testing this again, I no longer see the successful demo deployment tooltip (with the link to the demo web app) so I don't get any further prompts after that first deployment.
From settings API:
`"onboarding":{
	"complete":false,
	"showTips":true,
	"deviceType":"qemux86-64",
	"approach":"virtual",
	"artifactIncluded":null,
	"progress":"onboarding-start",
	"address":"http://localhost:85",
	"showArtifactCreation":false
}`

mzedel:
Thanks for taking another look at it!
Do you mean the tooltip no longer showing up when accepting the demo device or it doesn't continue once you pass that tooltip? and if it's the latter, was this the last call to/ from the settings endpoint?

michaelatmender (approved):

PR #1004, opened 2020-10-14
added diff like coloring to audit log change details

mzedel:
Changelog: None


tranchitella (approved):
Pre-approving, but the snap needs to be updated to have unit tests green.

PR #1022, opened 2020-10-26
MEN-3510 use attribute retrieval endpoint

mzedel:
this relies on the changes in #1002 to line up the app initialisation steps
main change is in f76bc0dee687b55af08780f263b520172cfd54a5

### DIFF
--- a/src/js/actions/deviceActions.js
+++ b/src/js/actions/deviceActions.js
@@ -3,7 +3,7 @@ import pluralize from 'pluralize';
 import { setSnackbar } from '../actions/appActions';
 import GeneralApi, { headerNames } from '../api/general-api';
 import * as DeviceConstants from '../constants/deviceConstants';
-import { deriveAttributesFromDevices, duplicateFilter, getSnackbarMessage, mapDeviceAttributes, preformatWithRequestID } from '../helpers';
+import { getSnackbarMessage, mapDeviceAttributes, preformatWithRequestID } from '../helpers';
 
 // default per page until pagination and counting integrated
 const defaultPerPage = 20;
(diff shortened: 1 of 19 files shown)


src/js/actions/deviceActions.js
```diff
@@ -549,6 +536,24 @@ export const getAllDevicesByStatus = status => (dispatch, getState) => {
   return getAllDevices();
 };
 
+export const getDeviceAttributes = () => dispatch =>
+  GeneralApi.get(`${inventoryApiUrlV2}/filters/attributes`).then(({ data = [] }) => {
```
tranchitella, line 540: We return up to 500 attributes here, not sure you need a limit in the UI... maybe not,

src/js/helpers.js
```diff
@@ -318,35 +318,6 @@ export const mapDeviceAttributes = (attributes = []) =>
...
-  // sort in reverse order, to have most common attribute at the top of the select
-  const inventoryAttributes = Object.entries(availableAttributes.attributes)
-    .sort((a, b) => b[1] - a[1])
-    .map(a => a[0]);
-  const identityAttributes = Object.entries(availableAttributes.identity_data)
-    .sort((a, b) => b[1] - a[1])
-    .map(a => a[0]);
-  return { identityAttributes, inventoryAttributes };
-};
-
```
tranchitella, line 349: So nice to get rid of this custom logic in the UI, yay!

tranchitella (approved):
Great job!

src/js/actions/deviceActions.js
mzedel: a valid point indeed - we once had a limit, but soon after got feedback that there are some attributes missing 🙈... while the list of identity attributes is still limited, the filters might show all... the component we use there is definitely capable of a few hundred items though so we should be good 👌

PR #1025, opened 2020-10-29
Fix/upgrade+virtual device

mzedel:


src/js/helpers.test.js
```diff
@@ -120,6 +121,24 @@ systemctl restart mender-client'`
...
+    expect(code).not.toMatch(/\{([^}]+)\}/);
+    code = getDemoDeviceCreationCommand(token);
+    expect(code).not.toMatch(/\{([^}]+)\}/);
+  });
+  it('should return a sane result', () => {
+    let code = getDemoDeviceCreationCommand();
+    expect(code).toMatch('./demo --client up');
+    code = getDemoDeviceCreationCommand(token);
+    expect(code).toMatch(
+      `TENANT_TOKEN='${token}'\ndocker run -it -p 85:85 -e SERVER_URL='https://localhost' \\\n-e TENANT_TOKEN=$TENANT_TOKEN mendersoftware/mender-client-qemu:latest`
```
merlin-northern, line 137: the port forward `85:85` is present only here, is that on purpose?

merlin-northern: ah ok I see it in 
```${onboardingReducerState.demoArtifactPort}:${
        onboardingReducerState.demoArtifactPort
      }
```

mzedel: in the tests it is static anyways, so I can expect the `85` 👌

merlin-northern (approved):
lgtm!

PR #1048, opened 2020-11-11
fixed font inclusion on webpack build

mzedel:
Changelog: None


tranchitella (approved):
Works fine, now, thanks!

PR #1065, opened 2020-11-19
Fix URL in billing/card/confirm API call when confirming CC change

tranchitella:
Changelog: none


mzedel (approved):
Thanks for solving this!

PR #1067, opened 2020-11-19
made card update interaction react to update success

mzedel:
- should now switch back to just showing card details

Changelog: None


tranchitella (approved):
Yup!

PR #1069, opened 2020-11-19
Fix/deployment widgets

mzedel:


src/js/components/dashboard/deployments.js
```diff
@@ -55,11 +55,14 @@ export class Deployments extends React.Component {
 
   getDeployments() {
     const self = this;
-    return self.props
-      .getDeployments(1, 20)
+    const updateRequests = Object.keys(DEPLOYMENT_STATES).map(status =>
+      self.props.getDeploymentsByStatus(status, 1, status === DEPLOYMENT_STATES.finished ? 100 : 20)
```
tranchitella, line 59: I'm not sure I understand why per_page is 100 for finished, 20 for the others...

tranchitella (approved):
Looks good and works fine here. I have one question for my learning, though.

src/js/components/dashboard/deployments.js
alfrunes: Also, there is a `X-Total-Count` header returned from this endpoint (though hidden from the API docs).

mzedel: The 20 for the others was just the fallback to the default as I wasn't 100% sure what else the deployment counts for pending/ inprogress would be used for, however the finished deployments had to be retrieved in full, as they used to be counted in the corresponding widget based on their finished date and the last user session (since we didn't have the `X-Total-Count` then and should reduce the number of api calls). I've updated the call & the past deployments widget to just rely on the api response here.
Thanks to both of you for looking at this and chiming in - otherwise I would have just kept the local counting, but this feels much better!

tranchitella (approved):

PR #1082, opened 2020-11-25
forced rerender when stripe dependencies are loaded

mzedel:
this is to ensure the stripe elements are also visible on first visit to the settings
that is not the case on slow connections
Changelog: None


tranchitella (approved):
This was annoying, indeed. Nice catch!

PR #1123, opened 2020-12-07
Add Dec discount offer to hosted header

michaelatmender:
Discount offer should show for hosted Trial users only, until the end of December


src/js/components/settings/settings.js
```diff
@@ -106,7 +106,8 @@ export const Settings = ({ allowUserManagement, currentUser, hasMultitenancy, hi
 
 const mapStateToProps = state => {
   const currentUser = state.users.byId[state.users.currentUser];
-  const { plan = 'os', trial = false } = state.organization.organization;
+  const plan = 'os';
+  const trial = true;
```
mzedel, line 110: this is probably left over I think...

src/js/components/settings/settings.js
```diff
@@ -50,12 +50,12 @@ export const Settings = ({ allowUserManagement, currentUser, hasMultitenancy, hi
...
     upgrade: {
-      admin: false,
-      enterprise: false,
+      admin: true,
+      enterprise: true,
       multitenancy: true,
       trial: true,
-      userManagement: false,
-      component: <Upgrade history={history} />,
+      userManagement: true,
```
mzedel, line 57: I could see this only being allowed for admins (though I personally would accept an upgrade from every user 😅) but `enterprise` and `userManagement` are probably not required right?

src/less/main.less
```diff
@@ -21,7 +21,7 @@ body {
   display: grid;
   height: 100%;
   grid-template-columns: 200px 1fr;
-  grid-template-rows: 57px 1fr;
+  grid-template-rows: min-content 1fr;
```
mzedel, line 24: very nice 👍 - I still haven't internalised these...

mzedel (changes requested):
nicely done, just two things that might be still in from the development...

src/js/components/settings/settings.js
michaelatmender: Yes, this and the previous comment are leftovers. Careless committing from me, sorry!

src/js/components/settings/settings.js
```diff
@@ -106,7 +106,6 @@ export const Settings = ({ allowUserManagement, currentUser, hasMultitenancy, hi
 
 const mapStateToProps = state => {
   const currentUser = state.users.byId[state.users.currentUser];
-  const { plan = 'os', trial = false } = state.organization.organization;
```
mzedel, line 109: this is still required 😉

mzedel (changes requested):
that was a tiny bit too much... sorry about the back and forth, I'm a bit too deep in the testing branch right now, otherwise I'd push it myself 😟

mzedel (approved):
👍

michaelatmender:
@tranchitella do you have the ability to merge this? The ci/gitlab status seems fine on https://gitlab.com/Northern.tech/Mender/gui/-/pipelines/227214309 but is not reporting success here

PR #1137, opened 2020-12-15
CI: Allow test:staging-deployments job to fail

alfrunes:
The job has known issues, let's keep monitoring the job until the issue
is resolved.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -120,6 +120,8 @@ test-prep:
     expire_in: 30 days
 
 test:staging-deployment:
+  # Allow failure until timing issues are resolved
+  allow_failure: true
   image:
     name: cypress/included:5.4.0
     # cypress/included images have entrypoint set to "cypress"


.gitlab-ci.yml
```diff
@@ -120,6 +120,8 @@ test-prep:
     expire_in: 30 days
 
 test:staging-deployment:
+  # Allow failure until timing issues are resolved
+  allow_failure: true
```
lluiscampos, line 124: @mzedel @alfrunes Please let's create a JIRA ticket (and add a reference here in the comment) before merging the PR.

lluiscampos (changes requested):
Good initiative, thanks. But see comment below.

mzedel (changes requested):
Honestly I'm not a fan of that setting for an entire test job, since this will hide any and all failures on staging... while I know that the morning status notification gets additional text, the nature of the failure should stay in there...
Preferably the failing test should be fixed and only the flaky ones allowed to fail/ repeated. Creating a ticket for that would be helpful, as I don't think I'll get to anything else than tests this week...

PR #1139, opened 2020-12-16
MEN-4046: use the /me API end-point to update user's own profile

tranchitella:
Changelog: none

Depends on:
* https://github.com/mendersoftware/useradm/pull/209
* https://github.com/mendersoftware/useradm-enterprise/pull/171


mzedel (approved):
🧐 I was not aware that we have a `/me` endpoint - apparently the current useradm is not yet aware either 😅

tranchitella:
@mzedel I introduced the /me end-point because we need a unique URI to allow it with RBAC. We cannot write a URI-based rule in the RBAC role if it contains the ID.

PR #1143, opened 2020-12-17
fixed cookie paths to work with ui redirect to subpath

mzedel:
+ properly clean up on logout

Changelog: Title

### DIFF
--- a/src/js/actions/userActions.js
+++ b/src/js/actions/userActions.js
@@ -35,13 +35,13 @@ export const loginUser = userData => (dispatch, getState) =>
       if (!token) {
         return;
       }
-      let options = {};
+      let options = { sameSite: 'strict', path: '/' };
       if (userData.hasOwnProperty('noExpiry')) {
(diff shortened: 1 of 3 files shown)


src/js/actions/organizationActions.js
```diff
@@ -25,7 +25,7 @@ export const createOrganizationTrial = data => dispatch =>
       }
     })
     .then(res => {
-      cookies.set('JWT', res.text, { maxAge: 900, sameSite: 'strict', path: '/' });
+      cookies.set('JWT', res.text, { maxAge: 900, sameSite: 'strict', path: '/ui' });
```
tranchitella, line 28: We cannot do this because it breaks deviceconnect and websockets: when we connect to the API end-point which establishes a WebSocket connection, we cannot set additional headers because of limitations in the browser's implementation of WebSockets. We use Cookies to transfer the JWT token, but if we set the Cookie path to /ui it won't be transferred when calling /api/ end-points.

tranchitella (changes requested):

tranchitella (approved):

PR #1145, opened 2020-12-18
MEN-4259: update demo bash snippet with the mender-shell installation

tranchitella:
Changelog: none

### DIFF
--- a/src/js/components/help/application-updates/__snapshots__/mender-deb-package.test.js.snap
+++ b/src/js/components/help/application-updates/__snapshots__/mender-deb-package.test.js.snap
@@ -148,7 +148,14 @@ sudo bash -c 'DEVICE_TYPE="generic-armv6" && \\
 mender setup \\
   --device-type $DEVICE_TYPE \\
   --quiet --demo  && \\
-systemctl restart mender-client'
+systemctl restart mender-client && \\
+(cat &gt; /etc/mender/mender-shell.conf &lt;&lt; EOF
+{
+  "ServerURL": "http://localhost",
+  "User": "pi"
+}
+EOF
+) && systemctl restart mender-shell'
 
     </span>
   </div>
--- a/src/js/helpers.js
+++ b/src/js/helpers.js
@@ -458,7 +458,14 @@ TENANT_TOKEN="${token}" && \\`
 mender setup \\
   --device-type $DEVICE_TYPE \\
 ${connectionInstructions} && \\
-systemctl restart mender-client'
+systemctl restart mender-client && \\
+(cat > /etc/mender/mender-shell.conf << EOF
+{
+  "ServerURL": "${document.location.origin}",
+  "User": "pi"
+}
+EOF
+) && systemctl restart mender-shell'
 `;
   return codeToCopy;
 };
--- a/src/js/helpers.test.js
+++ b/src/js/helpers.test.js
@@ -84,7 +84,7 @@ describe('stringToBoolean function', () => {
 
 describe('getDebInstallationCode function', () => {
   it('should not contain any template string leftovers', () => {
-    expect(getDebInstallationCode()).not.toMatch(/\{([^}]+)\}/);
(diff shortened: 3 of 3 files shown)


src/js/components/help/application-updates/__snapshots__/mender-deb-package.test.js.snap
```diff
@@ -148,7 +148,15 @@ sudo bash -c 'DEVICE_TYPE="generic-armv6" && \\
 mender setup \\
   --device-type $DEVICE_TYPE \\
   --quiet --demo  && \\
-systemctl restart mender-client'
+systemctl restart mender-client && \\
+(cat &gt; /etc/mender/mender-shell.conf &lt;&lt; EOF
+{
+  "ServerURL": "https://hosted.mender.io",
+  "ShellCommand": "/bin/sh",
```
lluiscampos, line 155: This file is not required so I think we should drop it.

lluiscampos (approved):

src/js/components/help/application-updates/__snapshots__/mender-deb-package.test.js.snap
tranchitella: I removed the ShellCommand, let's use the default, you are right!

src/js/helpers.js
```diff
@@ -458,7 +458,14 @@ TENANT_TOKEN="${token}" && \\`
 mender setup \\
   --device-type $DEVICE_TYPE \\
 ${connectionInstructions} && \\
-systemctl restart mender-client'
+systemctl restart mender-client && \\
+(cat > /etc/mender/mender-shell.conf << EOF
+{
+  "ServerURL": "https://hosted.mender.io",
```
mzedel, line 464: This might be `"ServerURL": "${document.location.origin}"` - the file itself is likely not problematic for on-prem OS installations, but I guess on-prem enterprise could have deviceconnect installed as well and could then reuse the snippet...

mzedel (changes requested):
just a minor thing to allow this to work on-premise as well - apart from that: is mender-shell included in the install from `https://d1b0l86ne08fsf.cloudfront.net/master/dist-packages/debian/armhf/mender-client_master-1_armhf.deb` @lluiscampos? - if not then please don't merge before the get.mender.io script is close to being released... also the reverts for the get.mender.io change need to be brought back to master as well...

lluiscampos:
> just a minor thing to allow this to work on-premise as well - apart from that: is mender-shell included in the install from `https://d1b0l86ne08fsf.cloudfront.net/master/dist-packages/debian/armhf/mender-client_master-1_armhf.deb` @lluiscampos?

No, `mender-shell` binary is not installed with that package.

tranchitella:
@mzedel master is using get.mender.io, and this is what we want to release in staging soon (early next week?)
I believe we should merge it as it is, I reverted the get.mender.io in staging to release the bugfix yesterday
We can create a feature branch, but in any case we need these commits in staging to test it there.

src/js/helpers.js
tranchitella: Updated, thanks!

mzedel:
if you need to really see this locally, run the gui from the PR and activate the hosted features by putting the following in your `docker-compose.override.yml`:
```
  mender-gui:
    image: mendersoftware/gui:pr
    environment:
      MENDER_HOSTED: 1
```

the `get.mender.io` script won't work on staging, since we can't modify the hosted location in the as it is set with `--hosted-mender`... so unless we get this into the client, there is no real benefit over running it locally - it does however block us from releasing master as it is... afaik the release of the get.mender.io functionality is coupled to the next full release and that does not look likely to happen before January...

tranchitella:
@mzedel @lluiscampos I'm closing this PR, launched a new one to revert all mender-shell / get.mender.io changes in master and then I'll start a new feature branch with those commits + the ones from this PR:
https://github.com/mendersoftware/gui/pull/1146

PR #1158, opened 2020-12-23
Remote Terminal: do not send the userId in the "new" shell message

tranchitella:
The deviceconnect service will enrich the message we send over websocket
adding the userID from the JWT token as `user_id` property.  Thus, we
don't need to add it explicitly to the body of the message "new", shell
protocol.

Changelog: none


merlin-northern (approved):
lgtm after merge I will test again with all components.

PR #1163, opened 2020-12-28
MEN-4172: dynamic resizing of the terminal window

tranchitella:
Changelog: none

Depends on: https://github.com/mendersoftware/mender-connect/pull/2


mzedel (changes requested):
unfortunately the interval based solution might lead to memory leaks, as the size might be tried to be set while the component is already or is in the process of getting unmounted (and I guess it would mostly remain unchanged)...

using effects I think the closest might be
```
  const [size, setSize] = useState({ height: window.innerHeight, width: window.innerWidth });
  useLayoutEffect(() => {
    const updateSize = () => {
      setSize({ height: window.innerHeight, width: window.innerWidth });
    }
    window.addEventListener('resize', updateSize);
    updateSize();
    return () => window.removeEventListener('resize', updateSize);
  }, []);
  useEffect(() => {
    fitAddon.fit();
    const newDimensions = fitAddon.proposeDimensions();
    if (newDimensions.rows != dimensions.rows || newDimensions.cols != dimensions.cols) {
      dimensions = newDimensions;
      //
      const proto_header = {
        proto: MessageProtocolShell,
        typ: MessageTypeResize,
        sid: sessionId,
        props: { 'terminal_height': dimensions.rows, 'terminal_width': dimensions.cols }
      };
      const msg = { hdr: proto_header };
      const encodedData = MessagePack.encode(msg);
      socket.send(encodedData);
    }
  }, size);
```

using the resize listener the dimension change condition can most likely be omitted - but I haven't tried any of that so far, so if you know about anything that would prevent the layout hook/ listener based solution from failing - please let me know... I could imagine the listener from maybe getting called too frequent/ not debounced on window size change, which might require debouncing or sticking with the interval based solution and encapsulating the intervals setup and cleanup in a `useEffect`... probably something like:
```
useEffect(() => {
  resizeInterval = setInterval(() => { ... });
  return () => {
    clearInterval(resizeInterval);
  }
}, [])
```

tranchitella:
@mzedel does this https://github.com/mendersoftware/gui/pull/1163/commits/dcd57241abfa488981320e9b6fe1b84007b8fe71 avoids the memory leak? I did not know useEffect supported a clean-up function you can return from it, sorry.

mzedel (approved):
that should work fine I guess 👍 thanks a lot!

PR #1184, opened 2021-01-06
MEN-4315: handle error messages when starting a new remote terminal

tranchitella:
Changelog: none


alfrunes (approved):
LGTM. I also verified that it works.

PR #1185, opened 2021-01-07
Fix: onboarding shall use get.mender.io from 2.6.0 onwards

tranchitella:
Changelog: none


merlin-northern (approved):
lgtm lets get it to staging and wait for Manuel's review

PR #1188, opened 2021-01-07
MEN-4260 adding e2e tests for remote terminal setup

mzedel:


tranchitella (approved):
This is outside my competences, but I could read the changes and they make sense.
The pipeline is green, thus... I believe we can merge!

kacf:
Nice!

PR #1193, opened 2021-01-08
MEN-4305: send shell pong messages in response to shell ping messages

tranchitella:
Changelog: none

### DIFF
--- a/src/js/components/devices/terminal.js
+++ b/src/js/components/devices/terminal.js
@@ -14,10 +14,13 @@ import { setSnackbar } from '../../actions/appActions';
 //     for the description of proto_header and the consts
 // *Note*: this needs to be aligned with mender-connect and deviceconnect.
 const MessageProtocolShell = 1;
-const MessageTypeShell = 'shell';
+
 const MessageTypeNew = 'new';
-const MessageTypeStop = 'stop';
+const MessageTypePing = 'ping';
+const MessageTypePong = 'pong';
 const MessageTypeResize = 'resize';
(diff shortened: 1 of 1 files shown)


src/js/components/devices/terminal.js
```diff
@@ -118,6 +131,19 @@ export const Terminal = props => {
...
+        } else if (obj.hdr.typ == MessageTypePing) {
+          if (healthcheckTimeout) {
+            clearTimeout(healthcheckTimeout);
+          }
+          const proto_header = { proto: 1, typ: MessageTypePong, sid: sessionId, props: null };
+          const msg = { hdr: proto_header, body: null };
+          const encodedData = MessagePack.encode(msg);
+          socket.send(encodedData);
+          //
+          var timeout = parseInt((obj.hdr.props || {}).timeout);
```
mzedel, line 143: In what case would the timeout not be set? should the connection maybe still set a timeout in that case? I'm not steeped in the protocol details or the entire interactions between the device/ backend/ gui - so I'll appreciate any details possible!

mzedel (approved):

src/js/components/devices/terminal.js
tranchitella: If the mender-connect disables it; it is not supported as of today (the interval is hard-coded), but who knows in the future... in this case, the UI still replies to "ping" requests with a pong, but does not set any client-side timeout. To be honest, the important thing is being sure the user (UI) is still around, the other way around is a "nice to have" feature.

PR #1194, opened 2021-01-08
MEN-4293 extended device software column space

mzedel:
Screenshot 2021-01-08 at 16 17 16
+ based devicelist column styling on grid to ease per column scaling

Changelog: None


michaelatmender (approved):
Looks good, thanks!

I noticed that [resizable columns](https://material-ui.com/components/data-grid/columns/#column-resizing) are possible now, so perhaps in the near future we can introduce them and let the user choose.

mzedel:
Good point, I'll take a look - might be a good opportunity when we switch to the slide out panel (as the device list will be changed in that case anyway)

PR #1205, opened 2021-01-11
Fix/user login + auditlog download

mzedel:


tranchitella (approved):
LGTM. I've also tested it locally and it works great!

PR #1206, opened 2021-01-11
Do not leak websockets on errors

tranchitella:
Changelog: none

### DIFF
--- a/src/js/components/devices/terminal.js
+++ b/src/js/components/devices/terminal.js
@@ -101,14 +101,17 @@ export const Terminal = props => {
       }, 1000);
     };
 
+    var snackbarAlreadySet = false;
     var healthcheckHasFailed = false;
     socket.onclose = event => {
-      if (healthcheckHasFailed) {
+      if (!snackbarAlreadySet && healthcheckHasFailed) {
         setSnackbar('Health check failed: connection with the device lost.', 5000);
-      } else if (event.wasClean) {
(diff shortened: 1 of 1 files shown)


src/js/components/devices/terminal.js
```diff
@@ -136,12 +139,17 @@ export const Terminal = props => {
...
             setSnackbar('Error: ' + byteArrayToString(obj.body), 5000);
+            snackbarAlreadySet = true;
+            socket.close();
             onCancel();
           } else {
             setSessionId(obj.hdr.sid);
           }
         } else if (obj.hdr.typ === MessageTypeShell) {
           term.write(byteArrayToString(obj.body));
+        } else if (obj.hdr.typ === MessageTypeStop) {
```
alfrunes, line 150: Maybe this should be a "catch all" (else) case?

alfrunes (approved):
I can indeed verify that it works. Well done!

src/js/components/devices/terminal.js
tranchitella: uhm, not really: we are at risk of closing the socket if we do not recognize a message...

tranchitella: Maybe this one is safer: https://github.com/mendersoftware/gui/pull/1206/commits/78a203f06a82fd61fa43cef92b2e1e5f85375c34

PR #1212, opened 2021-01-14
We introduced "retry" as a concept after this label, which was a

estenberg:
short-term solution. Recreate deployment should better explain
what it does vs. today's per-device automatic retry.

### DIFF
--- a/src/js/components/deployments/deployment-report/overview.js
+++ b/src/js/components/deployments/deployment-report/overview.js
@@ -182,7 +182,7 @@ export const DeploymentOverview = ({
               placement="bottom"
             >
               <Button color="secondary" startIcon={<RefreshIcon fontSize="small" />} onClick={onRetryClick} style={{ alignSelf: 'baseline', marginTop: 45 }}>
-                Retry deployment?
+                Recreate deployment?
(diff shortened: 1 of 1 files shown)


estenberg:
Screenshot from 2021-01-14 09-57-49

estenberg:
Hopefully simple change to reword "Retry deployment" in this screen to "Recreate deployment".

It did not seem to be mentioned anywhere else nor have any dependencies, so hoping this is as simple as it looks. :)

mzedel:
Ah right, that should clear up some of the confusion around the retries 👍
Unfortunately the license page in the docs site is currently having problems (fix is in https://github.com/mendersoftware/mender-docs/pull/1286), due to the link checking in the UI, the pipeline will likely only be able to succeed tomorrow morning (I think).

One tiny thing though: the `Changelog` tag is missing in your commit, I'll happily add it if you haven't done so before the docs issue is fixed...

estenberg:
Great, thank you @mzedel !
I have fixed the changelog, it is too long between I commit these days so I forgot :)

mzedel (approved):

PR #1216, opened 2021-01-15
MEN-4339 socket independence to allow disconnection confirmation

mzedel:
This turned out to be a bit more involved than I expected and there were several approaches I tried - and I'm still not sure if this fixes all possible scenarios - so please let me know if there is something I missed. The refactorings were a bit a byproduct of the different approaches 🙈.


mzedel:
@tranchitella is this ready to be merged?

tranchitella (approved):
It works, but to be honest I still had issues in reconnecting but it could be because of my setup. It's an improvement, in any case, so let's merge it. I'll test it again together with the latest mender-connect and see what happens.

mzedel:
that's what I was afraid of 🙈 - please let me know what you come across!

PR #1217, opened 2021-01-16
fixed onboarding tips dismissal not being saved right away

mzedel:
Changelog: Title

### DIFF
--- a/src/js/actions/onboardingActions.js
+++ b/src/js/actions/onboardingActions.js
@@ -55,7 +55,7 @@ export const getOnboardingState = () => (dispatch, getState) => {
   const progress = Object.keys(onboardingSteps).findIndex(step => step === onboardingStepNames.ARTIFACT_CREATION_DIALOG);
   const currentProgress = Object.keys(onboardingSteps).findIndex(step => step === onboardingState.progress);
   onboardingState.showArtifactCreation = Math.abs(currentProgress - progress) <= 1;
-  if (onboardingState.showArtifactCreation) {
+  if (onboardingState.showArtifactCreation && !onboardingState.complete && onboardingState.showTips) {
     dispatch(setShowCreateArtifactDialog(true));
     onboardingState.progress = onboardingStepNames.ARTIFACT_CREATION_DIALOG;
     // although it would be more appropriate to do this in the app component, this happens here because in the app component we would need to track
--- a/src/js/actions/onboardingActions.js
+++ b/src/js/actions/onboardingActions.js
@@ -119,7 +119,11 @@ export const setOnboardingCanceled = () => dispatch =>
     Promise.resolve(dispatch(setShowOnboardingHelp(false))),
     Promise.resolve(dispatch(setShowDismissOnboardingTipsDialog(false))),
     Promise.resolve(dispatch({ type: OnboardingConstants.SET_ONBOARDING_COMPLETE, complete: true }))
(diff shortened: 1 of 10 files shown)


tranchitella (changes requested):
This was annoying me a lot: my brain learned to automatically click "dismiss" without even noticing it!

I have an issue, though: when I dismiss the onboarding tips, the menu item in the dropdown (top-right) still says "Hide help tooltips". Clicking once, it switches to "Show help tooltips". Clicking it again, it finally shows the tooltips, but if I refresh the page, it goes back to the "Hide help tooltips" and the onboarding doesn't continue.

What do you think?

mzedel:
Thanks for spotting this! That actually refers to a different set of tips (the maroon ones as can be seen e.g. on the release page) and they can be toggled through the menu... I adjusted the PR to ensure they are also toggled on onboarding dismissal 👌  However I'm reluctant to toggle the entire onboarding with this, as this would require more work to ensure useful behaviour also for more advanced users (e.g. having 10 devices accepted, but still just a single artifact and deployment)...

tranchitella (changes requested):
I'm sorry, @mzedel, but it doesn't work very well here. See the attachment:
image

I started onboarding, then dismissed. Refreshed several times, moved around, enabled tooltips again (using the top-right menu), refreshed again a couple of times and after a while multiple steps of the onboarding started appearing together. At this point, i get all the onboarding tips wherever I go in the UI, but please note I haven't accepted any device so far.

tranchitella:
I'm really sorry, but I cannot get this PR working as I would expect:
- on refresh, I get onboarding tooltips on other next reload (yes, no, yes, no, yes, no, and so on)
- once I dismiss the onboarding tooltips, I cannot get them back in any way, not even logging out & in or clicking Hide/Show help tooltips

I know I'm pedantic, but I have the fear we won't show the onboarding tips to users...

mzedel:
Thanks a lot for being pedantic!
The second point is as intended, as the help tooltips refer only to the few purple tips we have in the UI (e.g. for groups, authsets and artifacts explanations) - these should show up again when clicking in the menu.... once you dismiss the onboarding tooltips, the onboarding will be treated as cancelled (also in the tracking) - I would like to move that to the settings to allow returning to the onboarding, but that requires several additional adjustments in the onboarding progress tracking (which e.g. relies heavily on the artifact, device & deployment count)... I'll adjust the text in the dialog to clarify that...
The first point however shouldn't happen - it could take a few seconds for the settings to be retrieved and the onboarding state being retrieved but alternating on reload is very much unexpected 😟 -> I'll look into that

mzedel:
@tranchitella could you take another look? it should now work + have the adjustments in for the onboarding tooltips not showing up since the recent layout changes in master...

since this is closely related to the onboarding I included 953857ab568c185778e0138717e7e47c7f439c13 - @lluiscampos could you check if this is how you imagined it? to activate this on staging, the integration version on the staging container should be set to `next` and so no additional changes should be required in the gui code

tranchitella (approved):
@mzedel it works fine, now. Thanks for your patience.

lluiscampos:
> since this is closely related to the onboarding I included [f806934](https://github.com/mendersoftware/gui/commit/f8069340f9cd449f5ddf5f885ee47c8ba065ff3a) - @lluiscampos could you check if this is how you imagined it? to activate this on staging, the integration version on the staging container should be set to `next` and so no additional changes should be required in the gui code

I think so :+1:

merlin-northern:
I did a small round of testing with the onboaring on staging -- seems fine.

PR #1218, opened 2021-01-17
MEN-4368: rename "Device group" column to "Target device(s)"

tranchitella:
Changelog: none


mzedel (commented):
Yeah the deployment naming is indeed a bit tricky, as we use the deployment name for the group/ device id tracking, sounds reasonable to me! But to be sure I'll defer to @michaelatmender on this one (as I'm not quite sure how much we want to nudge to using groups with the taxonomy)...

michaelatmender:
I think this is a good change, however it should also be changed in the following places to match:

- Deployment report modal (both 'overview' and 'details')
- Audit log filters (filter by change -> Deployment > to device group)

tranchitella:
@mzedel @michaelatmender done!

mzedel (approved):

PR #1219, opened 2021-01-17
MEN-4367: fix position of the onboarding tip for "accept pending device"

tranchitella:
Changelog: none


mzedel (approved):
Thanks for this, that should improve the positioning in the bulk of screen sizes I think 👍.  Unfortunately that still depends on the screen you are seeing this on - ideally we would anchor this on the corresponding button, but I guess that is not that urgent...

PR #1220, opened 2021-01-17
MEN-4369: device identity and inventory fields' values are copyable

tranchitella:
Changelog: none

### DIFF
--- a/src/js/components/artifacts/__snapshots__/artifactmetadatalist.test.js.snap
+++ b/src/js/components/artifacts/__snapshots__/artifactmetadatalist.test.js.snap
@@ -17,6 +17,8 @@ Array [
   >
     <div
       onClick={[Function]}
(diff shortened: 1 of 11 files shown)


michaelatmender (changes requested):
I like the feature, but I would suggest to adjust the design a little. I think having the icon visible on all fields all the time is a bit noisy. But we could [show it on hover like this](https://ibb.co/cDcNvGw):

- Right align the icon
- Make it visible when hovering over the whole field, combined with a [tooltip](https://material-ui.com/components/tooltips/) 
- Clicking the whole field will trigger the copy action (use cursor:pointer style to indicate it is clickable)

tranchitella:
@michaelatmender @mzedel thank you for the feedback, I updated the PR to match the design, looks way better!

michaelatmender (approved):
Looks and works great, thanks!

PR #1221, opened 2021-01-17
MEN-4360, MEN-4361: update onboarding snippet for mender-connect

tranchitella:
- Set ServerURL to https://docker.mender.io for on-prem onboarding
- Set permissions of /etc/mender/mender-connect.conf to 0600

Changelog: none

### DIFF
--- a/src/js/components/help/application-updates/__snapshots__/mender-deb-package.test.js.snap
+++ b/src/js/components/help/application-updates/__snapshots__/mender-deb-package.test.js.snap
@@ -151,12 +151,13 @@ mender setup \\
 systemctl restart mender-client && \\
 (cat &gt; /etc/mender/mender-connect.conf &lt;&lt; EOF
 {
-  "ServerURL": "http://localhost",
+  "ServerCertificate": "/usr/share/doc/mender-client/examples/demo.crt",
   "User": "pi",
   "ShellCommand": "/bin/bash"
 }
 EOF
-) && systemctl restart mender-connect'
+) && chmod 0600 /etc/mender/mender-connect.conf && \\
+systemctl restart mender-connect'
 
     </span>
   </div>
--- a/src/js/helpers.js
+++ b/src/js/helpers.js
@@ -499,12 +499,13 @@ ${
     ? `systemctl restart mender-client && \\
 (cat > /etc/mender/mender-connect.conf << EOF
 {
-  "ServerURL": "${document.location.origin}",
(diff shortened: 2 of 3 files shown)


lluiscampos (approved):
This looks good to me, but I'll appreciate @mzedel eyes on it.

mzedel (changes requested):
unfortunately this will break the onboarding on custom onprem setups - we could use the `HOSTNAME` var we pass in the demo environment, but that comes with other issues (see: https://tracker.mender.io/browse/MEN-2803)... I guess the proper fix would be to start the demo environment and point the user to `https://docker.mender.io` or check if the `origin` is localhost and change that to `docker.mender.io`... or change `mender-connect` to use the same backend as `mender-client`, as this needs to know which backend to connect to... (also @lluiscampos)

lluiscampos:
> [...] or change `mender-connect` to use the same backend as `mender-client`, as this needs to know which backend to connect to... (also @lluiscampos)

@mzedel And how is this done today? Doesn't the gui sets up the client with `mender setup --demo ...`, which in effect sets `mender.conf`'s `ServerUrl` to `docker.mender.io`?

mzedel:
it uses the ip address if there is one (`--demo ${ipAddress ? `--server-ip ${ipAddress}` : ''}`) and leaves it to the client otherwise... the `ipAddress` however has the issue described in the ticket...

lluiscampos:
> it uses the ip address if there is one (`--demo ${ipAddress ? `--server-ip ${ipAddress}` : ''}`) and leaves it to the client otherwise... the `ipAddress` however has the issue described in the ticket...

Got it now, thanks. I'll summarize our chat in Slack and get feedback on how to proceed.

lluiscampos (changes requested):
> Got it now, thanks. I'll summarize our chat in Slack and get feedback on how to proceed.

Discussed in Slack: we will de-scope this responsibility from the on-boarding bash snippets. So this PR needs to drop the `docker.mender.io` change but keep the `chmod` one.

tranchitella:
Let's remove ServerURL all together, now this PR depends on:
https://github.com/mendersoftware/mender/pull/701
https://github.com/mendersoftware/mender-connect/pull/13

mzedel (approved):
👍

mzedel:
@lluiscampos I'll merge this without your re-approval to allow picking it to 2.6 - please ping me if there are any additional changes needed...

PR #1235, opened 2021-01-18
fixed the positioning of the 2fa info tooltip as it was blocking the 2fa code entry

mzedel:
Changelog: None


tranchitella (approved):
Tested, works

PR #1237, opened 2021-01-18
fixed an issue that prevented existing user roles from being removed

mzedel:
Changelog: Title


tranchitella (approved):
Tested ok!

PR #1238, opened 2021-01-18
Fix/device group retrieval

mzedel:
this fixes the issue that dynamic groups would not get their devices shown or populated when returning from a different part of the UI to the device groups


tranchitella (approved):
Thanks for the quick fix!|

PR #1248, opened 2021-01-21
Remote terminal: consider status unknown if no status can be retrieved

tranchitella:
Changelog: none


tranchitella:
@mzedel sorry Manuel, I need this in prod very soon, I'll merge without waiting for review directly in staging.

PR #1249, opened 2021-01-21
Remote terminal: consider status unknown if no status can be retrieved

tranchitella:
Changelog: none

### DIFF
--- a/src/js/components/devices/expanded-device.js
+++ b/src/js/components/devices/expanded-device.js
@@ -94,7 +94,16 @@ export class ExpandedDevice extends React.Component {
       showHelptips,
       unauthorized
     } = self.props;
-    const { auth_sets = [], attributes, created_ts, id, identity_data, status = DEVICE_STATES.accepted, connect_status, connect_updated_ts } = device;
+    const {
+      auth_sets = [],
+      attributes,
+      created_ts,
+      id,
+      identity_data,
+      status = DEVICE_STATES.accepted,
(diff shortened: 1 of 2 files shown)


tranchitella:
Counterpart for master of https://github.com/mendersoftware/gui/pull/1248

src/js/components/devices/expanded-device.js
```diff
@@ -174,7 +174,7 @@ export class ExpandedDevice extends React.Component {
           <div className="device-connect bordered report-list">
             <h4 className="margin-bottom-small">Remote Terminal</h4>
             <div className="flexbox" style={{ flexDirection: 'row' }}>
-              {connect_status === DEVICE_CONNECT_STATES.unknown && (
+              {(connect_status === undefined || connect_status === DEVICE_CONNECT_STATES.unknown) && (
```
mzedel, line 177: you could set it like we do for the [device auth status](https://github.com/mendersoftware/gui/blob/master/src/js/components/devices/expanded-device.js#L97) but for the single check here the undefined condition will do just as good... though instead of masking it, a migration of existing devices in the backend would be my first choice 😉

mzedel (approved):

src/js/components/devices/expanded-device.js
tranchitella: @mzedel does it look good, now? https://github.com/mendersoftware/gui/pull/1249/commits/efcb99cda2cc61309bef466211c1366efa03e737

PR #1250, opened 2021-01-21
e2e tests: adjusted unknown path check to work with nicer redirects

mzedel:
This is required since the traefik routing will redirect to the UI, even if an unknown subpath is queried, whereas nginx would return a 404.

Changelog: None


kjaskiewiczz (approved):
looks good

PR #1253, opened 2021-01-22
MEN-4386 single device config editor

mzedel:
additional tests + the API interaction missing + likely needs to be hidden behind a feature flag too, reviews would be helpful already


tranchitella (commented):
Amazing, tested locally and I love it.
The implementation is clean and clear.

tranchitella (approved):
It works fine, but I have one question. At the API level, we have two different end-points to save the configuration (the PUT you are consuming now) and to deploy it (see: https://github.com/mendersoftware/deviceconfig/blob/master/docs/management_api.yml#L119). I don't know if it makes sense to split the "Save" and the "Deploy" UI actions in two buttons, but it definitely makes sense at the API level.

The configuration deployment button is tracked in this task: https://tracker.mender.io/browse/MEN-4430

Maybe it makes sense to add it straight away in this PR?

mzedel:
Thanks @tranchitella for the review, I added the basic deploy call - I would prefer to leave the additional adjustments (log access + listing in the deployments) for the MEN-4430 task though... - OTOH I need to fix the (albeit unrelated) tests to allow merging anyway...

tranchitella (approved):

PR #1270, opened 2021-01-25
fix for coveralls failing to report PR coverage

mzedel:
my best guess is that the lack of git information disturbed the PR association + the dotfile for good measure (based on the comments in the referenced coveralls issue), however I'm not super confident that this fixes the issue...

Changelog: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -178,17 +178,19 @@ test:staging-deployment:
 publish:tests:
   stage: publish
   except:
-    - /^saas-[a-zA-Z0-9.]+$/
+    - /^(staging|saas-[a-zA-Z0-9.]+)$/
   dependencies:
     - test:unit
-  script:
+  before_script:
     - export CI_BUILD_REF=${CI_COMMIT_SHA}
     - export CI_BUILD_REF_NAME=${CI_COMMIT_REF_NAME}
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -178,17 +178,19 @@ test:staging-deployment:
 publish:tests:
   stage: publish
   except:
-    - /^saas-[a-zA-Z0-9.]+$/
+    - /^(staging|saas-[a-zA-Z0-9.]+)$/
```
lluiscampos, line 181: Why do you want to except `staging`. It is reported in other repos so that a PR to staging can still measure coverage.

lluiscampos (approved):
Unsure if this will fix it, but worth trying :)

.gitlab-ci.yml
mzedel: the tests should never change between master and staging, so the coverage will be the same for _master@commit_ - I plan to exempt the unit tests job from staging in a different PR...

lluiscampos: > the tests should never change between master and staging

That is quite an statement. For consistency, we should first add `except:staging` branch from this job in the template, then follow-up down to the repos that have their own coverage publishing (like gui). And of course remove the possible PR checks on staging on all the repos.

In any case, quite unrelated to this PR so go ahead with the merge.

PR #1274, opened 2021-01-28
MEN-4338: set error message on abnormal closure of websocket: Forbidden

tranchitella:


mzedel (approved):
👌 thanks for the cleanup!

tranchitella:
@mzedel is the failure in the acceptance tests some-how expected?

mzedel:
no, I've seen it before but then a restart solved it... I'll have to look into the failures at some point, but if you can feel free to force merge

PR #1295, opened 2021-02-04
Adjustments to offer announcement banner for hosted Mender

michaelatmender:
Updated with Remote terminal link, and it should now be possible to hide/show for different plans + trial tier if needed.

### DIFF
--- a/src/js/components/header/header.js
+++ b/src/js/components/header/header.js
@@ -142,7 +142,7 @@ export class Header extends React.Component {
         className={showOffer ? 'header banner' : 'header'}
         style={{ backgroundColor: '#fff', minHeight: 'unset', paddingLeft: 32, paddingRight: 40 }}
       >
-        {showOffer && <OfferHeader organization={organization} onHide={() => self.setHideOffer()} />}
+        {showOffer && <OfferHeader docsVersion={docsVersion} organization={organization} onHide={() => self.setHideOffer()} />}
         <div className="flexbox">
           <Link to="/">
             <img id="logo" src={isEnterprise ? enterpriseLogo : logo} />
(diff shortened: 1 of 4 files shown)


src/js/components/header/header.js
```diff
@@ -83,11 +94,23 @@ export class Header extends React.Component {
...
+  }
+
   onLogoutClick() {
     this.setState({ gettingUser: false, loggingOut: true, anchorEl: null });
     this.props.logoutUser();
   }
 
+  setHideOffer() {
+    const self = this;
+    this.cookies.set('offer', 'remote', { path: '/', maxAge: 2629746 });
```
mzedel, line 110:
```suggestion
    this.cookies.set('offer', currentOffer.name, { path: '/', maxAge: 2629746 });
```

mzedel (approved):
looks good! one small thing I would change, that might get overlooked otherwise...

src/js/components/header/header.js
michaelatmender: Thanks for catching that!

src/js/components/header/offerheader.js
```diff
@@ -1,14 +1,22 @@
...
+    {organization && organization.trial ? (
+      <span>
+        Try out the new <b>Remote terminal</b> feature for free with your current trial.&nbsp;
+      </span>
+    ) : (
+      <span>
+        Try out the new <b>Remote terminal</b> feature for free on your current plan until March 31st.&nbsp;
+      </span>
+    )}
+    <a href="https://docs.mender.io/add-ons/remote-terminal" target="_blank" rel="noreferrer">
```
mzedel, line 16: based on the tests I think this lacks:
```suggestion
    <a href={`https://docs.mender.io/${docsVersion}add-ons/remote-terminal`} target="_blank" rel="noreferrer">
```

tranchitella (approved):

PR #1296, opened 2021-02-04
MEN-4335

mzedel:
the relevant change is in https://github.com/mendersoftware/gui/pull/1296/commits/57e2a98c5f8545e0df8b069aae6bcb200291ebb2

### DIFF
--- a/src/js/helpers.js
+++ b/src/js/helpers.js
@@ -460,6 +460,7 @@ export const standardizePhases = phases =>
 export const getDebConfigurationCode = (ipAddress, isHosted, isEnterprise, token, deviceType = 'generic-armv6') => {
   let connectionInstructions = ``;
   let demoSettings = `  --quiet --demo ${ipAddress ? `--server-ip ${ipAddress}` : ''}`;
+  const setupConfirmation = `echo "Running mender setup for ${window.location.hostname}" && \\`;
   if (isEnterprise || isHosted) {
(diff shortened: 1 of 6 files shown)


src/js/helpers.js
```diff
@@ -494,7 +498,7 @@ systemctl restart mender-client && \\
   "ShellCommand": "/bin/bash"
 }
 EOF
-) && systemctl restart mender-connect'
+) && systemctl restart mender-connect${setupConfirmation}'
```
lluiscampos, line 501: This is a bit cheating: the "Running mender setup for hosted.mender.io" message should be printed *before* calling `mender-setup`, line 491.

Also, the "hosted.mender.io" part was meant to be an example, if do onboarding with a demo om-prem installation I expect this to print something else. Maybe is better to just say "Running mender setup"

lluiscampos (changes requested):

src/js/helpers.js
mzedel: ah right, I guess I misunderstood the reasoning behind the call... should now be more in line with the intention (I hope)

lluiscampos (approved):
Now it looks good! Thanks and sorry for the bad phrasing in the JIRA task description

mzedel:
Ah I should have just looked better and switched my brain on a bit 🙈

PR #1297, opened 2021-02-04
Fix/attribute scaling issues+release sorting

mzedel:
That should provide some improvements for the dashboard performance and ensure that the release sorting works as expected (uppercase releases no longer land before all lowercase ones)
while I was at it I also added sorting by release modification date - that might help a bit when working with many releases...

### DIFF
--- a/src/js/components/artifacts/__snapshots__/releaseslist.test.js.snap
+++ b/src/js/components/artifacts/__snapshots__/releaseslist.test.js.snap
@@ -2,103 +2,98 @@
 
 exports[`ReleasesList Component renders correctly 1`] = `
 <div
-  className="repository-list flexbox column"
-  style={
-    Object {
-      "alignItems": "flex-start",
-    }
-  }
+  class="repository-list flexbox column"
+  style="align-items: flex-start;"
 >
   <div
-    className="flexbox"
-    style={
-      Object {
-        "alignItems": "center",
-      }
-    }
+    class="flexbox"
+    style="align-items: center;"
   >
     <h3>
       Releases
     </h3>
     <div
-      className="MuiFormControl-root MuiTextField-root search"
-      style={
-        Object {
-          "marginLeft": 30,
-          "marginTop": 0,
-        }
-      }
+      class="MuiFormControl-root MuiTextField-root search"
+      style="margin-left: 30px; margin-top: 0px;"
     >
       <div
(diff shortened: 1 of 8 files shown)


src/js/components/artifacts/releaseslist.js
```diff
@@ -13,17 +13,36 @@ const filters = ['device_types_compatible', 'descriptions', 'Name'];
 
 const filter = createFilterOptions({ stringify: option => filters.map(item => option[item]).join(' ') });
 
+const sortingOptions = [
+  { title: 'Name', value: 'Name' },
+  { title: 'Modification Date', value: 'latestModified' }
```
michaelatmender, line 18: I'd suggest "Date modified" for the label, it's a bit more common phrasing. Nice addition though!

src/js/components/artifacts/__snapshots__/releaseslist.test.js.snap
```diff
@@ -2,103 +2,98 @@
 
 exports[`ReleasesList Component renders correctly 1`] = `
 <div
-  className="repository-list flexbox column"
-  style={
-    Object {
-      "alignItems": "flex-start",
-    }
-  }
+  class="repository-list flexbox column"
```
michaelatmender, line 5: Something in these changes has affected the layout a bit – if you compare it to https://hosted.mender.io/ui/#/releases, you can see this left column is now being cut off a bit

mzedel: hmm that should not have changed with the changes in here (the snapshot-change is just because I had to render it differently to also see the selection menu...) - it might have been from the earlier PR that fixed the scrolling behaviour for the releases and devices list...

src/js/components/artifacts/releaseslist.js
mzedel: Should now be "Date modified" 👌

tranchitella (approved):
Looks great

src/js/components/artifacts/__snapshots__/releaseslist.test.js.snap
michaelatmender: You're right, sorry! I see this and couple of other small things on master too. Will follow up on those separately.

michaelatmender (approved):

PR #1301, opened 2021-02-08
made device listing calls use attribute selection

mzedel:
this is to limit size of response body and UI degradation on large attribute counts

counterpart to mendersoftware/inventory/pull/247

with both PRs the impact of a device with >400 attributes is no longer noticeable - unless the device is expanded (in that case a lot of the time is spent parsing the response where I don't think we can do a lot...)

Changelog: None


tranchitella (approved):
Perfect, LGTM

PR #1302, opened 2021-02-10
MEN-4291 allow add-on + plan changes

mzedel:
@michaelatmender this might already allow adding in MEN-4171 in the [contact form](https://github.com/mendersoftware/gui/blob/4691fff85cd99cb5c36b0015d23386c49fd753ce/src/js/components/settings/quoterequestform.js) - I'll happily add this if I understood the task correctly...


mzedel:
merging after approval on slack 👌

PR #1303, opened 2021-02-10
Fix e2e + unit tests - also update dependencies

mzedel:


webpack.config.js
```diff
@@ -67,6 +74,14 @@ module.exports = {
       }
     ]
   },
+  optimization: {
+    minimize: process.env.NODE_ENV === 'production',
```
merlin-northern, line 78: so it does not influence development builds?

tests/e2e_tests/cypress/integration/04-deviceDetails.spec.js
```diff
@@ -29,11 +29,9 @@ context('Layout assertions', () => {
         () =>
           cy
             .get('.deviceListItem')
-            .click()
             // the deviceconnect connection might not be established right away,
             // due to polling intervals the fastest way to refresh is collapsing & expanding again...
```
merlin-northern, line 33: does this comment still make sense?

tests/setupTests.js
```diff
@@ -21,6 +21,7 @@ let server;
 const oldWindowLocation = window.location;
 const oldWindowSessionStorage = window.sessionStorage;
 
+jest.retryTimes(3);
```
merlin-northern, line 24: nitpick: a const for the 3?

merlin-northern (approved):
lgtm! take or leave the comments.

tests/setupTests.js
mzedel: That whole file is basically just a collection of consts 😅 but it's probably as good as opportunity as any to clean that up a bit

tests/e2e_tests/cypress/integration/04-deviceDetails.spec.js
mzedel: Yep, the clicking will still be done repeatedly, but now just once and then wait until the animation had settled, Check if the button is usable and repeat

webpack.config.js
mzedel: Yep, the dev builds are faster and the prod builds result in a smaller payload this way...

mzedel: Previously this happened automatically, but due to the improved loader we're now using, that setting is required

merlin-northern (approved):
thank you for the answers!

PR #1308, opened 2021-02-12
MEN-4430 show device config deployments

mzedel:
this does not yet include jumping to the device config deployments from the device due to the information available from the device config

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -25,6 +25,7 @@ cat >/var/www/mender-gui/dist/env.js <<EOF
     hostedAnnouncement: "$ANNOUNCEMENT",
     isDemoMode: "$DEMO",
     features: {
+      hasDeviceConfig: "$HAVE_DEVICECONFIG",
       hasDeviceConnect: "$HAVE_DEVICECONNECT",
       hasMultitenancy: "$HAVE_MULTITENANT",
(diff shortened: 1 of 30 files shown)


tranchitella (commented):
I tested it, but I have a few (minor) issues.

First of all, I don't know how to enable it in the OS version. I ended up patching the UI to use `haveDeviceConfig: true` hardcoded. Can you point me to the right way to enable the feature in the OS version? This way, we can set the right env variable in integration and have it enabled out of the box.

Second issue i have is that during deployment, the status reported in the expanded device doesn't update, it stays at "pending". The status are correctly updated in the backend, and I can see them in the deployment details in the "deployments" menu.

Other than that, the feature is okay and I could follow the code, though I didn't go through all of it because it is a bit outside my comfort zone.

mzedel:
Thanks for looking into it! I'll get the refresh in right asap... as to where to activate it... that is solely based on the information from tenantadm (so not available in OS) so far - I added a `addons: [{ Name: 'configure', Enabled: true }]` in the tenant object... but if it should also be available in OS through an env var I'll gladly add this too...

tranchitella:
@mzedel yes, device configuration is an open source feature!

tranchitella (approved):

PR #1321, opened 2021-02-19
MEN-4313 Terminal session playback + download

mzedel:
the final backend change alignment is still pending, but the remaining changes should be tiny


merlin-northern (commented):
Manuel, I am putting the comments also here, apart from the private chat.

* it does replay with delays now, but not always (shared privately the particular recordings and control data) the previous version at 00afff168a28f74402662e6903bb0194aa549193 played..
* Changed in audit logs: always device id
* Session ID in session details in the player: set to device id

merlin-northern:
@mzedel I found it extremely challenging to debug the player, if you have any hints on how to add console logs in _terminalplayer.js_ and in `processContent` in particular that would be lovely. (I even tried alert to no luck whatsoever)

src/js/components/auditlogs/eventdetails/terminalsession.js
```diff
@@ -0,0 +1,107 @@
...
+      action.startsWith('close') ? time : undefined
+    ).then(setSessionDetails);
+  }, []);
+
+  if (!(sessionDetails && device)) {
+    return <Loader show={true} />;
+  }
+
+  const { name, device_type, artifact_name } = device.attributes;
+  const usesId = !idAttribute || idAttribute === 'id' || idAttribute === 'Device ID';
```
merlin-northern, line 43: `'Device ID'` is used a bit all over the code, perhaps time to define it? (not for this PR, but maybe one day)

src/js/components/auditlogs/eventdetails/terminalsession.js
```diff
@@ -0,0 +1,107 @@
...
+import { getDeviceById, getSessionDetails } from '../../../actions/deviceActions';
+import { getIdAttribute } from '../../../selectors';
+import theme from '../../../themes/mender-theme';
+import Loader from '../../common/loader';
+import { Terminal } from '../../devices/terminal';
+import { formatTime } from '../../../helpers';
+
+momentDurationFormatSetup(moment);
+
+const BEGINNING_OF_TIME = '2020-01-01T00:00:00.000Z';
```
merlin-northern, line 19: if someone is mean, then she will exploit it by changing the time on the device. I would set EPOCH but your call. strange that we do not allow to list all the entries. my second question is: it returns really all of them without pages?

src/js/components/devices/terminal.js
```diff
@@ -120,11 +125,11 @@ export const Terminal = ({ onCancel, sendMessage, setSnackbar, setSessionId, set
...
           body
         } = MessagePack.decode(data);
-        if (proto !== MessageProtocolShell) {
+        if (proto !== MessageProtocol.Shell) {
           return;
         }
         switch (typ) {
-          case MessageTypeNew: {
+          case MessageTypes.New: {
             if (props.status == 2) {
```
merlin-northern, line 134: the only const missing now here :) this `2` being an `ErrorMessageStatus`.

src/js/components/devices/terminal.js
```diff
@@ -9,27 +9,32 @@ import { SearchAddon } from 'xterm-addon-search';
...
-const MessageTypeShell = 'shell';
-const MessageTypeStop = 'stop';
+export const MessageTypes = {
+  New: 'new',
+  Ping: 'ping',
+  Pong: 'pong',
+  Resize: 'resize',
+  Shell: 'shell',
+  Stop: 'stop'
+};
```
merlin-northern: lovely! I quess the 'delay' comes in the following commits.

src/js/components/auditlogs/eventdetails/terminalsession.js
mzedel: yep... definitely planned... it gets all the more important with the vision of relying also on custom names in the future...

src/js/components/auditlogs/eventdetails/terminalplayer.js
```diff
@@ -18,6 +18,128 @@ const searchAddon = new SearchAddon();
...
+        const processContent = () => {
+          if (contentIndex === content.length) {
+            return resetPlayer();
+          }
+          const item = content[contentIndex];
+          contentIndex += 1;
+          let delay = 1;
+          if (item.content) {
+            const buffer = JSON.parse(atob(item.content));
+            term.write(byteArrayToString(buffer.data || []))
```
merlin-northern, line 111: during debugging I added console log here, but it did not work, it was next to impossible to debug, could you teach me how to debug here in case I need it? (and assume I only have the webdev tools from the browser and a text editor (vim).

src/js/components/auditlogs/eventdetails/terminalsession.js
mzedel: nah that is mostly just since the beginning of us possibly having the recording feature in any way - if someone wants to spoof anything anywhere they are on their own...
to the second question: it will just get you to the paginated page - with the proper filter set... however - atm we can't filter by `object_type=device` unfortunately... but by `object_id=<id>` works fine 😄

merlin-northern (approved):
lgtm. tested and works wonders! thanks!

src/js/components/auditlogs/eventdetails/terminalplayer.js
mzedel: there you will only see the effects if you download the recording and open the html file - that is why this is in the `generateHtml` function... if you want to debug anything in the UI itself look a few lines below (around L195) and edit the `useEffect` function... 👍

merlin-northern:
> lgtm. tested and works wonders! thanks!

@mzedel and also: you fixed the "docker virtual machine" in the on boarding icon disapearing.

src/js/components/devices/terminal.js
mzedel: in the _other_ terminal PR there is some additional work to bring those into the device constants... I'll rebase there and will adjust accordingly - including the error and stop codes 👌

PR #1322, opened 2021-02-19
Use device drawer

mzedel:


michaelatmender (approved):
Looks good, just had one or two small adjustments to buttons for opening/closing the device drawer. Added a commit - let me know if it looks ok to you!

PR #1342, opened 2021-03-01
MEN-4320 file up-/ download

mzedel:
this should also allow cmd+click on a filename-like text on a new terminal line (if downloading the clicked content fails, the text is used as download path) + dropping a file on the terminal should select that file for upload & use the filename as a target location

depends partly on #1321 (mainly for the links to the auditlogs to work)


src/js/actions/deviceActions.js
```diff
@@ -590,6 +592,28 @@ export const getSessionDetails = (sessionId, deviceId, userId, startDate, endDat
   );
 };
 
+export const deviceFileDownload = (deviceId, path) => () => GeneralApi.post(`${deviceConnect}/devices/${deviceId}/download`, { path });
+
+export const deviceFileUpload = (deviceId, path, file) => dispatch => {
+  var formData = new FormData();
+  formData.append('file', file);
+  formData.append('path', path);
```
tranchitella, line 600: File must be the last field in the multipart form-data:

```suggestion
  formData.append('path', path);
  formData.append('file', file);
```

src/js/actions/deviceActions.js
```diff
@@ -590,6 +592,28 @@ export const getSessionDetails = (sessionId, deviceId, userId, startDate, endDat
...
+  const cancelSource = axios.CancelToken.source();
+  return Promise.all([
+    dispatch(setSnackbar('Uploading file')),
+    dispatch({ type: AppConstants.UPLOAD_PROGRESS, inprogress: true, uploadProgress: 0, cancelSource }),
+    GeneralApi.upload(`${deviceConnect}/devices/${deviceId}/upload`, formData, e => progress(e, dispatch), cancelSource.token)
```
tranchitella, line 605: File upload is (now) implemented as PUT, not POST.
```suggestion
    GeneralApi.uploadPut(`${deviceConnect}/devices/${deviceId}/upload`, formData, e => progress(e, dispatch), cancelSource.token)
```

tranchitella: This obviously requires something like:
```
diff --git a/src/js/api/general-api.js b/src/js/api/general-api.js
index 873d38ac3..b2fdfb39c 100644
--- a/src/js/api/general-api.js
+++ b/src/js/api/general-api.js
@@ -33,7 +33,8 @@ const Api = {
   post: (url, data) => authenticatedRequest.post(url, data),
   postUnauthorized: (url, data) => axios.post(url, data, commonRequestConfig),
   put: (url, data) => authenticatedRequest.put(url, data),
-  upload: (url, formData, progress, cancelToken) => authenticatedRequest.post(url, formData, { onUploadProgress: progress, timeout: 0, cancelToken })
+  upload: (url, formData, progress, cancelToken) => authenticatedRequest.post(url, formData, { onUploadProgress: progress, timeout: 0, cancelToken }),
+  uploadPut: (url, formData, progress, cancelToken) => authenticatedRequest.put(url, formData, { onUploadProgress: progress, timeout: 0, cancelToken })
 };

 export default Api;
 ```

File: src/js/actions/deviceActions.js [hunk 3]
```diff
@@ -590,6 +592,28 @@ export const getSessionDetails = (sessionId, deviceId, userId, startDate, endDat
   );
 };
 
+export const deviceFileDownload = (deviceId, path) => () => GeneralApi.post(`${deviceConnect}/devices/${deviceId}/download`, { path });
```
tranchitella, line 595: Download technically works (I can see the response with the dev tools), but nothing happens in the UI and the file is not downloaded. This is tricky, I would try to let the browser handle the download of the content-disposition: attachment file, but to do that you cannot rely on XHR requests. The only (hackish?) solution I can think of is creating a hidden iframe with the post form, target a new window and submit it, it should trigger the download). Another option would be to change the backend and implement the download as a GET end-point, passing the params as GET parameters and relying on the Cookie to pass in the JWT token. What do you think?

src/js/actions/deviceActions.js
mzedel: thanks for these => I took them as they were and it worked flawlessly 😃

src/js/actions/deviceActions.js
mzedel: 🤦 sorry - is now fixed!

src/js/components/devices/troubleshoot/filetransfer.js
```diff
@@ -0,0 +1,137 @@
...
+  };
+
+  const onPasteUploadClick = async () => {
+    const path = await navigator.clipboard.readText();
+    setUploadPath(path);
+  };
+
+  const onFileSelect = file => {
+    setFile(file);
+    setUploadPath(`${uploadPath}/${file?.name || ''}`);
```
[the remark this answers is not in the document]
mzedel: argh, sorry - should now as intended (clears the filename if the filename exists in the path and leaves it untouched otherwise)

mzedel:
Thanks for going through the PR and the detailed report @tranchitella! Unfortunately resolving the error combined with the direct access to the file transfer part took a little longer than expected... but now it should be a bit clearer how things work...
The bigger problem is indeed the download part... Sorry for the extra work this might now need - I was expecting more a redirect to the download location or the same behaviour as for the auditlogs csv download (POST with params, get a download location in return)... I guess the easiest for the UI would be proxying this with a GET request `/deviceconnect/devices/<deviceId>/download?path=<base64Path>` as this should then be possible to be handed off to a regular `<a href={`/deviceconnect/devices/<deviceId>/download?path=<base64Path>`} download="theFilename.ext" />`...

tranchitella:
@mzedel I tested again this PR, it works fine but the download needs to be updated to use the GET end-point, instead of the POST. File upload and user interface are fine, I tried it both with small and large (400MB) files, no issues found.

tranchitella:
One thing: the auditlogs page crashes with this error:
image

mzedel:
Oh my... I spent so much time on getting the e2e tests for this to work that I forgot to push the changes.. I'm terribly sorry!

tranchitella (approved):
Tested and it works perfectly, uploading and downloading large files with the current master versions of the client and mender-connect. Well done!

PR #1353, opened 2021-03-04
onboarding: Pass `--demo` flag to `get.mender.io` script.

kacf:
Changelog: Title

Fixes MEN-4461.

### DIFF
--- a/src/js/helpers.js
+++ b/src/js/helpers.js
@@ -494,7 +494,7 @@ ${enterpriseSettings}`;
   }
   const debInstallationCode = `wget -q -O- https://get.mender.io/${
     isPreRelease && window.location.hostname.includes('staging') ? 'staging' : ''
-  } | sudo bash -s`;
+  } | sudo bash -s -- --demo`;
   let codeToCopy = `${debInstallationCode} && \\
 sudo bash -c 'DEVICE_TYPE="${deviceType}" && \\${
     token
--- a/src/js/helpers.js
+++ b/src/js/helpers.js
@@ -494,7 +494,7 @@ ${enterpriseSettings}`;
   }
   const debInstallationCode = `wget -q -O- https://get.mender.io/${
     isPreRelease && window.location.hostname.includes('staging') ? 'staging' : ''
-  } | sudo bash -s -- --demo`;
+  } | sudo bash -s${isPreRelease && window.location.hostname.includes('staging') ? ' -- --demo' : ''}`;
   let codeToCopy = `${debInstallationCode} && \\
 sudo bash -c 'DEVICE_TYPE="${deviceType}" && \\${
     token
--- a/src/js/components/common/dialogs/__snapshots__/physicaldeviceonboarding.test.js.snap
+++ b/src/js/components/common/dialogs/__snapshots__/physicaldeviceonboarding.test.js.snap
@@ -1993,7 +1993,7 @@ exports[`PhysicalDeviceOnboarding Component tiny onboarding tips renders Install
           <span
             style="white-space: pre-wrap; word-break: break-word;"
           >
-            wget -q -O- https://get.mender.io/ | sudo bash -s && \\
+            wget -q -O- https://get.mender.io/ | sudo bash -s -- --demo && \\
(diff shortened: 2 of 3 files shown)


src/js/helpers.js
```diff
@@ -494,7 +494,7 @@ ${enterpriseSettings}`;
   }
   const debInstallationCode = `wget -q -O- https://get.mender.io/${
     isPreRelease && window.location.hostname.includes('staging') ? 'staging' : ''
-  } | sudo bash -s`;
+  } | sudo bash -s -- --demo`;
```
lluiscampos, line 497: This will only work for `https://get.mender.io/staging`, so merging it as-is will block any HM upgrade that we would like to do before Mender 2.7

I don't know if is the best solution, but in order to not add more hacks to the `gui` repo, my proposal is to modify production script to accept (but silently ignore) the new `--demo` flag.

Other ideas?

lluiscampos (changes requested):

src/js/helpers.js
mzedel: preferably there would be yet another condition in there, as the `--demo` likely should only be applied during onboarding, but we know that user use that code snippet all the time - so we'll have to pass the onboarding state in there too and that function will be even more hackish... with that in mind (and the possible reduction in complexity): could we get the script to accept the flags directly ready for 2.7? 👍

and with regards to the staging limitation: would the additional flag actually disturb the `get.mender.io` script or would it just be ignored?

kacf: > we know that user use that code snippet all the time

I wasn't aware it was used elsewhere. I thought it was only for onboarding. 

> with that in mind (and the possible reduction in complexity): could we get the script to accept the flags directly ready for 2.7?

I'm not sure I understand. Isn't this accepting them directly?

> This will only work for `https://get.mender.io/staging`, so merging it as-is will block any HM upgrade that we would like to do before Mender 2.7

Good thinking! This is easy to fix: Feature branch, and add an extra action point in https://tracker.mender.io/browse/MEN-4499.

lluiscampos: I am a bit confused, can we have chat about this today after SU?

kacf: Discussed privately, will include a conditional in the UI for staging environment instead of using a feature branch. Will not add any other flags right now, since it's not part of this task which has already been going on for a long time.

kacf: @mzedel: How about now?

Note that I left the original commit intact in the history, so that this is an easy revert later. I suppose this can be added as a cleanup item to https://tracker.mender.io/browse/MEN-4499.

src/js/helpers.js
```diff
@@ -494,7 +494,9 @@ ${enterpriseSettings}`;
   }
   const debInstallationCode = `wget -q -O- https://get.mender.io/${
     isPreRelease && window.location.hostname.includes('staging') ? 'staging' : ''
-  } | sudo bash -s`;
+  } | sudo bash -s${
+    isPreRelease && window.location.hostname.includes('staging') ? ' -- --demo' : ''
+  }`;
```
mzedel, line 499:
```suggestion
  } | sudo bash -s${isPreRelease && window.location.hostname.includes('staging') ? ' -- --demo' : ''}`;
```
should make the linter happy (we have a prettier config with a 160 col line width 😉)

mzedel (changes requested):
👌 - that should work and be good to get it tested on staging... the formatting + the test output need to be adjusted but apart from that it should pass (interestingly even the otherwise failing tests passed in the previous run 🤯)

lluiscampos (approved):
Looks good from my side :+1: 

@mzedel have the last word

kacf:
@mzedel: Still timing out due to the unstable e2e tests, but other tests seem to be fine. I leave it in your hands at this point. Either merge immediately if you are comfortable with it, or wait until you have fixed/worked around the other issues.

mzedel (approved):
👍 should be fine now! unfortunately I don't have the rights to (force-)merge - but https://github.com/mendersoftware/gui/pull/1351 should hopefully fix _everything_ 😅

kacf:
Thanks!

PR #1360, opened 2021-03-09
fixed auditlog filter-by-url to work with device type

mzedel:
- also ensures the set filter is showing up readable
depends on MEN-4491
Changelog: None


tranchitella (approved):
Works fine!

PR #1363, opened 2021-03-12
made the inclusion of the livechat depend on the environment the bundle was created for

mzedel:
- this is purely to avoid errors appearing in the debugger during development

Changelog: None


tranchitella (approved):
Finally, it is was so annoying...

PR #1364, opened 2021-03-12
MEN-4484 user verification + 2fa fixes

mzedel:


merlin-northern:
I am getting this when I try to enable 2fa:
`The settings couldn't be saved. Two factor authentication must be enabled prior to generating QR code... [Request ID: f60adaee]`
with a new user I get the same. have not tried a new tenant.

mzedel:
Thanks for taking a look @merlin - what steps did you take to get that error? I just tried it again both on just a fresh user and on a tenant and I couldn't get there... my steps are as follows:
- create user or tenant
- try to activate 2fa -> verification required
- set user as `verified: true`
- get 2fa code
- 🎉

merlin-northern:
> Thanks for taking a look @Merlin - what steps did you take to get that error? I just tried it again both on just a fresh user and on a tenant and I couldn't get there... my steps are as follows:
> 
> * create user or tenant
> * try to activate 2fa -> verification required
> * set user as `verified: true`
> * get 2fa code
> * 🎉

in here: I already have a tenant, I tried with new user and existing user.
new user:
* create a new user
* assign admin role
* login as new user in another browser
* try to enable/disable tfa

with existing user:
* just try to enable/disable tfa

merlin-northern:
there is another problem, apart the ones we faced and that you solved.
* I can disable TFA with the current implementation
* once I disable-enable the TFA in order to force the secrete renewal and leave it in an unverified state, and logout then, I lock myself out.
I am sure we cant get around that without modifications to the backend. I will try to fix that while keeping the API as is.

merlin-northern:
> 
> I am sure we cant get around that without modifications to the backend. I will try to fix that while keeping the API as is.

Could you give this a try: https://github.com/mendersoftware/useradm-enterprise/pull/235
(quick hack, I tested it, but briefely)

tranchitella (approved):
I confirm it's working as expected

PR #1378, opened 2021-03-16
Speedup build time

mzedel:


merlin-northern:
with it:
```bash
webpack 5.26.0 compiled successfully in 45722 ms
```

without it:
```
webpack 5.24.3 compiled with 2 warnings in 48039 ms
```

3s is always something!

mzedel:
sorry to see that 😞 on Gitlab that looks more like:
without it (https://gitlab.com/Northern.tech/Mender/gui/-/jobs/1101460085):
```
webpack 5.24.3 compiled with 2 warnings in 92480 ms
```
and with it (https://gitlab.com/Northern.tech/Mender/gui/-/jobs/1101427580):
```
webpack 5.26.0 compiled with 2 warnings in 46931 ms
```

which aligns more with my local difference of ~50s vs ~25s

merlin-northern:
> sorry to see that 😞 on Gitlab that looks more like:
> without it (https://gitlab.com/Northern.tech/Mender/gui/-/jobs/1101460085):
> 
> ```
> webpack 5.24.3 compiled with 2 warnings in 92480 ms
> ```
> 
> and with it (https://gitlab.com/Northern.tech/Mender/gui/-/jobs/1101427580):
> 
> ```
> webpack 5.26.0 compiled with 2 warnings in 46931 ms
> ```
> 
> which aligns more with my local difference of ~50s vs ~25s

incredible. further more, with the `npm run watch` it seems a bit slower:
```bash
webpack 5.26.0 compiled successfully in 11869 ms
```
where without it I have:
```bash
webpack 5.24.3 compiled successfully in 4594 ms
```

 (with the same change, and started from scratch each time)

mzedel:
the watch times are likely just due to the other dependencies as only the `npm run build` part is changed... but still sad to see the updated dependencies causing slow downs...

tranchitella (approved):
Looks good to me

PR #1385, opened 2021-03-17
improved deployment creation for large device fleets

mzedel:
- removed need to retrieve all device ids on deployment creation for all device

Changelog: Title


tranchitella (approved):
Well done! To be 100% honest, I didn't remember we had an end-point to deploy to all the accepted devices, I used dynamic groups to achieve it. I tested it and it works just fine, indeed.

mzedel:
It was only added somewhat recently and I didn't get around to add it earlier - but the nice stress test demo made the absence unbearable

PR #1389, opened 2021-03-18
prevented global 2fa setting from being saved for unverified users

mzedel:
Changelog: None


merlin-northern (approved):
tested. works for me.

tranchitella (approved):

PR #1400, opened 2021-03-22
MEN-4641 improved dashboard active devices widget + added 'all devices retrieval' cancellation

mzedel:
Changelog: None


tranchitella (approved):
Looks good, tested locally with a small fleet. I don't have a large fleet at hand, but I trust you tested it there as well :)

mzedel:
not quite the scale where I noticed it, but I tried it with my ~150 stress-test-client instances and adjusted the device limit from 5000 to 50 + the `getAllDevices...` page size to 30 - that should work similarly, as my local setup already starts to respond slower in this situation... and it worked out as expected 👌

PR #1402, opened 2021-03-22
MEN-4564: update RBAC roles' names to be more descriptive

tranchitella:
Changelog: none

### DIFF
--- a/src/js/components/settings/__snapshots__/roles.test.js.snap
+++ b/src/js/components/settings/__snapshots__/roles.test.js.snap
@@ -84,7 +84,7 @@ exports[`Roles Component renders correctly 1`] = `
         <td
           class="MuiTableCell-root MuiTableCell-body"
         >
-          Read only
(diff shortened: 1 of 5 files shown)


kjaskiewiczz (approved):
looks more intuitive

mzedel (commented):
while I agree with the _Read Access_ title, I'm torn about the _Release Manager_ as to me this implies UI access, but limited to the releases tab 😕 - whereas I would associate _CI_ with API (release + deployment creation) but not UI access...

tranchitella:
@mzedel isn't it the same with "Deployments Manager"? I tried to make it consistent.

mzedel (approved):
👍 good point! and I just checked the role config in useradm - my understanding of the role was wrong and _Release Manager_ is indeed very fitting!

michaelatmender (approved):
This looks clear to me, very good!

PR #1403, opened 2021-03-24
brought back click-to-copy on device attributes

mzedel:
needs #1409 to be merged

Changelog: None


michaelatmender (changes requested):
This works fine, but it is a bit hidden now. Could you add in the copy icon when the user hovers on a row too? (if it requires the rows to have a little more space between them, that's ok)
hovercopy

mzedel:
thanks @michaelatmender! should now be very close - I kept it triggering on the value part only though (as I would expect the row to include both key & value as they are now almost equally prominent)...

oleorhagen (commented):
:eyes: 
This javascript is some 20 years to modern for my liking

michaelatmender (approved):

PR #1404, opened 2021-03-24
Fix id and position of auth set tooltip

michaelatmender:
Changelog: None


mzedel (approved):
👍 good point - that should keep it out of the way and in a stable position even during expands... sorry about the ids 🙈

PR #1406, opened 2021-03-24
fixed release list filter & upload heights to prevent squashed/ invisible elements

mzedel:
Changelog: None


michaelatmender (approved):
Tested on Safari 14.0.1 which showed the bug before - looks good now

PR #1409, opened 2021-03-24
Revert revert device drawer

mzedel:


mzedel:
@tranchitella this _should_ be good as is - just want to make sure that the string of follow up PRs is not disturbed...

tranchitella (approved):
Yup, that's what I was thinking about, too: reverting the revert.

PR #1410, opened 2021-03-24
Fix/device details expansions

mzedel:
this should make the expansion of the detailed information appear more fluent

depends on #1403 (and #1409) being merged


michaelatmender (approved):
Looks a lot smoother, thanks!

tranchitella (approved):

PR #1411, opened 2021-03-25
Revert "Enable get.mender.io `--demo` flag only on staging for now."

kacf:
Fixes MEN-4499.

This reverts commit a77d02cafa0050ccde8ac88aef57a2b963278ed9.

Changelog: None


kacf:
@tranchitella, @mzedel: After this is merged, it is important that Hosted Mender is not upgraded until the client is released. If this is problematic, we could consider merging [2.7.x only](https://github.com/mendersoftware/gui/pull/1412) for now and wait with this one.

mzedel (approved):

tranchitella (approved):

PR #1415, opened 2021-03-25
prevented enter press on 2fa code entry causing a redirect

mzedel:
- this is due to the default form behavior causing a GET via the window location unless stopped

not strictly 2.7.x relevant

Changelog: Title

### DIFF
--- a/src/js/components/common/forms/form.js
+++ b/src/js/components/common/forms/form.js
@@ -205,6 +205,11 @@ export default class Form extends React.Component {
       this.updateModel();
     }
   }
+
+  onSubmit(event) {
+    event.preventDefault();
+  }
+
   render() {
     var uploadActions = !!this.props.showButtons && (
       <div
@@ -230,7 +235,12 @@ export default class Form extends React.Component {
     );
 
     return (
-      <form key={this.props.uniqueId} className={this.props.className || ''} autoComplete={this.props.autocomplete || undefined}>
+      <form
+        key={this.props.uniqueId}
+        className={this.props.className || ''}
+        autoComplete={this.props.autocomplete || undefined}
+        onSubmit={e => this.onSubmit(e)}
+      >
         {this.newChildren}
         {uploadActions}
       </form>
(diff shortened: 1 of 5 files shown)


merlin-northern:
testing it now

src/js/components/common/forms/form.js
```diff
@@ -201,10 +201,13 @@ export default class Form extends React.Component {
   }
   _handleKeyPress(event) {
     event.stopPropagation();
-    if (event.key === 'Enter' && this.state.isValid) {
-      this.updateModel();
-    }
```
merlin-northern, line 206: side effect on the login screen, when focus is in the password entryfield (you just finished typing password) pressing 'Enter' does not submit the form -- you have to use the mouse.

merlin-northern (commented):
works and looks good, but has a side effect on the login screen, which drives at least me crazy ;-)

mzedel:
thanks @merlin-northern for checking - I tried a few other forms in the UI but failed to check the login form... now it should work for both!

merlin-northern:
> thanks @merlin-northern for checking - I tried a few other forms in the UI but failed to check the login form... now it should work for both!

works like a charm!

there is one problem: with this branch I cannot disable Two Factor Authentication. steps to reproduce:
* get a user with 2FA enabled
* login to that user
* go to profile and disable 2FA
* logout
* try to login

mzedel:
can you please double check this, as the button to enable/disable 2fa is outside a `Form` component (where this change is in) - so it should be unrelated...

merlin-northern:
> can you please double check this, as the button to enable/disable 2fa is outside a `Form` component (where this change is in) - so it should be unrelated...

I understand that, but the fact remains, that at `687a43d294dbbdfeaac6e72f3e6b7f443e8cfdf0` I cant disable 2fa while on hosted I can.

merlin-northern:
> > can you please double check this, as the button to enable/disable 2fa is outside a `Form` component (where this change is in) - so it should be unrelated...
> 
> I understand that, but the fact remains, that at `687a43d294dbbdfeaac6e72f3e6b7f443e8cfdf0` I cant disable 2fa while on hosted I can.

ok lets not block this one, since I cant disable 2fa for my user in general now.

PR #1444, opened 2021-03-30
prevented device type selection tooltip if onboarding state doesn't allow it

mzedel:
Changelog: None


merlin-northern (approved):
lgtm! tested -- works like a charm!

PR #1452, opened 2021-03-30
explicitly retrieved full device details on onboardingtip creation

mzedel:
- this should get the full attribute list incl. device ips irrespective of the way to the dialog
Changelog: None


mzedel:
yeah... it is limited to the first 20 devices and should mostly just retrieve the one from onboarding...

PR #1460, opened 2021-04-01
MEN-4568: fix calendar view of dynamic deployments

tranchitella:
Changelog: none


mzedel (approved):
the comment right above could be adjusted too, but OTOH it is still close enough...

tranchitella:
@mzedel I adjusted the comment

PR #1461, opened 2021-04-01
MEN-4597: fix session logging playback

tranchitella:
Changelog: none


mzedel (approved):

tranchitella:
@mzedel let's keep this on hold until we check why we get audit logs without the session_id

mzedel:
For now/ to potentially get it into another release build I think this would suffice, but longer term a closer look at the backend would be best I think

PR #1463, opened 2021-04-01
explicitly close event details on device listing click

mzedel:
- this is needed to also close the drawer if the resulting location doesn't differ

Changelog: None


tranchitella (approved):
Thanks, I also noticed it but did get to fix it.

PR #1464, opened 2021-04-01
allowed device details to show a single config setting as teaser

mzedel:
- unlike for the rest of the device details sections we can't make an informed selection
- thus bluntly pick a single one and hide the rest
Changelog: None


tranchitella (approved):
It works fine and fixes all the quirks I had (better used with https://github.com/mendersoftware/deviceconfig/pull/28).
I have one additional problem, though, you can help me with.

Can you please tell me if setDeviceConfig and applyDeviceConfig are executed serially or in parallel [here](https://github.com/mendersoftware/gui/blob/master/src/js/components/devices/device-details/configuration.js#L244)? The backend expects it to be executed serially, because the deployment of a configuration snapshots the configuration at that moment.

I'm asking because in my tests it happened a few times the configuration deployed to the device was the previous value, and not the one I just changed in the UI.

mzedel:
apparently I misinterpreted the `{ retries: 0 }` as an example config... - now it should be run serially...

PR #1467, opened 2021-04-02
Date time copy-paste.

merlin-northern:
ChangeLog:none

### DIFF
--- a/src/js/components/common/configurationobject.js
+++ b/src/js/components/common/configurationobject.js
@@ -8,9 +8,13 @@ import { FileCopyOutlined as CopyToClipboardIcon } from '@material-ui/icons';
 
 const ValueColumn = ({ value, setSnackbar }) => {
   const [tooltipVisible, setTooltipVisible] = useState(false);
-  const onClick = ({ target: { dateTime } }) => {
+  const onClick = () => {
     if (setSnackbar) {
-      copy(dateTime || value);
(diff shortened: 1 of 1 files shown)


mzedel (approved):
that should work as it is - I would suggest:
```js
const onClick = () => {
  if (setSnackbar) {
    let copyable = value;
    if (React.isValidElement(value)) {
      copyable = value.props.value;
    }
    copy(copyable);
  }
};
```
to clarify a bit more what value is in that situation... in general I think that component provides a bit too much flexibility in rendering things, while also trying to return copied content... (say you provide a component that renders text like things that you would expect to copy, but the provided component is a `canvas` or several components...) 😕

merlin-northern:
> that should work as it is - I would suggest:
> 
> ```js
> const onClick = () => {
>   if (setSnackbar) {
>     let copyable = value;
>     if (React.isValidElement(value)) {
>       copyable = value.props.value;
>     }
>     copy(copyable);
>     setSnackbar('Value copied to clipboard');
>   }
> };
> ```
> 
> to clarify a bit more what value is in that situation... in general I think that component provides a bit too much flexibility in rendering things, while also trying to return copied content... (say you provide a component that renders text like things that you would expect to copy, but the provided component is a `canvas` or several components...) 😕

understood, fixed. thank you.

PR #1477, opened 2021-04-06
made device status changes trigger list updates more aggressively

mzedel:
Changelog: None


tranchitella (approved):
I tested it and works fine; to be honest, I'm not 100% sure about what this PR is fixing, but at least I'm not experiencing any regression.

merlin-northern (approved):
tested, although I did see that:
`There was a problem decommissioning the device: Cannot read property '0' of undefined` when I decommissioned a device that was not connected at all, and right after decommissioning started clicking like crazy. it maybe something local.

merlin-northern:
> tested, although I did see that:
> `There was a problem decommissioning the device: Cannot read property '0' of undefined` when I decommissioned a device that was not connected at all, and right after decommissioning started clicking like crazy. it maybe something local.

and the device details almost immediately disappears on decommission. so nice!

PR #1486, opened 2021-04-06
Fix/config tooltips

mzedel:


merlin-northern (approved):
works (tested) there is a way to fool the tooltips but it involves the duplicates and we decided not the make it too complicated.

src/js/components/devices/device-details/configuration.js
```diff
@@ -181,7 +181,8 @@ export const DeviceConfiguration = ({
 
   useEffect(() => {
     const { config = {} } = device;
-    if (!changedConfig && device.config && (!config.deployment_id || deployment.status)) {
+    const isDeploying = deployment_id && deployment_id !== '00000000-0000-0000-0000-000000000000';
```
tranchitella, line 184: This shouldn't happen anymore with the latest deviceconfig (see https://github.com/mendersoftware/deviceconfig/pull/33)
Are you sure you are using the right version of the backend?

tranchitella (changes requested):

kacf (approved):
I can only comment on the tooltip id collision, but that part looks good!

src/js/components/devices/device-details/configuration.js
mzedel: it used to happen only immediately after the config was requested to be deployed, with the version I was running it was returned all the time... I'll try once more with the most recent one, but I'd prefer to keep it in to prevent the post-deploy-request deployment id from wrongly signaling the beginning of the deployment.

tranchitella (approved):

src/js/components/devices/device-details/configuration.js
tranchitella: Thanks for removing it; it was a bug in the backend, I'm 100% sure you won't get the "zero-value" UUID.

PR #1490, opened 2021-04-09
MEN-4610: Fix search parameters not included when exporting auditlogs

alfrunes:


mzedel (approved):
🙈 thank you very much!

tranchitella (approved):

PR #1508, opened 2021-04-14
MEN-4503: track the campaign parameter in the sign up page

tranchitella:
Changelog: none

### DIFF
--- a/src/js/components/user-management/signup.js
+++ b/src/js/components/user-management/signup.js
@@ -16,7 +16,7 @@ import { OAuth2Providers } from './oauth2providers';
 
 const cookies = new Cookies();
 
-export const Signup = ({ createOrganizationTrial, currentUserId, loginUser, setFirstLoginAfterSignup, recaptchaSiteKey, setSnackbar }) => {
+export const Signup = ({ createOrganizationTrial, currentUserId, loginUser, setFirstLoginAfterSignup, recaptchaSiteKey, setSnackbar, match }) => {
   const [step, setStep] = useState(1);
   const [email, setEmail] = useState('');
   const [loading, setLoading] = useState(false);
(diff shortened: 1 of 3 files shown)


tranchitella:
Depends on: https://github.com/mendersoftware/tenantadm/pull/276

tranchitella:
@mzedel I have no idea why unit tests are failing in components I haven't touched. Ideas?

mzedel:
yep, they weren't touched for a while, but fail suddenly consistently on Gitlab (while passing locally)... https://github.com/mendersoftware/gui/pull/1507 tries to fix this, but the underlying reason is still a mystery to me...

tranchitella:
@mzedel shall we ignore that specific failure? the PR is ready-for review, my change here is minimal.

src/js/components/user-management/signup.js
```diff
@@ -27,6 +27,7 @@ export const Signup = ({ createOrganizationTrial, currentUserId, loginUser, setF
   const [organization, setOrganization] = useState('');
   const [tos, setTos] = useState(false);
   const [redirectOnLogin, setRedirectOnLogin] = useState(false);
+  const campaign = match?.params?.campaign;
```
mzedel, line 30: `match` and `params` should always be defined, due to the router wrapping the `Signup` component - but these conditions won't hurt anyone 🙂

mzedel (approved):

src/js/components/user-management/signup.js
tranchitella: @mzedel they are not in unit tests, though... can I fix it somehow?

src/js/components/user-management/signup.test.js
```diff
@@ -19,7 +19,7 @@ describe('Signup Component', () => {
     const { baseElement } = render(
       <MemoryRouter>
         <Provider store={store}>
-          <Signup location={{ state: { from: '' } }} />
+          <Signup location={{ state: { from: '' }, match: { params: { campaign: '' } } }} />
```
mzedel, line 22: @tranchitella the location is a misleading copy&paste remnant from the login component test
```suggestion
          <Signup match={{ params: { campaign: '' } }} />
```
should work & allow the access to the match without conditions

tranchitella: Thanks!

PR #1509, opened 2021-04-14
MEN-4609 ensured ungrouped group won't suggest removing device from group

mzedel:
closing: MEN-4609
Changelog: Title


merlin-northern (approved):
works like a charm! checked adding and removing a device from group, the removal option is not present for an ungrouped device.

PR #1551, opened 2021-04-23
use tfa endpoints

mzedel:


mchalski (commented):
lgtm, the new WIP endpoints are exactly this - enable/disable without any POST bodies

tranchitella (approved):

PR #1603, opened 2021-05-12
made running requirement to run acceptance tests optional before publishing images

mzedel:
- the requirement is needed to prevent untested images from being published
- however, due to the reliance on the imported templates, we can't switch to dependencies, thus making the acceptance test need optional
=> hoping that this still requires them to run if the job is applicable (not excluded by only/except)

==> NB! we need to double check (after merging this) if the publish job also runs if the acceptance test job fails (based on the [suggested implementation in Gitlab](https://gitlab.com/gitlab-org/gitlab/-/issues/30680#note_514181529) I think it would run)... if it runs I honestly would suggest reverting to dependencies for the publishing jobs in the template @lluiscampos.

Changelog: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -177,12 +177,14 @@ publish:image:
     - job: build:docker
       artifacts: true
     - job: test:acceptance
+      optional: true
 
 publish:image:mender:
   needs:
     - job: build:docker
       artifacts: true
     - job: test:acceptance
+      optional: true
 
 publish:tests:
   stage: publish


tranchitella (approved):
@mzedel let's give it a try, and see if it unlocks staging

mzedel:
Thanks @tranchitella, that fixed the staging pipeline! The `optional` argument also behaves just as intended - to test I cancelled the e2e tests and the following publishing jobs did not start 👌.

lluiscampos:
> ... The `optional` argument also behaves just as intended - to test I cancelled the e2e tests and the following publishing jobs did not start ok_hand.

I am not sure if cancelling a job is equivalent to a failure, though.

> if it runs I honestly would suggest reverting to dependencies for the publishing jobs in the template @lluiscampos.

For future reference,

When we did the switch from needs to dependencies, the focus was on faster pipelines. Note that in mender-qa pipeline (used for releases) we use `dependencies` for the publish jobs and `needs` for everything else, so that we accomplish exactly what you 
suggest.

I agree with you that we should have the same approach in the template, so that we don't rely on users (i.e. gui repo) knowing about this detail when designing the pipeline. I will follow up partially reverting:
* https://github.com/mendersoftware/mendertesting/pull/141
* https://github.com/mendersoftware/sre-tools/pull/124
* https://github.com/mendersoftware/workflows/pull/120
  * and enterprise fork + release branches

lluiscampos:
> > ... The `optional` argument also behaves just as intended - to test I cancelled the e2e tests and the following publishing jobs did not start ok_hand.
> 
> I am not sure if cancelling a job is equivalent to a failure, though.
> 
> > if it runs I honestly would suggest reverting to dependencies for the publishing jobs in the template @lluiscampos.
> 
> For future reference,
> 
> When we did the switch from needs to dependencies, the focus was on faster pipelines. Note that in mender-qa pipeline (used for releases) we use `dependencies` for the publish jobs and `needs` for everything else, so that we accomplish exactly what you
> suggest.
> 
> I agree with you that we should have the same approach in the template, so that we don't rely on users (i.e. gui repo) knowing about this detail when designing the pipeline. I will follow up partially reverting:
> 
>     * [mendersoftware/mendertesting#141](https://github.com/mendersoftware/mendertesting/pull/141)
> 
>     * [mendersoftware/sre-tools#124](https://github.com/mendersoftware/sre-tools/pull/124)
> 
>     * [mendersoftware/workflows#120](https://github.com/mendersoftware/workflows/pull/120)
>       
>       * and enterprise fork + release branches

PRs:
* https://github.com/mendersoftware/mendertesting/pull/151
* https://github.com/mendersoftware/sre-tools/pull/136
* https://github.com/mendersoftware/workflows/pull/134
* enterprise fork + release branches _TODO_

PR #1604, opened 2021-05-12
Use deployment sorting

mzedel:


tranchitella (approved):
It works fine, the only thing which bothers me is that the date filters are automatically populated with today's date, both for from and to, despite the fact the list is not limited to today.

mzedel:
Thanks for looking into this @tranchitella! The filter reset should now be fixed and also the past deployment pagination is now enabled, independent of the number of deployments following a recommendation by @michaelatmender a looong time ago 🙈.

tranchitella:
@mzedel let's merge it!

tranchitella (approved):

PR #1631, opened 2021-05-21
MEN-4706 added tracking event for device configuration application

mzedel:
Changelog: None


tranchitella (approved):

estenberg:
Do you know if it has been deployed yet? If yes, there might be a problem because I can't see any data from it.

mzedel:
> Do you know if it has been deployed yet? If yes, there might be a problem because I can't see any data from it.

not yet, but I don't think there is anything preventing us from deploying it this week 👌

estenberg:
Merci :+1: 

Not a big rush, just wanted to make sure it wasn't broken/I did something wrong. :)

PR #1643, opened 2021-05-25
made deployments fitlerable by deployment type

mzedel:
closing: men-4667
Changelog: Title

### DIFF
--- a/src/js/components/deployments/__snapshots__/pastdeployments.test.js.snap
+++ b/src/js/components/deployments/__snapshots__/pastdeployments.test.js.snap
@@ -158,6 +158,95 @@ exports[`PastDeployments Component renders correctly 1`] = `
         </div>
       </div>
     </div>
+    <div
+      aria-expanded="false"
+      class="MuiAutocomplete-root MuiAutocomplete-hasClearIcon MuiAutocomplete-hasPopupIcon"
+      role="combobox"
+    >
+      <div
+        class="MuiFormControl-root MuiTextField-root MuiFormControl-fullWidth"
+        style="margin-top: 0px;"
+      >
+        <label
+          class="MuiFormLabel-root MuiInputLabel-root MuiInputLabel-formControl MuiInputLabel-animated"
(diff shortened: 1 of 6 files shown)


src/js/components/deployments/__snapshots__/pastdeployments.test.js.snap
```diff
@@ -158,6 +158,95 @@ exports[`PastDeployments Component renders correctly 1`] = `
...
+        </label>
+        <div
+          class="MuiInputBase-root MuiInput-root MuiInput-underline MuiAutocomplete-inputRoot MuiInputBase-fullWidth MuiInput-fullWidth MuiInputBase-formControl MuiInput-formControl MuiInputBase-adornedEnd"
+        >
+          <input
+            aria-autocomplete="list"
+            aria-invalid="false"
+            autocapitalize="none"
+            autocomplete="off"
+            class="MuiInputBase-input MuiInput-input MuiAutocomplete-input capitalized MuiAutocomplete-inputFocused MuiInputBase-inputAdornedEnd"
```
michaelatmender, line 186: The "capitalized" class can be removed, it's forcing the placeholder text to be bit inconsistent

michaelatmender (changes requested):
Looks good, only 2 small details I noticed: 
The '0 results' text here can be changed to "Try adjusting the filters"
empty text edit

And the comment inline regarding "Select A Type" capitalization ("Select a type" would be consistent with the other inputs)

src/js/components/deployments/__snapshots__/pastdeployments.test.js.snap
mzedel: 👌 now it will only appear capitalized once there actually is text entered, so the placeholder will show up consistent with the rest

mzedel:
Thanks for going through this 👍 - I adjusted the empty state as you suggested and did the same for the auditlogs...

michaelatmender (approved):

PR #1698, opened 2021-06-11
fixed an issue that could prevent deployment creation after switching to a lower plan

mzedel:
closing: men-4765
Changelog: Title

### DIFF
--- a/src/js/components/deployments/createdeployment.js
+++ b/src/js/components/deployments/createdeployment.js
@@ -13,6 +13,7 @@ import { selectDevice } from '../../actions/deviceActions';
 import { selectRelease } from '../../actions/releaseActions';
 import { advanceOnboarding } from '../../actions/onboardingActions';
 import { saveGlobalSettings } from '../../actions/userActions';
+import { PLANS } from '../../constants/appConstants';
 import { UNGROUPED_GROUP } from '../../constants/deviceConstants';
 import { onboardingSteps } from '../../constants/onboardingConstants';
 import { getIsEnterprise, getOnboardingState } from '../../selectors';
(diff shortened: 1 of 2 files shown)


tranchitella (approved):

src/js/components/deployments/createdeployment.js
```diff
@@ -99,9 +100,20 @@ export class CreateDialog extends React.Component {
...
+      isEnterprise,
+      isHosted,
+      isOnboardingComplete,
+      onScheduleSubmit,
+      plan,
+      saveGlobalSettings
+    } = self.props;
     const { deploymentDeviceIds, device, filterId, group, phases, release, retries } = settings;
     const startTime = phases?.length ? phases[0].start_ts || new Date() : new Date();
+    const retrySetting = isEnterprise || (isHosted && plan !== PLANS.os.value) ? { retries } : {};
```
merlin-northern, line 116: I had this vision about an interface like that:
```js
class Plan { // or of course function ;-)
/* ... */
}
plan=new Plan(PLANS.os.value); // "pseudo" code, just to note that we have some plan
/* ... */
if (plan.canAccess(FEATURES.RETRIED_DEPLOYMENTS) { // canAccess does all the magic to check if the plan includes a feature
 /* retrySetting = ... */
}
```

mzedel: we have something similar [here](https://github.com/mendersoftware/gui/blob/master/src/js/selectors/index.js#L52-L85) - but I would need some focused time on that to move some more decisions in there...

merlin-northern: ah, yes I remember. right! thank you for taking time to answer.

PR #1711, opened 2021-06-15
jest update to jest 27 + increased test robustness

mzedel:


tranchitella (approved):
@mzedel this is way too advanced to really understand it without context.
I'm approving because the tests pass, and I can guess why the changes are in place, but I cannot say I'm able to really review it sorry.

PR #1712, opened 2021-06-15
MEN-4662 devicelist refactoring

mzedel:
mostly because of some internal requirements for the main task this should tackle:
- MEN-4662
- MEN-4517
- MEN-4674
- MEN-4696
- MEN-4496 (and with it the follow up comment from MEN-2703)
- MEN-4828

so please also check what's missing with these...


lluiscampos:
I run this PR together with latest client and backend from feature branches. At least on my machine seems to work pretty well:

image

michaelatmender (changes requested):
Thanks for all these adjustments @mzedel!

Much of it looks good, but testing this morning I had some trouble with the UI killing my browser when creating a dynamic group from a filter:

Screenshot 2021-07-02 at 09 54 43
Screenshot 2021-07-02 at 09 55 25

There was also still an issue with the slow typing creating and adding several 'hidden' rules to the filter:
image

Can you recreate these issues?

michaelatmender (changes requested):
This looks good now, just a couple of small things left I found when testing!

- The Save button for editing a dynamic groups didn't seem to save the changes 
  - The tooltip is good, but maybe the text can be adjusted slightly to: "Saved changes will not change the target devices of any ongoing deployments to this group, but will take effect for new deployments"
- I suggest we just hide the "clear filter" button from editing dynamic groups, as the UX gets a bit complicated (and changing existing groups will likely involve adjusting or adding just one condition at a time, not clearing the whole lot!). So "clear filters" would only be shown when using filters in the "All devices" group. Does that make sense?

Other than that, all good!

michaelatmender (approved):
With those latest changes this looks to be in pretty good shape! Thanks @mzedel

PR #1724, opened 2021-06-21
prevented misleading success responses on missing auditlog config from breaking the UI

mzedel:
Changelog: None


tranchitella (approved):
Works fine here!

PR #1767, opened 2021-07-07
Fix/3.0.x build2 issues.

mzedel:
closing:
- MEN-4862
- MEN-4856
- MEN-4858
- MEN-4861

### DIFF
--- a/src/js/components/deployments/deployment-report/phaseprogress.js
+++ b/src/js/components/deployments/deployment-report/phaseprogress.js
@@ -177,10 +177,14 @@ export const PhaseProgress = ({ className = '', deployment = {}, onAbort, onUpda
   const [shouldAbort, setShouldAbort] = useState(false);
   const [isLoading, setIsLoading] = useState(false);
 
-  const { id, stats } = deployment;
+  const { id, stats, update_control_map = {} } = deployment;
+  const { states = {} } = update_control_map;
   const { failures: totalFailureCount } = groupDeploymentStats(deployment);
 
   const status = getDeploymentState(deployment);
+  const currentPauseState = Object.keys(pauseMap)
+    .reverse()
+    .find(key => stats[key] > 0);
 
   useEffect(() => {
     if (!isLoading) {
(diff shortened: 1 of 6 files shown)


src/js/components/deployments/deployment-report/phaseprogress.js
```diff
@@ -196,13 +200,12 @@ export const PhaseProgress = ({ className = '', deployment = {}, onAbort, onUpda
...
     setIsLoading(true);
-    const currentPauseState = Object.keys(pauseMap)
-      .reverse()
-      .find(key => stats[key] > 0);
     setShouldContinue(false);
     onUpdateControlChange({ states: { [pauseMap[currentPauseState].followUp]: { action: 'continue' } } });
   };
 
+  const disableContinuationButtons =
+    isLoading || (status === deploymentDisplayStates.paused && states[pauseMap[currentPauseState].followUp].action === 'continue');
```
kacf, line 208: :+1:

src/js/components/deployments/deployment-report/phaseprogress.js
```diff
@@ -104,7 +104,7 @@ const ProgressChart = ({ deployment = {}, showDetails, style }) => {
     <>
       {!showDetails && (
         <>
-          Phase 1: {Math.round((device_count / max_devices) * 100)}% ({device_count} {pluralize('device', device_count)})
+          Phase 1: {Math.round((device_count / totalDeviceCount || 0) * 100)}% ({device_count} {pluralize('device', device_count)})
```
tranchitella, line 107: Maybe you wanted:
```suggestion
          Phase 1: {Math.round((device_count / (totalDeviceCount || 1)) * 100)}% ({device_count} {pluralize('device', device_count)})
```

?

tranchitella (approved):

src/js/components/deployments/deployment-report/phaseprogress.js
mzedel: while that would lead to the same, what I actually wanted to write was `((device_count / totalDeviceCount) || 0)`, to drive home the point that the entire expression should evaluate to _0_ - but `prettier` taught me that the calculation will evaluate before the boolean expression anyway... so to not mislead my future self I would like to point the counts to 0 where possible...

PR #1772, opened 2021-07-09
Fix/3.0.x build2 pt2

mzedel:
closing:
- men-4855
- some bugs without a ticket + a regression in the deployments report, which I introduced in the previous fix PR 😞

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -52,22 +52,6 @@ cat >/var/www/mender-gui/dist/env.js <<EOF
   }
 EOF
 
-cat >/var/www/mender-gui/dist/livechat.js <<EOF
-  window.HFCHAT_CONFIG = {
-    EMBED_TOKEN: '$HAPPYFOX_KEY',
-    ASSETS_URL: 'https://widget.happyfoxchat.com/v2/visitor'
-  };
-  (function () {
-    var scriptTag = document.createElement('script');
-    scriptTag.type = 'text/javascript';
-    scriptTag.async = true;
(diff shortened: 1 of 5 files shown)


entrypoint.sh
```diff
@@ -52,22 +52,6 @@ cat >/var/www/mender-gui/dist/env.js <<EOF
   }
 EOF
 
-cat >/var/www/mender-gui/dist/livechat.js <<EOF
```
merlin-northern, line 55: I verified, no left-overs:
```bash
[ fix/3.0.x-build2-pt2 ]
#grep -rniHF happy .
```

src/js/components/devices/authorized-devices.js
```diff
@@ -78,15 +78,10 @@ export const Authorized = props => {
...
   useEffect(() => {
-    onSelectionChange([]);
     clearAllRetryTimers(setSnackbar);
     if (!filters.length && selectedGroup && groupFilters.length) {
       setDeviceFilters(groupFilters);
     }
-    clearInterval(timer);
-    // no group, no filters, all devices
-    timer = setInterval(getDevices, refreshDeviceLength);
-    getDevices();
```
merlin-northern, line 94: so now it will only get once when mounted, right?

mzedel: well, both effects/ trigger were run when mounted and the second trigger also is run on any relevant state change afterwards... This first one was basically a duplicate (based on the previous implementation, where these effects were split up) + relying only on the second one allows triggering it from a parent component...

merlin-northern (commented):
lgtm, but I will defer to Fabio for his approval.

tranchitella (approved):

PR #1775, opened 2021-07-09
use deployment device pagination

mzedel:


src/js/actions/deploymentActions.js
```diff
@@ -113,28 +113,41 @@ export const getSingleDeploymentStats = id => dispatch =>
...
 
-export const getSingleDeployment = id => (dispatch, getState) =>
-  Promise.all([GeneralApi.get(`${deploymentsApiUrl}/deployments/${id}`), GeneralApi.get(`${deploymentsApiUrl}/deployments/${id}/devices`)]).then(responses => {
-    const [deploymentRes, deploymentDeviceListRes] = responses;
-    const storedDeployment = getState().deployments.byId[id] || {};
-    const deploymentDevices = storedDeployment.devices || {};
-    const devices = deploymentDeviceListRes.data.reduce((accu, item) => {
+export const getDeploymentDevices = (id, options = {}) => (dispatch, getState) => {
+  const { page = defaultPage, perPage = defaultPerPage } = options;
+  return GeneralApi.get(`${deploymentsApiUrl}/deployments/${id}/devices/list?deployment_id=${id}&page=${page}&per_page=${perPage}`).then(response => {
```
tranchitella, line 118: From the backend PoV, there is no need to pass in thee `deployment_id` GET parameter.
Why are you sending it?

tranchitella (commented):
It works fine, but I have two doubts. One is in a comment, the other one is the fact that when I open the deployments details, I see multiple API calls to `/list`. The component performs the API call twice, and then a third time after about one second. Why?

Screenshot 2021-09-30 at 06 58 04

src/js/actions/deploymentActions.js
mzedel: oops, I think I misread the docs there and treated the path parameter as a query parameter 😞

mzedel:
the calls are triggered by more data rolling in about the state of the deployment... especially the newly retrieved statistics trickle trigger a slightly delayed response 😕 - I'll delay the listing for a bit...

tranchitella (approved):

PR #1785, opened 2021-07-12
Select experimental channel for get.mender.io on pre-releases

lluiscampos:
On a pre-release (iow testing a release candidate) the corresponding
client components are in the experimental channel of the APT repository.


lluiscampos:
This is a request for comments.

I don't know if it is a good idea or not. O the one hand it will help our manual testing of releases (we lost one day this time due to this) but on the other hand introduces a code difference between what we are testing and what we will get out once tagged which is an small risk.

@mzedel Also, I don't know neither if it is correct :man_shrugging:  I couldn't trace where this `isPreRelease` is set...

mzedel (commented):
Is this not doubled then or in other words: what would we need the routing to staging for? So far the only situation that staging is added is if it `isPreRelease`...
My assumption would be to expect any pre-release software to sit under staging - same with the installation script and the corresponding client (so targeting a `next` release).

mzedel:
@lluiscampos the [`isPreRelease` ](https://github.com/mendersoftware/gui/blob/master/src/js/components/common/dialogs/physicaldeviceonboarding.js#L189) is based on the `$INTEGRATION_VERSION` that is passed to the container on startup, so that it only shows for `next` releases, as we also sometimes have areas that should show only from a specific version (e.g. `2.6.2`)

kacf:
:warning: Illegal pull request detected! Submission from author not valid during this time range! :warning: 

But seriously, the way I read `isPreRelease`, it will never be set for a `3.0.0-build3` type build, because `next` always compares higher. So will this actually accomplish anything on a release candidate?

mzedel:
> But seriously, the way I read `isPreRelease`, it will never be set for a `3.0.0-build3` type build, because `next` always compares higher. So will this actually accomplish anything on a release candidate?

Yes, this needs to be marked by a change to the saas settings - as this might be tested the entire time before a release is tagged (so `next` targets anything after the most recent release). A beta release, releases candidate, ... and any changes after a release should in almost all cases (the only exception might be breaking changes) be covered by this.

lluiscampos:
> Is this not doubled then or in other words: what would we need the routing to staging for? So far the only situation that staging is added is if it `isPreRelease`...

Got it. In other words, once we have a configuration to use `get.mender.io/staging` instead of the production one, that same configuration should use the `-c experimental` and, generally speaking, every other tweak to get the bleeding edge software.

(I got confused originally between the "staging" concept of Hosted Mender, and the "staging" of the get.mender.io script, but I see we have two different tests for that so we are treating them differently).

@mzedel please review/approve the actual code and let's merge it for the next release :)

lluiscampos:
> warning Illegal pull request detected! Submission from author not valid during this time range! warning

@kacf I promise I wrote this one liner while with sun glasses on and enjoying myself :sunglasses: :beach_umbrella:

mzedel:
> Got it. In other words, once we have a configuration to use `get.mender.io/staging` instead of the production one, that same configuration should use the `-c experimental` and, generally speaking, every other tweak to get the bleeding edge software.
> (I got confused originally between the "staging" concept of Hosted Mender, and the "staging" of the get.mender.io script, but I see we have two different tests for that so we are treating them differently).

That would be the way to go in my opinion, aligning the script with all the other parts of staging, that always use the bleeding edge version... That should also limit the confusion on what is happening when running anything from staging (which in case of the client results in a "stale" version being retrieved for the latest backend/frontend combination).

> @mzedel please review/approve the actual code and let's merge it for the next release :)

With this, the content of this PR is not needed, but rather work on the script... we could maybe add parameter handling to the script, to allow passing versions to the script and install the correct version... If this is the case, the distinction between `get.mender.io` and `get.mender.io/staging` could be removed, and then maybe the `-c experimental` flag added, but I think a `--version next` or `--version 3.0.0` handling would be cleaner (and could go along with MEN-4452).

mzedel (approved):

lluiscampos:
> With this, the content of this PR is not needed, but rather work on the script... we could maybe add parameter handling to the script, to allow passing versions to the script and install the correct version... If this is the case, the distinction between `get.mender.io` and `get.mender.io/staging` could be removed, and then maybe the `-c experimental` flag added, but I think a `--version next` or `--version 3.0.0` handling would be cleaner (and could go along with MEN-4452).

Discussed in person. Basically: we cannot get the rid of `get.mender.io/staging` (it is a different script) and we cannot use "integration" versions in the script as it only understands stable/master kind of versions (APT repository channels)

@mzedel See my latest rework, as I had to rewrite it based on #1956

lluiscampos:
@mzedel See my new commit also - if I understood you correctly, we might have a demo installation locally with integration version set to "next" and then we want to have the edge software for the client too (for example today that would imply installing `mender-monitor`).

mzedel (approved):

PR #1794, opened 2021-07-12
unified snapshot tests to use rtl rendering

mzedel:
Changelog: None


tranchitella (approved):
All I can do here is stating that the tests pass... the changes are really too much to be reviewed.

PR #1801, opened 2021-07-16
Fix/onboarding tracking

mzedel:
depends on #1802

### DIFF
--- a/src/js/components/artifacts/dialogs/addartifact.js
+++ b/src/js/components/artifacts/dialogs/addartifact.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 
 import { Button, Dialog, DialogActions, DialogContent, DialogTitle } from '@material-ui/core';
 import ArtifactUpload from './artifactupload';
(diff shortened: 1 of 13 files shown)


src/js/components/artifacts/dialogs/addartifact.js
```diff
@@ -12,42 +12,46 @@ const steps = [
...
-    super(props, context);
-    this.state = {
-      activeStep: 0,
-      customDeviceTypes: '',
-      destination: !props.onboardingState.complete ? '/var/www/localhost/htdocs' : '',
-      file: null,
-      selectedDeviceTypes: []
-    };
-  }
+export const AddArtifactDialog = ({
```
tranchitella, line 15: I'm happy to see we are moving to functional components!

tranchitella (approved):

src/js/components/artifacts/dialogs/addartifact.js
mzedel: we're down to less than 20 class based components by now - unfortunately another small batch is waiting in the monitoring feature branch - but soon 🎉 (and related: #1803)

PR #1813, opened 2021-07-19
Remove menderDebPackageVersion

lluiscampos:
This is not in use anymore. During on-boarding we use get.mender.io
script to figure out which versions to install, so it is unlikely that
we need this setting in the future.


mzedel (approved):
Excellent, thank you!

PR #1814, opened 2021-07-20
MEN-4755 - device tags & device name

mzedel:
the device naming comes with some caveats: 
since the alternative means to set a device name in the backend I think it is tolerable to handle this with notifications... mostly because setting the device name based on a substring of the deviceId might lead to collisions, especially for larger device fleets...
Screenshot 2021-07-22 at 13 55 13
Besides the notification when sorting, here is also how devices without a set device name are displayed:
Screenshot 2021-07-22 at 13 54 58
Similar for filtering: 
Screenshot 2021-07-22 at 17 02 20
And a note for the tag editor:
Screenshot 2021-07-22 at 16 28 20

### DIFF
--- a/src/js/actions/deviceActions.js
+++ b/src/js/actions/deviceActions.js
@@ -144,25 +144,27 @@ const mapFiltersToTerms = filters =>
 const mapTermsToFilters = terms => terms.map(term => ({ scope: term.scope, key: term.attribute, operator: term.type, value: term.value }));
 
 export const getDynamicGroups = () => (dispatch, getState) =>
-  GeneralApi.get(`${inventoryApiUrlV2}/filters?per_page=${MAX_PAGE_SIZE}`).then(({ data: filters }) => {
-    const state = getState().devices.groups.byId;
-    const groups = (filters || []).reduce((accu, filter) => {
-      accu[filter.name] = {
-        deviceIds: [],
-        total: 0,
-        ...state[filter.name],
-        id: filter.id,
-        filters: mapTermsToFilters(filter.terms)
-      };
-      return accu;
-    }, {});
-    return Promise.resolve(
-      dispatch({
-        type: DeviceConstants.RECEIVE_DYNAMIC_GROUPS,
-        groups
-      })
-    );
-  });
+  GeneralApi.get(`${inventoryApiUrlV2}/filters?per_page=${MAX_PAGE_SIZE}`)
+    .then(({ data: filters }) => {
+      const state = getState().devices.groups.byId;
+      const groups = (filters || []).reduce((accu, filter) => {
+        accu[filter.name] = {
+          deviceIds: [],
+          total: 0,
+          ...state[filter.name],
+          id: filter.id,
+          filters: mapTermsToFilters(filter.terms)
+        };
+        return accu;
+      }, {});
+      return Promise.resolve(
+        dispatch({
+          type: DeviceConstants.RECEIVE_DYNAMIC_GROUPS,
+          groups
+        })
+      );
+    })
+    .catch(() => console.log('Dynamic group retrieval failed - likely accessing a non-enterprise backend'));
 
 export const addDynamicGroup = (groupName, filterPredicates) => (dispatch, getState) =>
   GeneralApi.post(`${inventoryApiUrlV2}/filters`, { name: groupName, terms: mapFiltersToTerms(filterPredicates) })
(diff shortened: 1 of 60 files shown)


estenberg (commented):
Wow this (device name) was fast, maybe too fast? :)
Maybe it would require some more discussion -- we do have design task https://tracker.mender.io/browse/MEN-4778 which is not completed.

A few comments/questions:
* Is the device name *not* stored in the backend? Based on your comment it seems so. It must be in the backend I think, otherwise it will be confusing to get data from the API vs. UI.
* Using a convention of the Tag "Name" and handle that specially in the UI is good. Aligned with other services, e.g. AWS.
* There was some discussion about emtpy name, and that it wasn't good design wise. However AWS does default to an empty name, so I don't think it can be a big problem.
* These tooltips are not necessary I think. It makes it very clear, but I would expect that users already expect this behavior (e.g. if you haven't set the name, then filtering on name won't find that device).

mzedel:
> * Is the device name _not_ stored in the backend? Based on your comment it seems so. It must be in the backend I think, otherwise it will be confusing to get data from the API vs. UI.
the name is stored in the backend as a tag (a tag with the key 'name')

> * There was some discussion about emtpy name, and that it wasn't good design wise. However AWS does default to an empty name, so I don't think it can be a big problem.
the empty name here is shown as the placeholder in the input (which is the beginning of the device id) - however, when acting on the empty names (through filtering, sorting) the results appear misleading... the device list gets sorted in another way, filtering does not return the devices with the displayed (placeholder) names => thus the tooltips
we could maybe generate names in the same style as e.g. docker does for started containers (`elated_margulis`) in a migration in the backend - depending on the number of devices (+ the used dictionary) that should give a reasonable number of identifiers and remain readable...

> * These tooltips are not necessary I think. It makes it very clear, but I would expect that users already expect this behavior (e.g. if you haven't set the name, then filtering on name won't find that device).
to me personally that just seemed odd when sorting, due to the displayed names (and then the rest was added for completeness) OTOH having many devices with empty names looked broken...

estenberg:
Alright, then it makes more sense. :tulip: 

Should get input from @michaelatmender  too. If it's ok then we can simply close those two last tasks in the name Epic.

src/js/actions/deviceActions.js
```diff
@@ -806,3 +811,16 @@ export const applyDeviceConfig = (deviceId, configDeploymentConfiguration, isDef
       }
       return Promise.all(tasks);
     });
+
+export const setDeviceTags = (deviceId, tags) => dispatch =>
+  // to prevent tag set failures, retrieve the device & use the freshest etag we can get
+  Promise.resolve(dispatch(getDeviceById(deviceId))).then(device => {
+    const headers = device.etag ? { 'If-Match': device.etag } : {};
```
tranchitella, line 818: The `etag` is not in the device object, but in the `etag` HTTP header.
Because of this, the UI is not sending the `If-Match` header on PUT.

tranchitella (changes requested):
One important observation about the Etag management, which I think is wrong.
On a side note, would it be possible to filter devices by tag? I see the backend returns them from `/api/management/v2/inventory/filters/attributes` but the UI only shows "Name" among the tags. It would be important for the users to be able to filter by tags.

src/js/actions/deviceActions.js
mzedel: [the `getDeviceById` call](https://github.com/mendersoftware/gui/pull/1814/files/0727e9298b913f573392b2a9b8ffbeb0b8ebe878#diff-74f29c7fa70ecbddf5cd2b41184645057374578e6b26e9a4ff4ff49d3cabb157R402) sets the etag on the device and [returns it](https://github.com/mendersoftware/gui/pull/1814/files/0727e9298b913f573392b2a9b8ffbeb0b8ebe878#diff-74f29c7fa70ecbddf5cd2b41184645057374578e6b26e9a4ff4ff49d3cabb157R405) to be able to rely on an always up-to-date etag (as it might soon become stale if the UI relied on the storage)... - so if the etag is present in the response from inventory it _should_ be sent during PUT

mzedel:
> On a side note, would it be possible to filter devices by tag? I see the backend returns them from `/api/management/v2/inventory/filters/attributes` but the UI only shows "Name" among the tags. It would be important for the users to be able to filter by tags.

👌 sure, I'll add this (I suppose the tags are also returned when calling the `/api/management/v2/inventory/filters/attributes` endpoint)

src/js/actions/deviceActions.js
tranchitella: Oh, sorry, my fault. When testing locally, it never sent the `If-Match` header, though.

tranchitella (approved):
Wonderful

michaelatmender (approved):
This all looks good, caveats included. The 'default device name' vs 'assigned device name' might be something to monitor with users though, and adjust later if any issues arise. But your approach makes sense now!

PR #1815, opened 2021-07-21
fixed an issue that caused the deployment report to crash due to missing update control information

mzedel:
- this also adds a note to the report to indicate that triggering a continuation through the UI is not possible
Changelog: Title

### DIFF
--- a/src/js/components/deployments/deployment-report/phaseprogress.js
+++ b/src/js/components/deployments/deployment-report/phaseprogress.js
@@ -177,9 +177,9 @@ export const PhaseProgress = ({ className = '', deployment = {}, onAbort, onUpda
   const [shouldAbort, setShouldAbort] = useState(false);
   const [isLoading, setIsLoading] = useState(false);
 
-  const { id, stats, update_control_map = {} } = deployment;
+  const { id, stats = {}, update_control_map = {} } = deployment;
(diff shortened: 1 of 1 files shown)


src/js/components/deployments/deployment-report/phaseprogress.js
```diff
@@ -204,17 +204,19 @@ export const PhaseProgress = ({ className = '', deployment = {}, onAbort, onUpda
     onUpdateControlChange({ states: { [pauseMap[currentPauseState].followUp]: { action: 'continue' } } });
   };
 
-  const disableContinuationButtons =
-    isLoading || (status === deploymentDisplayStates.paused && states[pauseMap[currentPauseState].followUp].action === 'continue');
+  const isPaused = status === deploymentDisplayStates.paused;
+  const canContinue = isPaused && states[pauseMap[currentPauseState].followUp];
+  const disableContinuationButtons = isLoading || (canContinue && states[pauseMap[currentPauseState].followUp].action === 'continue');
```
kacf, line 209: I would use `!= 'pause'` rather than `== 'continue'`. The reason is that you can also set it to `fail` or `force_continue` using the API.

kacf (approved):
One small adjustment I would make, otherwise it looks good to me!

PR #1834, opened 2021-07-26
MEN-4867 restrict remote terminal, port forward visibility by user roles

mzedel:


src/js/components/devices/troubleshootdialog.js
```diff
@@ -40,8 +40,8 @@ const ConnectionIndicator = isConnected => {
 };
 
 const tabs = {
-  terminal: { link: 'session logs', title: ConnectionIndicator, value: 'terminal' },
-  transfer: { link: 'file transfer logs', title: () => 'File transfer', value: 'transfer' }
+  terminal: { link: 'session logs', title: ConnectionIndicator, value: 'terminal', needsWriteAccess: true, needsTroubleshoot: true },
+  transfer: { link: 'file transfer logs', title: () => 'File transfer', value: 'transfer', needsWriteAccess: false, needsTroubleshoot: false }
```
tranchitella, line 44: Why file transfer doesn't need troubleshoot?

mzedel: Yeah, that's not ideal... Troubleshoot is needed on the device and in the account, but downloading a file should also work for read only users... I think the access restriction might be a bit too fine granular in this case, but on the other hand I think it is a valid point...

michaelatmender (changes requested):
This seems to cause an issue on the device view:

> accepted devices couldn't be loaded. failed to decode request body: json: cannot unmarshal object into Go struct field SelectAttribute.attributes.attribute of type string... [Request ID: f1beead1]

mzedel:
@michaelatmender can you try this again? I think that branch didn't yet include the name/ tags changes and thus sent the "new" default id attribute object (instead of just the name) as part of the default attributes to list devices... I rebased the branch and I think it should now work 🤞

michaelatmender (approved):

PR #1838, opened 2021-07-27
MEN-4915

mzedel:


tranchitella (approved):
Thanks a lot! Is this PR address MEN-4921 as well?

mzedel:
@tranchitella the 3. point in the ticket should be addressed with this, and with the new commit the 2. point also - at least for the session related cookies... afaik the JWT cookie is only set from the UI and not in the backend and I'm not aware of any other cookies that are in use in the communication... as for the first of the points - I'm not 100% sure what exactly this refers to... otherwise I'll happily add it to this PR

tranchitella:
@mzedel you can find the security audits as attachments here: https://tracker.mender.io/browse/MEN-4913

tranchitella (approved):

mzedel:
response header analysis [before this PR](https://securityheaders.com/?q=hosted.mender.io&hide=on&followRedirects=on) and [after this PR](https://securityheaders.com/?q=staging.hosted.mender.io&hide=on&followRedirects=on)...

PR #1851, opened 2021-07-30
MEN-4891 switched plan change request to use dedicated endpoint

mzedel:
depends on mendersoftware/tenantadm#314

Changelog: None


tranchitella (approved):
Reviewed and tested in combination with backend.

PR #1870, opened 2021-08-03
fixed an issue that caused the device list to misbehave when quickly changing pages

mzedel:
closing: men-4932
Changelog: title


tranchitella (approved):
Looks good, tested locally and solves the issue.

PR #1871, opened 2021-08-03
made device name placeholder connection to device id a bit more visible

mzedel:
Changelog: None


michaelatmender (approved):
Yes, this is good - I would also make the font a bit darker too. I can make a separate PR though

michaelatmender (approved):

PR #1887, opened 2021-08-13
MEN-4928: improvements to the monitoring UI

tranchitella:
- Improve margins and layout
- Use link instead of button to view logs
- Show the "view logs" link only if there are logs
- Use the right end-point (/alerts/latest)

Changelog: none


merlin-northern (approved):
I kind of lost track why /alerts was used. I thought it was because it did/suppose to do the proper grouping, but if /latest does the trick, then I am fine with that.

tranchitella:
> I kind of lost track why /alerts was used. I thought it was because it did/suppose to do the proper grouping, but if /latest does the trick, then I am fine with that.

`/alerts` list the records, `/alerts/latest` does the grouping.

tranchitella:
@michaelatmender can you double check it's okay?

michaelatmender (approved):
This looks good, now I can test it at last!

tranchitella:
@michaelatmender thanks a lot!

PR #1886, opened 2021-08-12
Show active and pending authset by default

michaelatmender:

PR #1888, opened 2021-08-13
Open authsets DataCollapse component by default if pending or a new r…

michaelatmender:
…equest MEN-4800

Changelog: None


tranchitella (approved):
Works fine, and I find it a better solution than #1886.
LGTM!

PR #1895, opened 2021-08-16
Adjust help text to say that RPi image is merely Mender ready.

kacf:
Changelog: Title

### DIFF
--- a/src/js/components/common/dialogs/__snapshots__/physicaldeviceonboarding.test.js.snap
+++ b/src/js/components/common/dialogs/__snapshots__/physicaldeviceonboarding.test.js.snap
@@ -1198,7 +1198,7 @@ label + .MuiInput-formControl {
           class="margin-top"
         >
           <p>
-            We prepared an image, fully integrated with Mender for you to start with. You can find it in the
+            We prepared an image, ready for Mender, for you to start with. You can find it in the
              
             <a
               href="https://docs.mender.io/get-started/preparation/prepare-a-raspberry-pi-device"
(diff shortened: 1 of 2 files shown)


kacf:
This follows from https://github.com/mendersoftware/mender-docs/pull/1518.

tranchitella (approved):

src/js/components/common/dialogs/__snapshots__/physicaldeviceonboarding.test.js.snap
```diff
@@ -1198,7 +1198,7 @@ label + .MuiInput-formControl {
           class="margin-top"
         >
           <p>
-            We prepared an image, fully integrated with Mender for you to start with. You can find it in the
+            We prepared an image, ready for Mender, for you to start with. You can find it in the
```
oleorhagen, line 1201: Nitpicky, I know, but `ready for Mender` sounds a little odd.
How about: `ready to be integrated with the Mender client`, or simply `Mender`?

kacf: I actually had that exact sentence as a candidate, but I wanted to avoid the word "integrated", because it can be confused with "Mender integration". This is the act of setting up the partitions, etc, which is already done here, so it's a bit ambiguous. Another candidate was "ready for Mender to be installed", but this sentence was also a bit awkward to me, and didn't really add anything, so I decided to follow Lluis' original suggestion in https://github.com/mendersoftware/mender-docs/pull/1518, which is what you see above.

PR #1896, opened 2021-08-17
Update feature toggles to respect env variables

tranchitella:
If `MENDER_HOSTED` is set, fature toggles do not respect the environment
variables and rely on the tenants' capabilities (enabled add-ons) only.
This makes impossible to disable a feature in Hosted Mender when the
backend already supports a new add-on.

Make the feature toggles always respect the environment variables.

Changelog: none


merlin-northern (approved):
lgtm

PR #1897, opened 2021-08-18
MEN-4922: require the current one when updating the user's password

tranchitella:
Changelog: title

### DIFF
--- a/src/js/components/user-management/__snapshots__/userform.test.js.snap
+++ b/src/js/components/user-management/__snapshots__/userform.test.js.snap
@@ -74,7 +74,7 @@ exports[`UserForm Component renders correctly 1`] = `
             />
           </div>
           <div
-            class="edit-pass margin-top-small"
+            class="edit-pass"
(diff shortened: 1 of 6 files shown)


michaelatmender (commented):
This looks good for changing the user's own password!
However, I'm wondering if this breaks the use case for an admin account to be able to change other users' passwords within the 'user management' page? Either for an easy password reset, or to temporarily block access to a user, for example.

tranchitella:
@michaelatmender it makes sense. On the other hand, the security audit we received stated explicitly we should not allow resetting the password without knowing the previous one (e.g. you leave your browser with an active session, and a malicious user can reset the password(s) of the user(s) to gain access). I don't have a strong opinion, I just want to be sure we solve the item in the security audit.

tranchitella:
@michaelatmender ready for another round of review, reworked as discussed to offer the possibility to trigger a "password reset" email for users.

michaelatmender (approved):

PR #1902, opened 2021-08-21
MEN-4967: use booleans (true/false) as values for the $exists filter

tranchitella:
Changelog: none


mchalski (approved):
verified with reporting, lgtm

mzedel (approved):

PR #1903, opened 2021-08-22
MEN-4945: prevent non-equality filters to be added to the query param…

tranchitella:
…eters

This is important for the filters on `id`, as the query parameter overrides the
active filters, making it impossible to filter by `id` using an operator other
than `$eq`.

Changelog: none


mzedel (commented):
This is a bit tricky, as the URL was requested to reflect the filter state to allow pointing to the same devices through a shared link, however this was when there only was equality based filtering... IMO it might be worth getting the filter operators into the query params (and fall back to equality for all queries without a defined operator)...
I _think_ the reset to equality for id based filters lies somewhere else (my suspicion is the initial param parsing on initial mount, but I'll need to check...)

tranchitella:
@mzedel can you take over this?

mzedel:
> @mzedel can you take over this?

Sure! for now I'm still playing catch up, but will get to it asap!

mzedel:
@tranchitella this works now with the id attribute - I'm not yet settled on how to structure the query for multiple attributes & different operators (without the query getting ugly quick), so I'm postponing this to an extra task...

PR #1924, opened 2021-08-24
removed references to gulp to standardize on npm based usage

mzedel:
- also to address some vulnerability notifications & speed up dependency installation

Changelog: None


mzedel:
@michaelatmender after merging this, the watching command would be limited to `npm run watch` and the non-docker build command `npm run build`

michaelatmender (approved):

PR #1933, opened 2021-08-27
MEN-5004 - roles changes

mzedel:
Changelog: None

### DIFF
--- a/src/js/actions/appActions.test.js
+++ b/src/js/actions/appActions.test.js
@@ -1,8 +1,7 @@
 import configureMockStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
 import { inventoryDevice } from '../../../tests/__mocks__/deviceHandlers';
-import { roles } from '../../../tests/__mocks__/userHandlers';
-import { defaultState } from '../../../tests/mockData';
+import { defaultState, receivedPermissionSets, receivedRoles } from '../../../tests/mockData';
 
 import { commonErrorHandler, initializeAppData, setSnackbar, setFirstLoginAfterSignup, setYesterday, setVersionInfo } from './appActions';
 import AppConstants from '../constants/appConstants';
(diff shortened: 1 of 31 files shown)


michaelatmender (commented):
When trying to create a new role, I get the error "bad role definition: role name cannot be empty". Is it likely to be something in my local setup, or can you reproduce?

michaelatmender (changes requested):
Ran into an error when editing a role:
```
userActions.js:427 Uncaught TypeError: Cannot read properties of undefined (reading 'permissionSets')
    at userActions.js:427
```

I found something after creating and then logging in as a 2nd user - the username on the top right bar isn't loaded, showing an error on the call to:
`https://localhost/api/management/v1/useradm/users/me`
This user has 'read' permission on useradm

mzedel:
@michaelatmender the error you encountered should now be fixed, but the useradm issue you came across might be something for @kjaskiewiczz instead...
to reproduce it I created a role with just read rights for user management, which leads to the following being sent to the backend:
`{name: "usermgmt", description: "", permission_sets_with_scope: [{name: "Basic"}, {name: "ReadUsers"}]}`
IMO that should be enough to allow at least logging in.. The `ReadUsers` permission set might already be covered by the `Basic` permission set - do they maybe interfere @kjaskiewiczz?

michaelatmender:
Editing a role now gives "There was an error editing the role: role names do not match" and it looks like the payload only has 'description' and the permissions sets. Looks like the role name is missing!

mzedel:
The role name is included in the url path now - so might also be something for @kjaskiewiczz to try...

mzedel:
@kjaskiewiczz can you please give this another try now? and check the two issues described [here](https://github.com/mendersoftware/gui/pull/1933#issuecomment-1004825116) and [here](https://github.com/mendersoftware/gui/pull/1933#issuecomment-1004890057)?

kjaskiewiczz:
about editing roles and names mismatch - backend is checking if the name from the path is the same as the one from the body;
this was original behavior:
https://github.com/mendersoftware/useradm-enterprise/blob/a380c4679d5cec4750a13f61d921f64c06d12ac5/api/http/api_useradm_enterprise.go#L402
and I didn't change it:
https://github.com/mendersoftware/useradm-enterprise/blob/f3a2573ae553912a3d71ad40fa8b1a4a86f5241d/api/http/api_useradm_enterprise.go#L880

@mzedel you think we should change it?

tranchitella (approved):
I tested this PR while updating the RBAC documentation (https://github.com/mendersoftware/mender-docs/pull/1728). It works fine and I couldn't spot any issue. Amazing work.

tranchitella:
A few random suggestions:
- There is a typo (search for "priviledge")
- I would hide the "DELETE ROLE" button (top-right) for built-in roles; now it is disabled.

mzedel:
Thanks a lot for the review @tranchitella!!

> * There is a typo (search for "priviledge")
> * I would hide the "DELETE ROLE" button (top-right) for built-in roles; now it is disabled.
> * I don't know why "Read Access", "Deployments Manager" and "Troubleshooting" have "User Management: Read"; according to the backend code, they should not have it.

all these should now be addressed 👌

michaelatmender (approved):

PR #1956, opened 2021-09-02
MEN-4759: mender-monitor onboarding support

lluiscampos:
Changelog: Install Mender commercial add-ons (namely: mender-monitor)
during user on-boarding.

### DIFF
--- a/src/js/components/common/dialogs/physicaldeviceonboarding.js
+++ b/src/js/components/common/dialogs/physicaldeviceonboarding.js
@@ -110,8 +110,8 @@ export const DeviceTypeSelectionStep = ({ docsVersion, hasConvertedImage, onboar
   );
 };
 
-export const InstallationStep = ({ advanceOnboarding, ipAddress, isHosted, isEnterprise, token, selection, isPreRelease }) => {
-  const codeToCopy = getDebConfigurationCode(ipAddress, isHosted, isEnterprise, token, selection, isPreRelease);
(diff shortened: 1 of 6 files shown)


src/js/components/common/dialogs/physicaldeviceonboarding.js
```diff
@@ -110,8 +110,8 @@ export const DeviceTypeSelectionStep = ({ docsVersion, hasConvertedImage, onboar
   );
 };
 
-export const InstallationStep = ({ advanceOnboarding, ipAddress, isHosted, isEnterprise, token, selection, isPreRelease }) => {
-  const codeToCopy = getDebConfigurationCode(ipAddress, isHosted, isEnterprise, token, selection, isPreRelease);
+export const InstallationStep = ({ advanceOnboarding, ipAddress, isHosted, isEnterprise, tenantToken, selection, isPreRelease }) => {
+  const codeToCopy = getDebConfigurationCode(ipAddress, isHosted, isEnterprise, tenantToken, selection, isPreRelease);
```
lluiscampos, line 114: This renaming I did because first I thought I would need to pass `jwtToken` into this function.

I didn't need it, though, as I ended up requesting the token from inside the function. So if that is correct I can drop the rename if you want (it is done on a separate commit).

mzedel: Thanks, the clarification is very useful! Before this, I never expected the JWT to be actively passed somewhere "outside" 😅.

mzedel (approved):

PR #992, opened 2020-10-06
made device activity widget rely on backend filtering instead of client side deduction

mzedel:
Changelog: None
Signed-off-by: Manuel Zedel <manuel.zedel@northern.tech>

PR #1974, opened 2021-09-07
MEN-4996 - filter by monitoring issues

mzedel:
Changelog: None


src/js/constants/deviceConstants.js
```diff
@@ -109,6 +109,35 @@ module.exports = {
...
+          yesterday.setDate(yesterday.getDate() - 1);
+          return yesterday.toISOString();
+        }
+      },
+      title: 'Offline devices'
+    },
+    monitoring: {
+      key: 'monitoring',
+      needsReporting: false,
+      filterRule: { scope: 'monitoring', key: 'alerts', operator: '$gt', value: 0 },
```
tranchitella, line 131:
```suggestion
      filterRule: { scope: 'monitoring', key: 'alerts', operator: '$eq', value: true },
```

src/js/constants/deviceConstants.js
```diff
@@ -109,6 +109,35 @@ module.exports = {
...
+      title: 'Devices reporting monitoring issues'
+    },
+    authRequests: {
+      key: 'authRequests',
+      needsReporting: true,
+      filterRule: { scope: 'monitoring', key: 'auth_requests', operator: '$gt', value: 1 },
+      title: 'Devices with new authentication requests'
+    }
+  },
```
tranchitella, line 140: This is not implemented (as of today). In any case, it wouldn't be `monitoring` scope.

mzedel: In order to provide the functionality as it is defined in the design we should get it implemented though... But once we get the reporting service in we can likely adjust a few places anyway (e.g. finally getting #992 in)...

tranchitella: @mzedel does this block the release? we don't have ANY plan today to implement this feature in the backend before the introduction of the reporting service.

mzedel: nono, that doesn't block at all - this (and the offline one) are just in preparation (to check the general approach) and are filtered based on the `needsReporting: true` 👌

tranchitella (approved):
Looks okay, and works fine. The only (minor) thing which annoys me is that it makes the request to count the number of devices with +1 auth sets despite the fact reporting is not enabled. Can you double-check and make sure the "count" API request is not fired in that case?

mzedel:
> Looks okay, and works fine. The only (minor) thing which annoys me is that it makes the request to count the number of devices with +1 auth sets despite the fact reporting is not enabled. Can you double-check and make sure the "count" API request is not fired in that case?

Thanks for going through this @tranchitella! Sorry I missed that - ab71ebba7039d8333efff01a5c71d243b6fb9f2d includes the adjustment and shouldn't call before there is reporting support 👍

tranchitella (approved):

mzedel:
@tranchitella the device list should now show the monitoring notifications 👌 - please take a look at the added commits if you have the time and I'll take care of the update on staging once you give the green light

tranchitella:
@mzedel it works fine, the only minor issue is that on my screen the monitoring badge is not fully visible, see screenshot:

image

mzedel:
Yeah, there is MEN-4988 and I'll take care of it in that context asap... This should be pre-existing though and appear with pending devices too - I'll merge regardless and try to improve it before the end of the sprint.

PR #1977, opened 2021-09-09
MEN-4759: on-boarding: support for commercial components for Enterprise

lluiscampos:
The install script requires now a new flag `--auth` to specify the
server: hosted uses `--jwt-token` while on-prem uses `--username`.


lluiscampos:
Discussed on SU today - we won't implement this.

PR #2004, opened 2021-09-14
improved version information parsing to allow prerelease tests

mzedel:
Changelog: None

### DIFF
--- a/src/js/reducers/appReducer.js
+++ b/src/js/reducers/appReducer.js
@@ -29,6 +29,8 @@ const menderEnvironment = {
   ...mender_environment
 };
 
+const getComparisonCompatibleVersion = version => (isNaN(version.charAt(0)) && version !== 'next' ? 'master' : version);
+
 export const initialState = {
   cancelSource: undefined,
   demoArtifactLink: 'https://dgsbl4vditpls.cloudfront.net/mender-demo-artifact.mender',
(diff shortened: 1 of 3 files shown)


src/js/reducers/appReducer.js
```diff
@@ -46,7 +48,7 @@ export const initialState = {
     hasDeviceConnect: stringToBoolean(menderEnvironment.features.hasDeviceConnect),
     hasMonitor: stringToBoolean(menderEnvironment.features.hasMonitor),
     hasReporting: stringToBoolean(menderEnvironment.features.hasReporting),
-    isHosted: stringToBoolean(menderEnvironment.features.isHosted) || window.location.hostname === 'hosted.mender.io',
+    isHosted: stringToBoolean(menderEnvironment.features.isHosted) || window.location.hostname.includes('hosted.mender.io'),
```
lluiscampos, line 51: (unrelated to this PR, but commenting once I am here).

One different problem in staging.mender.io is exactly this: when `isHosted` is set to "true", then the bash snippet configures mender client with flag `--hosted-mender` which sets the `ServerUrl` to `hosted.mender.io`... which is not what we want.

This has a difficult solution though so I didn't really came up with a proposal (yet :grin: )

lluiscampos (approved):
I think I got it :)

src/js/reducers/appReducer.js
mzedel: hmm we could also pass the server-url from the GUI if the `isPreRelease` flag is set (since we already assume staging based on the flag)... not sure if this would work for `mender setup` though...

PR #2014, opened 2021-09-17
MEN-4696 - persist UI state during internal navigation

mzedel:


michaelatmender (changes requested):
Looks good! I noticed one small thing - the date range on deployments filtering doesn't persist if you select 'last 7 days' or 'last 30 days'. Other ranges and custom ranges work

mzedel:
@michaelatmender thanks for checking! It might still happen on the initial load of deployments (then the selected timeframe will be overwritten with the timeframe that is deducted from the last 20 deployments), but otherwise the tracking should be the same... However, one thing that is not persisted, is the range selection... I made some changes to better reflect this, so the selector will now detect if the timeframe selected fits one of the ranges and show that as selected. Can you please take another look?

tranchitella (approved):
It works fine, couldn't spot any issue

mzedel:
👌 thanks @tranchitella and @michaelatmender! I'll merge this to unblock the work on the URL tracking...

PR #2028, opened 2021-09-23
fixed an issue that prevented filter selections from being persisted

mzedel:
Changelog: title


merlin-northern (approved):
lgtm

PR #2050, opened 2021-09-27
MEN-5062

mzedel:


tranchitella (commented):
@mzedel I have the (wrong, maybe) impression that this is breaking the file download, which relies on the Cookie and is not in the `/ui` path. What do you think?

mzedel:
@tranchitella thanks for checking that! The e2e tests should catch that if it doesn't work and... for some reason it does work there (a diff from a setup with the `/ui` path limitation [here](https://gitlab.com/Northern.tech/Mender/gui/-/jobs/1629300142/artifacts/file/screenshots/diffs/diff-1632826750654.png)) - however it fails locally for me too 🧐...
Then I'll align the path in useradm and here to be `/`, since we rely on it also in the rest of the communication with the backend...

.gitlab-ci.yml
```diff
@@ -47,7 +47,6 @@ test:lint:
       - screenshots
       - logs
       - junit
-    when: on_failure
```
tranchitella, line 50: Is this a left-over, or intended change?

tranchitella (approved):

.gitlab-ci.yml
mzedel: That's intended to allow inspection of screenshots and replays even if a test was just flaky (or in this case passed, but the terminal diff would have been interesting).

PR #2051, opened 2021-09-28
MEN-5123

mzedel:


merlin-northern (approved):
lgtm -- tested with both service and log alert.

PR #2052, opened 2021-09-29
MEN-5118 added azure oauth option

mzedel:
Screenshot 2021-10-04 at 11 44 51

### DIFF
--- a/src/js/components/user-management/__snapshots__/login.test.js.snap
+++ b/src/js/components/user-management/__snapshots__/login.test.js.snap
@@ -77,6 +77,34 @@ exports[`Login Component renders correctly 1`] = `
         class="MuiTouchRipple-root"
       />
     </button>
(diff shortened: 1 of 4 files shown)


tranchitella (approved):

mzedel:
Thanks @tranchitella for taking a look! For now I'll leave it unmerged until the backend can be deployed too (I don't want to add another environment flag just for azure)...

alfrunes:
Technically, signup can now be performed by all Microsoft accounts, so maybe s/azure/microsoft/ ? We can ofcourse limit the scope to Azure tenants only, but I see no reason why we should.

mzedel:
Good point @alfrunes - I adjusted the config accordingly 👌

tranchitella:
@mzedel can we merge this PR now that backend support is available?

PR #2053, opened 2021-09-29
MEN-4994 - make distribution text stay outside of the chart

mzedel:
the main change is in 38cdc7f5e8ab9ac97732527266870f56c41b7346


tranchitella (approved):
Tested with a few different values/combinations, it looks good all the time.

PR #2085, opened 2021-10-04
Improve release handling

mzedel:


tranchitella:
The PR looks good, but when testing it, I noticed this bug:



It happens when you click "Create deployment with this release" from the "Releases" view.
It is not related to this PR, the problem is already there, but maybe you can fix it here?

mzedel:
#2090 should fix that particular issue 😉, but I can rebase to get it in here...

PR #2088, opened 2021-10-05
ensured 2fa state is usable for qr code retrieval after email activation

mzedel:
Changelog: None


merlin-northern (approved):
let's try

PR #2092, opened 2021-10-06
MEN-4788, MEN-4958, MEN-4959 + cleanup tooltips + classes

mzedel:
the react-tooltip removal is needed to allow keeping up with react releases - or change all of the tooltips to the next major version in these (which would also come with more code changes)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -50,6 +50,7 @@ test:lint:
     reports:
       junit:
         - junit/results.xml
+    when: on_failure
 
 test:acceptance:
   extends: .template:test:acceptance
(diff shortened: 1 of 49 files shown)


tranchitella (approved):
I went through the code and it makes sense. I also tested it locally, playing around with tooltips and onboarding (at least the first steps), and it works fine. While testing it, though, I discovered an unrelated issue:

https://github.com/mendersoftware/gui/blob/master/src/js/components/deployments/deployment-wizard/softwaredevices.js#L36

This link is wrong, shouldn't it be /releases?

I'm approving, I will do a full onboarding test on staging as soon as this is released.

mzedel:
thanks @tranchitella for looking at and catching this - that seems to be there since the introduction of the releases concept 🙈 - I adjusted the link now and there should no longer be any other references to `/artifacts` in the UI

tranchitella (approved):

michaelatmender (approved):

PR #2093, opened 2021-10-06
restored respect for ip addresses in deb config snippet in enterprises

mzedel:
Changelog: None


lluiscampos:
> @lluiscampos sure, I'll choose this one. One thing I would be very happy about would be to get the switch away from the positional arguments to the passed object in your PR. The long list of arguments is painful to use without the names imo...

Sure! I just had no idea that was possible :sweat_smile: I'll get the inspiration from here.

PR #2098, opened 2021-10-06
MEN-5107 and MEN-5138: fix and improvements in shell snippets

lluiscampos:
Fixes MEN-5107.

Previous code basically assumed Enteprise=production and OS=demo. This
commit reworks getDebConfigurationCode to use isDemoMode independently
of isEnterprise, so that the four combinations are now possible.

The unit tests should be self-explanatory of all the use cases
considered.

Changelog: Restore support for Open Source Production setups in bash
snippets used during on-boarding and on "Connect a Device" dialog.


lluiscampos:
First part resubmitted at #2292. The second part will be submitted once the client side dependencies make it into production.

PR #2101, opened 2021-10-08
MEN-4411: Show dates formatted according to user locale

oleorhagen:
This changes the date-format string to rely on the locale, as defined by the
browser's locale settings, and can further be modified through a menu in 'User
Management', to override the standard.

The option is saved for the next session, as a standard user setting.

Changelog: Enable the user to pick the time locale, through the 'User
Management' interface. The default is to rely on the browser settings, but is
user modifiable through the drop-down menu.

Signed-off-by: Ole Petter <ole.orhagen@northern.tech>

### DIFF
--- a/src/js/components/common/localetime.js
+++ b/src/js/components/common/localetime.js
@@ -1,0 +1,8 @@
+import React from 'react';
+import Time from 'react-time';
+import LocaleFormatString from './timeformat';
+
+export function LocaleTime(props) {
+  return <Time {...props} format={LocaleFormatString()} />;
+}
+export default LocaleTime;
--- a/src/js/components/common/timeformat.js
+++ b/src/js/components/common/timeformat.js
@@ -1,11 +1,16 @@
 import moment from 'moment';
+// import { getUserSettings } from '../../selectors';
 
-export function LocaleFormatString(stripTime = false) {
-  console.log(moment.locale());
+export function LocaleFormatString(stripTime) {
+  // Get the saved user locale
+  console.log('Current user locale: ' + moment.locale());
+  // const { locale = 'automatic' } = getUserSettings();
+  // console.log('Stored locale: ' + locale);
+  var curLocale = moment.localeData('de');
   if (stripTime) {
-    return moment().creationData().locale._longDateFormat.L;
+    return curLocale.longDateFormat('L');
   }
-  return moment().creationData().locale._longDateFormat.L + ' ' + moment().creationData().locale._longDateFormat.LT;
+  return curLocale.longDateFormat('L') + ' ' + curLocale.longDateFormat('LT');
 }
 
 export default LocaleFormatString;
--- a/src/js/components/common/timeformat.js
+++ b/src/js/components/common/timeformat.js
@@ -1,0 +1,11 @@
+import moment from 'moment';
+
+export function LocaleFormatString(stripTime = false) {
+  console.log(moment.locale());
+  if (stripTime) {
+    return moment().creationData().locale._longDateFormat.L;
+  }
+  return moment().creationData().locale._longDateFormat.L + ' ' + moment().creationData().locale._longDateFormat.LT;
+}
+
+export default LocaleFormatString;
--- a/src/js/components/deployments/deployment-report/deploymentdevicelistitem.js
+++ b/src/js/components/deployments/deployment-report/deploymentdevicelistitem.js
@@ -6,7 +6,7 @@ import { Button, LinearProgress, TableCell, TableRow } from '@material-ui/core';
 
 import { formatTime } from '../../../helpers';
 import DeviceIdentityDisplay from '../../common/deviceidentity';
-import { LocaleTime } from '../../common/localetime';
+import LocaleTime from '../../common/localetime';
 
 const stateTitleMap = {
   noartifact: 'No compatible artifact found',
--- a/src/js/components/deployments/deployment-report/deploymentdevicelistitem.js
+++ b/src/js/components/deployments/deployment-report/deploymentdevicelistitem.js
@@ -7,6 +7,7 @@ import { Button, LinearProgress, TableCell, TableRow } from '@material-ui/core';
 
 import { formatTime } from '../../../helpers';
 import DeviceIdentityDisplay from '../../common/deviceidentity';
+import LocaleFormatString from '../../common/timeformat';
 
 const stateTitleMap = {
   noartifact: 'No compatible artifact found',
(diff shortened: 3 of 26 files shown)


src/js/components/user-management/selfusermanagement.js
```diff
@@ -136,6 +147,24 @@ export const SelfUserManagement = ({ canHave2FA, currentUser, editUser, hasTrack
...
+              saveUserSettings({ 'locale': event.target.value });
+              console.log('Setting the user locale to: ' + event.target.value);
+              moment.locale(event.target.value);
+            }}
+            value={locale}
+          >
+            <MenuItem value={window.navigator.userLanguage || window.navigator.language}>Automatic</MenuItem>
+            <MenuItem value="en">Retard mode</MenuItem>
+            <MenuItem value="de">Deutsche telekom</MenuItem>
+            <MenuItem value="es">Spanglish</MenuItem>
```
oleorhagen, line 164: I'm guessing these will have to be changed, question  is just, how many to list... :thinking:

src/js/components/user-management/selfusermanagement.js
```diff
@@ -136,6 +147,24 @@ export const SelfUserManagement = ({ canHave2FA, currentUser, editUser, hasTrack
...
+          <Select
+            onChange={event => {
+              saveUserSettings({ 'locale': event.target.value });
+              console.log('Setting the user locale to: ' + event.target.value);
+              moment.locale(event.target.value);
+            }}
+            value={locale}
+          >
+            <MenuItem value={window.navigator.userLanguage || window.navigator.language}>Automatic</MenuItem>
+            <MenuItem value="en">Retard mode</MenuItem>
```
oleorhagen, line 162: This one we can keep

src/js/components/common/timeformat.js
```diff
@@ -0,0 +1,11 @@
+import moment from 'moment';
```
mzedel, line 1: The "react"/ idiomatic way would be to not rely on setting the locale internally in `moment`, but in the redux store instead and transform this into a "connected component" instead of a wrapper call to `moment` privates 😕. This way you would not only remove the reliance of some hidden global set somewhere in moment (which might be changed by a datepicker e.g.), but you would also gain the re-initialization of the user-defined locale on page load for free.
 (An additional benefit would be to later on getting rid of the `react-time` package, as this is unmaintained and a blocker for a pending react upgrade...)

BTW: the [proper approach to the formatting using moment](https://momentjs.com/docs/#/i18n/locale-data/) you must have surely missed during your intense study of the docs 😛.

src/js/components/user-management/selfusermanagement.js
mzedel: in any case it should be a `.map` on an array of options "two or more, use a for" 😉, not sure after how many the `duplicate-code` lint rule might kick you otherwise.

src/js/components/user-management/selfusermanagement.js
```diff
@@ -136,6 +147,24 @@ export const SelfUserManagement = ({ canHave2FA, currentUser, editUser, hasTrack
           </p>
         </div>
       )}
+      <div className="flexbox space-between" style={{ alignItems: 'flex-start' }}>
+        <h2>Locale settings</h2>
+        <FormControl>
+          <Select
```
mzedel, line 153: The h2 here doesn't fit in with the rest of the settings I think, maybe something like the _Help us improve Mender_ setting fits in a bit better?

src/js/components/user-management/selfusermanagement.js
oleorhagen: haha, fair enough. But it will be a drop-down menu of a 100+ items, is that okay?

mzedel: You can also have an array of 3 elements... `automatic, en, de, es` - or is there a requirement to cover all the formats?

oleorhagen: Nope :)

src/js/components/deployments/deployment-report/deploymentdevicelistitem.js
```diff
@@ -1,12 +1,12 @@
 import React from 'react';
 import { Link } from 'react-router-dom';
-import Time from 'react-time';
 
 // material ui
 import { Button, LinearProgress, TableCell, TableRow } from '@material-ui/core';
 
 import { formatTime } from '../../../helpers';
 import DeviceIdentityDisplay from '../../common/deviceidentity';
+import { LocaleTime } from '../../common/localetime';
```
mzedel, line 9: Some of these are "named imports", some import the default (aka. no curly braces) - please pick one for consistency. As to which: if there is just a single component/ function/ object/ ... in a module, I usually stick to importing the default; if there are multiple things of interest "it depends"... but in general, I think consistency helps with navigation 👌

src/js/components/common/localetime.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import Time from 'react-time';
+import LocaleFormatString from './timeformat';
+
+export function LocaleTime(props) {
+  return <Time {...props} format={LocaleFormatString()} />;
+}
+export default LocaleTime;
```
mzedel, line 8: you could omit the reliance on the `timeformat` approach entirely if you took the selected format string straight from the redux store:

```suggestion
import { connect } from 'react-redux';
import { getUserSettings } from '../somewhere/are/the/selectors';

export const LocaleTime = ({ locale, stripTime, ...remainingProps }) => {
  const formatString = stripTime ? locale.dateFormat : `${locale.dateFormat} ${locale.timeFormat}`;
  return <Time {...remainingProps} format={formatString} />;
}

const mapStateToProps = state => {
  const { locale } = getUserSettings(state);
  return {
    locale
  };
};
export default connect(mapStateToProps)(LocaleTime);
```

you could of course also retrieve the `moment` localeData here and do the assembly based on this - but this way the component call could happen with the "proper" format already... I think it is rather unlikely that the constant `localeData` will change between selection and rendering, but the way you changed the call to `moment` now I'm fine with both approaches 👍

tests/setupTests.js
```diff
@@ -50,6 +50,7 @@ beforeAll(async () => {
     hostname: TEST_LOCATION,
     replace: jest.fn()
   };
+  window.navigator.userLanguage = 'en';
```
mzedel, line 53: that most likely won't work, as ?most of? the navigator attributes are immutable... but you should be able to overwrite them just like the `window.location` and `.sessionStorage` right above 👍 (and that should also make the snapshots pass again)

PR #2132, opened 2021-10-13
MEN-4989

mzedel:


michaelatmender (commented):
Manually testing this throughout the UI seems to work very well in most cases!
I ran into an issue with open pending deployments - refreshing seems to lose the deployment details:
image

michaelatmender:
> Thanks a lot for looking at this @michaelatmender! The refresh with a deployment open should now work with [this change](https://github.com/mendersoftware/gui/compare/c3b590c7c516cb8682eafcf7a8bd2a97ed704530..bb84b6eefc8b1929bab480f283c5fefc21e8f224#diff-67a9a43d2d7584861d1ff0447e47f95ff779b10965873b13ebc09ee9d6acff0bL98-R104) in...

Nice, looks good now!
One more small thing - when refreshing Finished Deployments, if either '7 days' or '30 days' filters are selected they don't persist after refresh (although 'yesterday' works perfectly!)

tranchitella (approved):
It works fine here.

PR #2163, opened 2021-10-22
MEN-5170 - the UI part

mzedel:
This contains primarily a switch to a virtualized release list, filterable by `name`, `device_type`, `description` and sorted by modification date by default - as well as an asynchronous release retrieval during deployment creation...

Since the code in `releaseActions.js` is not very pretty, a short note on what should happen:
- when searching for a release or artifact, the release retrieval will try getting releases filtered by name, if none exist it will try getting them filtered by device_type, followed by filtered by description - so at most 3 requests per search...
- sections for the release list that are not yet cached locally will be retrieved and shown as loading
- once a user has scrolled in the releases list, only the visible area + a few items before and after would be updated - as updating large parts of that list could otherwise soon become impactful for the performance... - in order to do the update, the window size will determine the number of releases retrieved
  - unfortunately this might cause a misaligned release list, if new releases were uploaded while looking at the release list and the uploaded releases land at a lower index in the overall release list... I tried to factor such releases in, but it would again require full knowledge of the release list (again too much data) to the start index of the retrieved window to get it right...
  - updates to a filtered list will only query the attribute that lead to results on the initial filtering (so when you searched for a `description`, there should only be queries for releases filtered by description (instead of for name + device_type first))
- the releases list during deployment creation will be filtered by name on entry... (that is what the `searchOnly` path in the release action is for)

I worked on it locally with ~5000 releases and the difference responsiveness is significant... a downside: the tests now take slightly longer and unfortunately checking the population of the release list in the unit tests is not working, due to the virtualization library but 🤞 the e2e tests work with it...


tranchitella:
@mzedel two issues I detected:

Uploading an artifact causes this error in a tooltip (despite being successfully uploaded & processed by backend):
```
Artifact couldn't be generated. Invariant failed: A state mutation was detected between dispatches, in the path 'app.cancelSource.token._listeners.0'.  This may cause incorrect behavior. (https://redux.js.org/style-guide...
```

The second issue is a double request to the backend, look at this screenshot:

image

I typed raspberrypi3 (copy-and-paste into the text field). The first call is okay, it filters by name and cannot find anything. It then triggers the second call filtering by device type, which returns the list. I don't understand why the UI is making yet another call (the third one), with the same parameters and the same response from backend.

michaelatmender (commented):
I also experienced the uploading Artifact error message and double calls.
Besides that, the actual UX of the search & the gradual loading of Artifacts using pagination worked nicely.

mzedel:
Thanks @michaelatmender & @tranchitella for looking into this!
The first issue seems to be pre-existing - the error handler is misleading there: 
> ```
> Artifact couldn't be generated. Invariant failed: A state mutation was detected between dispatches, in the path 'app.cancelSource.token._listeners.0'.  This may cause incorrect behavior. (https://redux.js.org/style-guide...
> ```
The `cancelSource.token` is needed to allow cancelling the artifact upload and is mutated outside of the redux flow during the upload (so far I'm not aware of a cleaner way to handle this)... The `Invariant failed: ...` error is caused by one of the dev dependencies, that checks for immutability on changes in the redux state (as a best practice). And since the error handler is implemented for the entire artifact generation, it catches the error message and thinks the generation failed...
These mutation checks are only run when using the dev/ watch command - I'll try it again with a production build, but I'm 
quite sure this won't be an issue there...

The second issue should be addressed soon though 👌!

mzedel:
@michaelatmender & @tranchitella the additional request should now be fixed 👌

tranchitella (approved):

PR #2211, opened 2021-11-01
MEN-5145 - device provider integration

mzedel:
there are still some changes needed elsewhere to get the per device configuration string and provider removal, but the overall UI part should be ready


src/js/components/settings/integrations.js
```diff
@@ -0,0 +1,142 @@
...
+  const onCancelClick = ({ connectionString }) => {
+    setIsCreating(!!connectionString);
+    setCurrentValue('');
+    setConfiguredIntegrations(integrations);
+  };
+
+  const onSaveClick = integration => {
+    changeIntegration(integration);
+    setCurrentValue('');
+    return setIsCreating(isCreating && !integration.connectionString ? false : isCreating);
```
tranchitella, line 105: This is not working here. I use, as a test value, `HostName=test.azure-devices.net;SharedAccessKeyName=iothubowner;SharedAccessKey=ZHVtbXkgLW4K`.
This is what I get when I try to save the connection string:

image

The connection string is correctly saved in the backend/database:

```
> db.settings.find().pretty()
{
	"_id" : ObjectId("6195dacc6c2c018095343290"),
	"connection_string" : {
		"access_key" : BinData(0,"ZHVtbXkgLW4K"),
		"hostname" : "test.azure-devices.net",
		"name" : "iothubowner"
	},
	"tenant_id" : ""
}
```

Despite it, though, refreshing still shows the "Add new integration" dropdown:

image

src/js/components/settings/settings.js
```diff
@@ -11,13 +11,53 @@ import { Elements } from '@stripe/react-stripe-js';
...
+  },
+  'role-management': { admin: true, enterprise: true, multitenancy: false, userManagement: false, component: <Roles />, text: () => 'Roles' },
+  'integrations': {
+    admin: true,
+    enterprise: false,
+    multitenancy: false,
+    userManagement: false,
+    component: <Integrations />,
+    text: () => 'Integrations'
+  },
```
tranchitella, line 49: Don't we need to add a menu item "Integrations" in the user dropdown, too?

image

src/js/constants/deviceConstants.js
```diff
@@ -34,6 +36,12 @@ module.exports = {
 
   SET_DEVICE_LIMIT: 'SET_DEVICE_LIMIT',
 
+  EXTERNAL_PROVIDER: {
+    amazon: { article: 'an', icon: AWS, title: 'AWS IoT core', key: 'amazon', managementUrl: '/v1/aws-iot-manager/settings', enabled: false },
+    azure: { article: 'an', icon: Azure, title: 'Azure IoT Hub', key: 'azure', managementUrl: '/v1/azure-iot-manager/settings', enabled: true },
+    google: { article: 'a', icon: GCP, title: 'Cloud IoT Core', key: 'google', managementUrl: '/v1/gcp-iot-manager/settings', enabled: false }
```
tranchitella, line 42: You are my hero

tranchitella (changes requested):

src/js/components/settings/integrations.js
mzedel: Thanks for that connection string - I tried some others I found, but none would validate - but using your format it should work now. However deleting the integration is not possible, as the validation rejects empty strings too 😕

tranchitella (approved):

PR #2252, opened 2021-11-17
MEN-5203: tests: Do not inject server certificate(s)

lluiscampos:
For `mender-connect` the setting is deprecated. The Docker image
mender-client-docker-addons has now the demo certificate installed in
the root, so it should be necessary.

Remove also the code for getting/injecting the staging.mender.io cert as
the server is CA-backed and shouldn't be necessary.


mzedel (approved):
Thank you very much!

lluiscampos:
The tests passed after merging https://github.com/mendersoftware/integration/pull/1753 :rainbow:

PR #2274, opened 2021-11-30
Fix/integrations+e2e tests

mzedel:


src/js/components/settings/integrations.js
```diff
@@ -76,14 +76,15 @@ export const Integrations = ({ integrations, changeIntegration, deleteIntegratio
...
-        return accu;
-      }, [])
-    );
+    const available = Object.values(EXTERNAL_PROVIDER).reduce((accu, provider) => {
+      if (provider.enabled && !integrations.some(integration => integration.provider == provider.provider)) {
+        accu.push(provider);
+      }
+      return accu;
+    }, []);
+    if (!available.every(integration => availableIntegrations.some(availableIntegration => availableIntegration.provider === integration.provider))) {
```
merlin-northern, line 85: just for my peace of mind, it is not possible for `available` to be null here, right?

merlin-northern (approved):
lgtm

src/js/components/settings/integrations.js
mzedel: nope, the reduce above gets an array (the last arg in the line right above (L84) and the `return accu` is always run. Even if the `EXTERNAL_PROVIDER` constant ever turns out to be an empty object, this would lead to the empty array being assigned. The only chance would be if the `EXTERNAL_PROVIDER` would be unset/ removed, but that would be either caught by the lint checks or we would have bigger problems 🙈.

merlin-northern: thank you for the explanation.

PR #2292, opened 2021-12-06
MEN-5107: Rework getDebConfigurationCode to support OS/Enterprise w/ demo/production

lluiscampos:
Fixes MEN-5107.
    
    Previous code basically assumed Enteprise=production and OS=demo. This
    commit reworks getDebConfigurationCode to use isDemoMode independently
    of isEnterprise, so that the four combinations are now possible.
    
    The unit tests should be self-explanatory of all the use cases
    considered.
    
    Changelog: Restore support for Open Source Production setups in bash
    snippets used during on-boarding and on "Connect a Device" dialog.


lluiscampos:
@mzedel,

This is a re-submit for the first part of #2098.

The second commit is just formatting that my editor did on his/her own, I can drop it if you don't want it.

mzedel (approved):
The formatting changes prevent the lint checks from passing I think... you can likely drop the last commit and run `npm run lint-fix`...
 unfortunately some of the `prettier` extensions haven't caught up with the formatting changes in the recent prettier releases (which I would guess caused the changes)...

And I do look forward to when the client side changes land, to get rid of that extra `--demo` 😄

lluiscampos:
> The formatting changes prevent the lint checks from passing I think... you can likely drop the last commit and run `npm run lint-fix`... unfortunately some of the `prettier` extensions haven't caught up with the formatting changes in the recent prettier releases (which I would guess caused the changes)...

Thanks. Dropped the second commit.

> 
> And I do look forward to when the client side changes land, to get rid of that extra `--demo` smile

Me too :grinning:

lluiscampos:
@mender-test-bot :cherries: 
cherry-pick to:
* 3.1.x

PR #2295, opened 2021-12-07
MEN-5147 - azure device twin in device details

mzedel:
Changelog: None


mzedel:
to test you need to at least configure the azure integration in the settings, but preferably you need to have a device twin in azure

src/js/actions/deviceActions.js
```diff
@@ -908,3 +933,47 @@ export const setDeviceTags = (deviceId, tags) => dispatch =>
       .catch(err => commonErrorHandler(err, `There was an error setting tags for device ${deviceId}.`, dispatch, 'Please check your connection.'))
       .then(() => Promise.resolve(dispatch({ type: DeviceConstants.RECEIVE_DEVICE, device: { ...device, tags } })));
   });
+
+export const getDeviceTwin = (deviceId, integrationProvider) => (dispatch, getState) =>
+  GeneralApi.get(`${apiBase}${DeviceConstants.EXTERNAL_PROVIDER[integrationProvider].managementUrl}/device/${deviceId}/twin`)
```
alfrunes, line 938: This was my fault, the path name should be `/devices/{id}/twin` the docs were wrong.
```suggestion
  GeneralApi.get(`${apiBase}${DeviceConstants.EXTERNAL_PROVIDER[integrationProvider].managementUrl}/devices/${deviceId}/twin`)
```

src/js/actions/deviceActions.js
```diff
@@ -908,3 +933,47 @@ export const setDeviceTags = (deviceId, tags) => dispatch =>
...
+                updated_ts: data.lastActivityTime
+              }
+            }
+          }
+        })
+      )
+    );
+
+export const setDeviceTwin = (deviceId, integrationProvider, settings) => (dispatch, getState) =>
+  GeneralApi.put(`${apiBase}${DeviceConstants.EXTERNAL_PROVIDER[integrationProvider].managementUrl}/device/${deviceId}/twin`, settings)
```
alfrunes, line 959:
```suggestion
  GeneralApi.put(`${apiBase}${DeviceConstants.EXTERNAL_PROVIDER[integrationProvider].managementUrl}/devices/${deviceId}/twin`, settings)
```

src/js/components/devices/device-details/devicetwin.js
```diff
@@ -0,0 +1,215 @@
...
+  };
+
+  const handleDiffEditorDidMount = (editor, monaco) => {
+    const modifiedEditor = editor.getModifiedEditor();
+    modifiedEditor.onDidChangeModelContent(() => setUpdated(modifiedEditor.getValue()));
+    editor.onDidUpdateDiff(onDidUpdateDiff);
+    editorRef.current = { editor, monaco, modifiedEditor };
+  };
+
+  const onDidUpdateDiff = () => {
```
alfrunes, line 120: The diff compares all JSON properties including the metadata such as when the properties that holds the individual update timestamp. Can we exclude properties starting with `$` from the diff?

src/js/components/devices/device-details/devicetwin.js
```diff
@@ -0,0 +1,215 @@
...
+    const changes = editorRef.current.editor.getLineChanges();
+    console.log(editorRef.current.modifiedEditor);
```
alfrunes, line 122: ... also, is this log for debugging? :)

alfrunes (changes requested):
The UI looks slick! :100: 
However, I made a mistake in API docs which I see you caught here. While you're at it, can you do some filtering magic on the diff view?

src/js/components/devices/device-details/devicetwin.js
mzedel: whoops 🙈 - should now be gone!

src/js/actions/deviceActions.js
mzedel: should now point to the right endpoint 👌

src/js/actions/deviceActions.js
mzedel: should now point to the right endpoint 👌

src/js/components/devices/device-details/devicetwin.js
mzedel: Since the entire object can be user defined I don't think excluding them all is a good idea, we might easily remove some user defined ones other wise... we could exclude the azure defined ones explicitly, but that still might cause trouble once a user defined one gets overwritten.

mzedel:
Thanks for looking into this @alfrunes, @michaelatmender and @tranchitella ! If you have the time I would appreciate another round of testing, as I adjusted the api endpoint. 
There is however one issue I came across: the API responds with a `401` if the defined connection string for the azure integration can't access the device twin.. On the Azure <=> Mender API level that is understandable, however it also causes a log out for the UI, which I find a bit excessive... maybe returning a `403` or `409` would be less problematic?

src/js/components/devices/device-details/devicetwin.js
alfrunes: As far as I can tell it's only the `$version` and `$metadata` properties in the root object that's set by the IoT Hub backend (although it looks like it's not always correctly set). However, these are not documented anywhere so I don't know what the correct approach is here :man_shrugging:
In any case, it will always show at least two lines differ because these properties will (almost) always mismatch.

alfrunes:
@mzedel there's only one small issue left. The schema for updating the device twin is wrong, it should be:
```json
{"properties": {<desired properties>}, "tags": {<device tags>}}
```

src/js/components/settings/__snapshots__/integrations.test.js.snap
```diff
@@ -35,15 +35,16 @@ exports[`Integrations Component renders correctly 1`] = `
...
       </label>
       <div
-        class="MuiInputBase-root MuiInput-root MuiInput-underline Mui-disabled Mui-disabled MuiInputBase-formControl MuiInput-formControl"
+        class="MuiInputBase-root MuiInput-root MuiInput-underline Mui-disabled Mui-disabled MuiInputBase-formControl MuiInput-formControl MuiInputBase-multiline MuiInput-multiline"
       >
-        <input
+        <textarea
           aria-invalid="false"
-          class="MuiInputBase-input MuiInput-input Mui-disabled Mui-disabled"
+          class="MuiInputBase-input MuiInput-input Mui-disabled Mui-disabled MuiInputBase-inputMultiline MuiInput-inputMultiline"
```
michaelatmender, line 42: You can also use the "multiline" property, which I think looks a little nicer https://mui.com/components/text-fields/#multiline

src/js/components/settings/integrations.js
```diff
@@ -25,17 +27,20 @@ export const IntegrationConfiguration = ({ integration, onCancel, onDelete, onSa
...
+      <h3 className="margin-bottom-none">{title}</h3>
       <div className="flexbox space-between relative" style={{ alignItems: 'flex-end', maxWidth }}>
         <TextField
           disabled={!isEditing}
           InputLabelProps={{ shrink: !!connectionConfig }}
-          label={`${EXTERNAL_PROVIDER[provider].title} connection string`}
+          label={`${title} connection string`}
+          multiline
           onChange={updateConnectionConfig}
+          rows={isEditing ? 1 : 2}
```
michaelatmender, line 41: Leaving "rows" blank looks a little nicer imo, as the height dynamically changes and there are no scrollbars

mzedel: good point! should now be adjusted 👌

michaelatmender (changes requested):
Works nicely! I encountered a couple of areas for polishing a bit, which were missed in the initial design docs. [I listed them here](https://tracker.mender.io/browse/MEN-5000?focusedCommentId=121775&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-121775) 
The main issue was when there's an error finding the device twin, it becomes a bit obtrusive with the constant error snackbar messages. The others are smaller QoL adjustments

src/js/components/devices/device-details/devicetwin.js
```diff
@@ -0,0 +1,250 @@
...
+    setIsEditing(false);
+  };
+
+  const onEditClick = () => setIsEditing(true);
+
+  return (
+    <DeviceDataCollapse
+      header={
+        !open &&
+        (twinError ? (
```
michaelatmender, line 176: This is throwing a "not defined" error now

mzedel: argh right... right now I'm adding the refresh button and making the external integration setup depend on the released version - once I've pushed this I'll ping you 🙂

mzedel: @michaelatmender now the refresh button should also work 👌
A note on testing: From now on you have to set `INTEGRATION_VERSION: 3.2.0` for the integration settings to show. I thought this might be the easiest way to decide if we can show it, to not require an additional flag when running the current GUI on Hosted Mender...

src/js/constants/deviceConstants.js
```diff
@@ -37,9 +39,47 @@ module.exports = {
...
+      managementUrl: '/v1/aws-iot-manager',
+      enabled: false,
+      configHint: <>For help finding your AWS IoT core connection string, check the AWS IoT documentation.</>
+    },
+    azure: {
+      article: 'an',
+      icon: Azure,
+      title: 'Azure IoT Hub',
+      provider: 'azure',
+      managementUrl: '/v1/azure-iot-manager',
```
michaelatmender, line 56: This must be renamed to "iot-manager" after latest changes to mendersoftware/integration

michaelatmender (changes requested):
Had some issues around expanding:

device twin
- the unexpanded 'summary' section is not up to date until you click to expand it
- the timestamp of the summary is not matching any correct time
- the summary disappears when "show less" is clicked

tranchitella:
@mzedel I rebased on master and adapted the UI to the latest version of the `iot-manager`APIs. I did not change tests, the only purpose of my commit (3fb12d3ef24acc234e1d1156a0206bb114b4b489) is to facilitate your work.

mzedel:
Thanks again @tranchitella for the adjustments! Based on this and the changes in the iot-manager I hope this gets as close to working, but I guess we'll have to wait until the [remaining iot-manager PR](https://github.com/mendersoftware/iot-manager/pull/14) is merged...

tranchitella:
@mzedel one additional comment from me: when testing it, I noticed you implemented a logic similar to what we apply in configuration add-on: you show the "reported" twin, a message if the twin is in sync or not, and then the editor for the "desired" state. Please be aware that in Azure IoT Hub world, reported and desired are completely independent objects, and the assumption that the device will keep them in sync, reporting the "desired" state back to the cloud, is wrong.

I don't know if this affects UX as well, but I believe it is important to highlight it.

mzedel:
Hmm.. good point. In my opinion highlighting the differences should be ok in any case, as we don't show it as an error, but just as a regular difference... but pinging @michaelatmender to be sure.

michaelatmender:
> Hmm.. good point. In my opinion highlighting the differences should be ok in any case, as we don't show it as an error, but just as a regular difference... but pinging @michaelatmender to be sure.

It seems I initially thought there was more of a correlation between desired and reported properties. But I agree that highlighting a difference is still useful, when there is a difference in values for the _same key_.

mzedel:
just to be sure, do you @michaelatmender mean with
>  when there is a difference in values for the _same key_.

that the diff should be modified to ignore all other changes? IMO it would be ok as is, and leave the assessment of the difference to the users - they are editing json after all (why I would assume they are familiar with a diff view).

As for the device connection tooltip that would show alongside the device type selection + clicking the link will open a new tab to allow returning to the same spot:
Screenshot 2022-01-05 at 13 35 07
Screenshot 2022-01-05 at 13 34 26

And during onboarding this would be shown alongside the tooltip for the device type selection as follow:
Screenshot 2022-01-05 at 13 36 15
Screenshot 2022-01-05 at 13 36 07

also maybe of interest for @eysteins

estenberg:
@mzedel @michaelatmender this looks like a fine starting point on both counts (diff view and onboarding). I agree there is some uncertainty how this will be used here, but I think any issues can be fixed later with relatively low effort. I would on the other hand not spend a lot of time discussing / designing this further right now. Just keep it simple (and low effort) for now. :)

A few specific notes / nitpicks:
* On the tooltip (i icon) about Azure IoT: It's great that it's context sensitive (knows if you did it or not). I would just keep it on the same place: don't move it left when onboarding is not active and right when it is active.
* It feels like the i icon should appear on the screen *before* this, because then it would cover virtual devices too and it wouldn't seem to be tied to the device type or adding a specific device. But again, I don't think it's critical for now, we can revise later.
* On the diff view, although I do agree with @tranchitella that desired/reported are technically separate object and may never correlate, I have not seen this to be the way people want to use it. When you set the desired property, you do want the device to report the property to be set. It's mainly a sync mechanism after all -- you don't want it to be out of sync forever. I also tested the UX with a few customers and they thought it was fine. Keeping it as-is (it just shows a diff) I think will work and be appreciated.
* For the future, I think what will be more relevant is turning this into nice forms (e.g. drop-down of values for Timezone, etc.), not optimizing how/if diffs are shown. But it's a larger discussion. :)

estenberg:
Also FYI it would be really good to have this in staging this week; we're having some demos/recordings that require it next week.. to me it seems pretty finished so hope it's ok but let me know if not :)

mzedel:
Thanks for the input @estenberg - the icon location is now adjusted and stays in the same place, also after onboarding:
Screenshot 2022-01-05 at 20 02 53

And rest assured, this will land in staging as soon as it's merged 😄

estenberg:
:+1: :+1:

tranchitella:
@mzedel is this PR ready for review?

mzedel:
Yes I think so!

tranchitella (approved):
LGTM

michaelatmender (approved):
All good 👍

PR #2299, opened 2021-12-09
MEN-5232

mzedel:


michaelatmender (changes requested):
Nince, just a couple of small adjustments requested to the integration section in the dialog:
onboarding-azure

mzedel:
@michaelatmender there should have already been a regular sized margin between the two 😕 - I increased it further and hope that works now 👍... as for the connection string that should now only be enabled if there already is a connection string set for the integration (and this will be filled in) - however, the population of the connection string details is part of the [device twin PR](https://github.com/mendersoftware/gui/pull/2295/files#diff-7cebf71fa7d168419328f3fbfd3b8c7f89691ecadce149c777813ba188491ac7R49-R54)...

michaelatmender (approved):

PR #2317, opened 2021-12-16
MEN-5331 - fixed an issue that limited profile edits to passwords

mzedel:
now editing the own email address is possible again

- also ensured the changed email is reflected in the UI right after editing
Changelog: title

### DIFF
--- a/src/js/actions/userActions.js
+++ b/src/js/actions/userActions.js
@@ -162,9 +162,14 @@ export const removeUser = userId => dispatch =>
     )
     .catch(err => userActionErrorHandler(err, 'remove', dispatch));
 
-export const editUser = (userId, userData) => dispatch =>
+export const editUser = (userId, userData) => (dispatch, getState) =>
   GeneralApi.put(`${useradmApiUrl}/users/${userId}`, userData)
-    .then(() => Promise.all([dispatch({ type: UserConstants.UPDATED_USER, userId, user: userData }), dispatch(setSnackbar(actions.edit.successMessage))]))
+    .then(() =>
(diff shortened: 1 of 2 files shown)


src/js/actions/userActions.js
```diff
@@ -162,9 +162,14 @@ export const removeUser = userId => dispatch =>
     )
     .catch(err => userActionErrorHandler(err, 'remove', dispatch));
 
-export const editUser = (userId, userData) => dispatch =>
+export const editUser = (userId, userData) => (dispatch, getState) =>
   GeneralApi.put(`${useradmApiUrl}/users/${userId}`, userData)
-    .then(() => Promise.all([dispatch({ type: UserConstants.UPDATED_USER, userId, user: userData }), dispatch(setSnackbar(actions.edit.successMessage))]))
+    .then(() =>
+      Promise.all([
+        dispatch({ type: UserConstants.UPDATED_USER, userId: userId === UserConstants.OWN_USER_ID ? getState().users.currentUser : userId, user: userData }),
```
tranchitella, line 169: I'm not sure I understand: why are you changing this API call?

mzedel: This has to be done so the email change is visible in the UI without a refresh. The user id that is used to communicate with the backend in this case is `me`, but to track the state in the UI I have to switch to the full uuid of the current user... (when editing other users the uuid is used and this is not a problem)

tranchitella (approved):

mzedel:
@mender-test-bot 

cherry-pick to:
 * 3.1.x

PR #2370, opened 2022-01-12
MEN-5250: Remove quiet flag from on-boarding bash snippet

lluiscampos:
To get an explicit error on connectivity issues if the device cannot
reach the Internet.


lluiscampos:
Still pending on decision from [JIRA discussion](https://tracker.mender.io/browse/MEN-5250?focusedCommentId=122125&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-122125).

@mzedel do you think this deserves a changelog entry?

mzedel (approved):
doesn't deserve a changelog flag, since this is not really part of the frontend functionality IMO

lluiscampos:
I managed to commit only the test changes :sweat_smile:

mzedel (approved):

PR #2376, opened 2022-01-14
improved monitor config absence notification granularity

mzedel:
- this primarily takes existing alerts into account
- unfortunately there is still a chance for the notification to be misleading - especially in case of offline devices

Changelog: None


merlin-northern (approved):
lgtm, I understand that the !isOffline check is on purpose, since we would be unable to show anything.

merlin-northern:
@oleorhagen if you have not started still, this is something we could take into the release and also briefly test.

oleorhagen:
Release is branched out now, so this will have to be cherry-picked if so

mzedel:
that is in already 👌

PR #2426, opened 2022-01-31
fix: disable the update control map control if not enterprise (MEN-5424)

tranchitella:
Changelog: title


merlin-northern (approved):
lgtm!

PR #2466, opened 2022-02-17
Small adjustments to theme styles

michaelatmender:
Changelog: None

### DIFF
--- a/src/js/themes/Mender/light.js
+++ b/src/js/themes/Mender/light.js
@@ -20,7 +20,8 @@ const variantPalette = {
     '50': '#d4e9e7'
   },
   background: {
-    default: '#fff'
+    default: '#fff',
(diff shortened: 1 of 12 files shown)


src/js/themes/Mender/light.js
```diff
@@ -78,6 +78,13 @@ export const light = {
           color: variantPalette.text.primary
         }
       }
+    },
+    MuiTooltip: {
+      styleOverrides: {
+        tooltip: {
+          backgroundColor: '#24444a'
```
mzedel, line 85: that could probably be defined in the `variantPalette` above to keep the color definitions in one place, OTOH it is a theme file already

mzedel (approved):
thank you very much! one suggestion, but also feel free to merge right away!

mzedel (approved):
excellent 👌!

PR #2468, opened 2022-02-17
MEN-5233 - customizable columns 

mzedel:
~e2e test failures are due to an [issue in playwright](https://github.com/microsoft/playwright/issues/11202) that coincides with the introduction of the changed device list structure. I'm not sure if/ when this will be picked in a coming patch release for 1.19 or if we have to wait for a few weeks for 1.20 to be released... OTOH I wouldn't want to introduce workarounds for this, as the tests in question are ok as they are + their passing/ failures would be very informative...~ - fixed & adjusted test approach for improved robustness

~besides this the current test image also needs [this issue](https://github.com/microsoft/playwright/issues/12106) to be fixed~ fixed!

while waiting for [this PR](https://github.com/mui/material-ui/pull/31313) to be merged, disabling the autocomplete will live with a workaround


michaelatmender (approved):
This looks great!

PR #2470, opened 2022-02-18
fix e2e test

mzedel:
Changelog: None


mzedel:
only show in enterprise runs (so on master) - successful run is [here](https://gitlab.com/Northern.tech/Mender/gui/-/jobs/2109821561)

oleorhagen (approved):

PR #2537, opened 2022-03-16
fix: allowed deployment creation to be retried on error

mzedel:
Changelog: None


merlin-northern (approved):
tested.

PR #2580, opened 2022-03-29
MEN-5562 - prevented an issue that allowed to create a group that could not be removed

mzedel:


tranchitella (approved):
Nice, tested locally and it works fine.
I went through the code and I couldn't spot anything.

PR #2581, opened 2022-03-30
Chore/get bulk statistics

mzedel:
for motivational purposes @oleorhagen

depends on mendersoftware/deployments#666


tranchitella (approved):
Looks good and works fine.

PR #2582, opened 2022-03-31
fix: aligned signup & login page rendering

mzedel:
Changelog: None


tranchitella (approved):
Thank you for the quick fix!

PR #2603, opened 2022-04-07
feat: Update Help topics and content #5580

michaelatmender:
Changelog: Title


src/js/components/help/releases-and-artifacts/build-demo-artifact.js
```diff
@@ -1,104 +0,0 @@
...
-sudo chmod +x mender-artifact && \\
-sudo mv mender-artifact /usr/local/bin`;
-const artifactGenerator = 'single-file-artifact-gen';
-const artifactName = 'demo-webserver-updated';
-const file_modification = `cat >index.html <<EOF
-Hello World!
-EOF
-`;
-
-export const BuildDemoArtifact = ({ demoArtifactLink, menderVersion = 'master', menderArtifactVersion = 'master' }) => {
```
mzedel, line 23: In this particular case there is not a lot of value in keeping this in the UI, for other commands however I could imagine it to be very useful to have a central place to find code snippets that use e.g. the JWT information or the tenant token... But maybe that is becoming more relevant once we have API key support...

src/js/components/common/left-nav.js
```diff
@@ -35,17 +36,25 @@ export const LeftNav = ({ sections }) => {
       {sections.map(({ itemClass = '', items = [], title = '' }, index) => (
         <React.Fragment key={`${itemClass}-${index}`}>
           <ListSubheader disableSticky={true}>{title}</ListSubheader>
-          {items.map(({ exact, path, secondaryAction = null, style = {}, title = '' }) => (
+          {items.map(({ exact, path, secondaryAction = null, style = {}, title = '', url }) => (
```
mzedel, line 39: with the added differentiation between regular `a` and `NavLink` there is an additional error in the browser console, since `a` doesn't understand the boolean `exact`, but `NavLink` also complains about this being just a string - so you could either have `exact={url ? `${exact}` : exact} or:
```suggestion
          {items.map(({ exact, path, secondaryAction = null, style = {}, title = '', url }) => {
            const props = url
              ? { component: 'a', exact: `${exact}`, href: url, rel: 'noopener noreferrer', target: '_blank', to: url }
              : { component: NavLink, exact, to: path };
            return (
              <ListItem className={`navLink ${itemClass} ${classes.list}`} key={path} secondaryAction={secondaryAction} style={style} {...props}>
                <ListItemText className={classes.listItem} primary={title} />
              </ListItem>
            );
          })}
```
- this would keep the `url`-based decisions all in one place...

src/js/components/help/help.js
```diff
@@ -63,24 +46,15 @@ const eachRecursive = (obj, path, level, accu, isHosted, spacing) =>
         path: this_path,
         hosted: value.hosted,
         style: { paddingLeft: `calc(${level} * ${spacing})` },
-        exact: true
+        exact: true,
+        url: value.url ? value.url : ''
```
mzedel, line 50: you could add a  `secondaryAction: value.url ? <LinkIcon /> : null,` here as well, then the extra icon at the end would be handled by the `LeftNav`/`ListItem` 👌  (this is what is used [here in the settings](https://github.com/mendersoftware/gui/blob/master/src/js/components/settings/settings.js#L94))

mzedel (approved):
Works good as it is, the 2 spots I commented are not necessarily required, but would still be nice to have 🙂

src/js/components/help/help.js
```diff
@@ -7,43 +7,26 @@ import { useTheme } from '@mui/material';
...
+  'mender-hub': {
+    title: 'Mender Hub',
+    component: MenderHub
   },
-  'more-help-resources': {
-    title: 'More resources',
-    component: MoreHelp
+  'documentation': {
+    title: 'Documentation',
+    url: 'https://docs.mender.io'
```
mzedel, line 29:
```suggestion
    url: 'https://docs.mender.io/'
```
might be what causes the link checker to fail...

src/js/components/common/left-nav.js
michaelatmender: Thanks, have updated with these adjustments!

src/js/components/help/help.js
mzedel: 🤦 sorry, only now I see it... the docs link checker expect a string surrounded/ ending with the `` ` `` marks instead of the `'`... sorry about that!

PR #2660, opened 2022-04-21
MEN-5563

mzedel:


tranchitella (approved):
It looks good and it works fine. The only (minor) issue I spotted is that scrolling of the side panel containing the device details doesn't work if the cursor is in the right/bottom corner of the screen, hover the `.fixedButtons` div. I don't know if this is easy to fix it.

mzedel:
Excellent find @tranchitella thanks a lot + that was indeed easy to fix!

PR #2677, opened 2022-04-25
Fix/columns reload

mzedel:


tranchitella (commented):
Honestly, I'm not able to review the changes in the code. I skimmed through the commits, I can understand them high-level, but that's it. I tested the PR quite extensively, though, and it works fine but for one thing: I cannot resize the columns. I don't know if this is intended, but when I hover the vertical segment on the right boundary of the column, it highlights but clicking on it I cannot change the size of the column.

Other than that, everything works fine and I cannot see any issue despite window resize events, pagination, filters and sorting on the list.

mzedel:
> I cannot resize the columns. I don't know if this is intended, but when I hover the vertical segment on the right boundary of the column, it highlights but clicking on it I cannot change the size of the column.

Thanks a lot for taking a look @tranchitella!
Resizing _should_ be possible, but to do so you need to drag the highlight/ slider  with some caveats:
- ID/ name + status columns are deliberately not resizable and can't be removed
- it would be an option to show a tooltip on click on the resize handle (if it's not followed by a drag soon)
- there still is a limitation that the initial movement needs to be towards the right/ a size increase, but after that resizing is possible in both directions - that became natural to me by now, but I'll preferably address it in another follow up (as it's pre-existing + doesn't break the page as the issues did that this PR is targeted at)

tranchitella:
@mzedel you are right, now that I know how to resize columns, it works fine.
However, I have another (small) issue: if I refresh the browser when I'm at the URL https://localhost/ui/#/devices/accepted, it doesn't show any device, and the message "No devices found" appears. Changing the status to any shows the devices, as well as selecting back accepted.

The same behaviour happens if I log out, refresh the page at the login form, then log in and go to the list of devices. Again, selecting a different status refreshes the device list.

Does it work fine for you?

mzedel:
@tranchitella for me it worked somewhat fine, with the "No devices found" being visible, but quickly disappearing once the page completes loading (so no state change needed). I adjusted the rendering to wait for the loading to properly finish - does this fix the device list for you too?

tranchitella (approved):
It works fine now, thank you.

PR #2714, opened 2022-05-05
Feat/staging test safari

mzedel:


mzedel:
this should also fix the parsing error gitlab shows for the current staging pipeline...

tranchitella (approved):

PR #2720, opened 2022-05-09
MEN-5611 fix: fixed an issue that would prevent the device list from getting shown on page load

mzedel:


tranchitella (approved):
I also noticed this when testing, just did not have time to report it!
I confirm it works fine now.

PR #2760, opened 2022-05-18
MEN-5620 - fix: ensured deployment release information doesn't overflow list info

mzedel:
Ticket: MEN-5620
Changelog: Title

### DIFF
--- a/src/js/components/deployments/__snapshots__/deploymentitem.test.js.snap
+++ b/src/js/components/deployments/__snapshots__/deploymentitem.test.js.snap
@@ -7,7 +7,10 @@ exports[`DeploymentItem Component renders correctly 1`] = `
   >
     Release
   </span>
-  <div>
+  <div
+    class="text-overflow "
+    title="test"
+  >
     test
   </div>
 </div>
--- a/src/js/components/deployments/deploymentitem.js
+++ b/src/js/components/deployments/deploymentitem.js
@@ -14,10 +14,10 @@ import { getDeploymentState } from '../../helpers';
 import { makeStyles } from 'tss-react/mui';
 
 export const deploymentTypeClasses = {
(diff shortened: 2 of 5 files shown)


src/js/components/deployments/deploymentitem.js
```diff
@@ -49,7 +49,11 @@ export const DeploymentRelease = props => {
     deployment: { artifact_name, type = DEPLOYMENT_TYPES.software }
   } = props;
   const deploymentRelease = type === DEPLOYMENT_TYPES.configuration ? type : artifact_name;
-  return <div key="DeploymentRelease">{deploymentRelease}</div>;
+  return (
+    <div className="text-overflow" key="DeploymentRelease" style={{ whiteSpace: 'initial' }}>
```
oleorhagen, line 53: Q: Why initial and not `normal`?

mzedel: to stick with the default, which might get modified by an outer component - setting it to normal would interfere with any inheritance, which in this case I don't intend to do, but rather just restore what the `text-overflow` class defines

src/js/components/deployments/__snapshots__/deploymentitem.test.js.snap
```diff
@@ -7,7 +7,10 @@ exports[`DeploymentItem Component renders correctly 1`] = `
   >
     Release
   </span>
-  <div>
+  <div
+    class="text-overflow"
+    style="white-space: initial;"
```
michaelatmender, line 12: It would also work without the "white-space: initial;" - that would leave it as a single line with truncated text. Might be preferable if there are many rows of deployments? What do you think?
Screenshot 2022-06-16 at 11 41 55

michaelatmender (approved):

PR #2762, opened 2022-05-19
MEN-5608 - fix: fixed an issue that let device searches show non-group devices in a group focused search

mzedel:
Ticket: MEN-5608
Changelog: Title

### DIFF
--- a/src/js/actions/deviceActions.js
+++ b/src/js/actions/deviceActions.js
@@ -599,10 +599,10 @@ const convertIssueOptionsToFilters = issuesSelection =>
     return DeviceConstants.DEVICE_ISSUE_OPTIONS[item].filterRule;
   });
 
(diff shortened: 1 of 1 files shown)


tranchitella (approved):
It works fine; however, I noticed that reloading the page with a filter (e.g. this URL: https://localhost/ui/#/devices/accepted?name=device-1) applies the filter but doesn't find any device.

image

If I remove the filter, and then add it again, it shows the device:

image

mzedel:
> It works fine; however, I noticed that reloading the page with a filter (e.g. this URL: https://localhost/ui/#/devices/accepted?name=device-1) applies the filter but doesn't find any device.

since the name value is based on a tag this is likely due to MEN-5127 (or MEN-4989)... the coming sprint might be a good time to finally tackle this 👌

PR #2778, opened 2022-05-23
chore: fixed image build stage platform to ensure tolerable build speed

mzedel:
- since the final stage will only serve static files, independent of their build architecture, the build step does not have to match the target platform

@alfrunes the current `Dockerfile` seems to build locally for me also for `arm/v6`, but in an awfully long time, with this the final image is still `arm/v6`, but build time is in line with the "classic" build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:18.2.0-alpine AS build
+FROM --platform=amd64 node:18.2.0-alpine AS build
 WORKDIR /usr/src/app
 COPY package-lock.json package.json ./
 RUN npm ci


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:18.2.0-alpine AS build
+FROM --platform=amd64 node:18.2.0-alpine AS build
```
alfrunes, line 1: This is not a very portable way to solve the issue. I could imagine that CI/CD would build and publish the docker images on actual ARM machines at some point in the future instead of using buildx/qemu; but this is indeed sufficient for now.

PR #2818, opened 2022-06-02
MEN-5525 - feat: added per user access token functionality to the user settings page

mzedel:
this is the UI part for mendersoftware/useradm#297

Ticket: MEN-5525
Changelog: Title

### DIFF
--- a/src/js/actions/userActions.js
+++ b/src/js/actions/userActions.js
@@ -595,3 +595,25 @@ export const setHideAnnouncement = (shouldHide, userId) => (dispatch, getState) 
   }
   return Promise.resolve();
 };
+
+export const getTokens = () => (dispatch, getState) =>
+  GeneralApi.get(`${useradmApiUrl}/settings/tokens`).then(({ data: tokens }) => {
+    const user = getCurrentUser(getState());
+    const updatedUser = {
+      ...user,
+      tokens
+    };
+    return Promise.resolve(dispatch({ type: UserConstants.UPDATED_USER, user: updatedUser, userId: user.id }));
+  });
+
+const ONE_YEAR = 31536000;
+
+export const generateToken =
+  ({ expiresIn = ONE_YEAR, name }) =>
+  dispatch =>
+    GeneralApi.post(`${useradmApiUrl}/settings/tokens`, { name, expires_in: expiresIn })
+      .then(({ data: token }) => Promise.all([dispatch(getTokens()), token]))
+      .catch(err => commonErrorHandler(err, 'There was an error creating the token:', dispatch));
+
+export const revokeToken = token => dispatch =>
+  GeneralApi.delete(`${useradmApiUrl}/settings/tokens/${token.id}`).then(() => Promise.resolve(dispatch(getTokens())));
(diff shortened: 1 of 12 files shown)


tranchitella (changes requested):
I have a few issues with this PR.

The first problem is that once I generated a new token, and I click on the CLOSE button when the dialog shows me the token's value, if I try creating a new token clicking on "GENERATE A TOKEN" the modal dialog shows again the old token and doesn't allow me to create a new one.

The second issue is that sometimes when refreshing the page (in https://localhost/ui/#/settings/my-profile), the list of tokens doesn't show up. Moving to a different view/page and then back to the "My profile" view doesn't fix the issue. A few full refresh actions, though, fix the problem and the list is back.

The third issue is about the expiration: in the API specs you can set an expiration in seconds (expires_in). I don't know what the UX design included, but would it be possible to have a dropdown with the expiration settings (e.g. 1 hour, 24 hours, 7 days, 30 days, 180 days, 1 year or something like that)? If you have other ideas just go for them.

mzedel:
The first issue you noted should now be fixed...

> The second issue is that sometimes when refreshing the page (in https://localhost/ui/#/settings/my-profile), the list of tokens doesn't show up. Moving to a different view/page and then back to the "My profile" view doesn't fix the issue. A few full refresh actions, though, fix the problem and the list is back.

That I can't reproduce, from what I see the token list is retrieved every time you go to the profile as expected 🤷‍♂️.


> The third issue is about the expiration: in the API specs you can set an expiration in seconds (expires_in). I don't know what the UX design included, but would it be possible to have a dropdown with the expiration settings (e.g. 1 hour, 24 hours, 7 days, 30 days, 180 days, 1 year or something like that)? If you have other ideas just go for them.

This is intended to be added at a later iteration.

src/js/actions/userActions.js
```diff
@@ -595,3 +595,23 @@ export const setHideAnnouncement = (shouldHide, userId) => (dispatch, getState)
...
+    };
+    return Promise.resolve(dispatch({ type: UserConstants.UPDATED_USER, user: updatedUser, userId: user.id }));
+  });
+
+const ONE_YEAR = 31536000;
+
+export const generateToken = name => dispatch =>
+  GeneralApi.post(`${useradmApiUrl}/settings/tokens`, { name, expires_in: ONE_YEAR })
+    .then(({ data: token }) => Promise.all([dispatch(getTokens()), token]))
+    .catch(err => commonErrorHandler(err, `There was an error creatin g the token:`, dispatch));
```
michaelatmender, line 614: Small typo here - remove space in "creatin g"

michaelatmender (changes requested):
Looks good, with just a couple of small comments from me!
I would also list tokens most recent first, if possible - but is that rather something to change on the API?

src/js/actions/userActions.js
mzedel: That should now be fixed 👌

mzedel:
> Looks good, with just a couple of small comments from me! I would also list tokens most recent first, if possible - but is that rather something to change on the API?

Thanks for the review! In this case I think the token sorting is ok to be done in the frontend as we've limited the number of tokens - the tokens should now be sorted 👌.

tranchitella:
> This is intended to be added at a later iteration.

Why?

tranchitella:
As discussed out of band, we won't add the expiration column to the UI.

michaelatmender (approved):

tranchitella (approved):

PR #2835, opened 2022-06-16
Fix/e2e tests

mzedel:
- this needs mendersoftware/mender-test-containers#178

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -47,7 +47,6 @@ test:lint:
     - export DOCKER_TAG=${CI_COMMIT_REF_SLUG:-master}
     - export SERVICE_IMAGE=${DOCKER_REPOSITORY}:${DOCKER_TAG}
     - apk add --no-cache bash git jq wget
-    - wget -qP tests/e2e_tests/fixtures "https://dgsbl4vditpls.cloudfront.net/mender-demo-artifact.mender"
   artifacts:
     expire_in: 2w
     paths:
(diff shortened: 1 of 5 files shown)


src/js/components/settings/settings.js
```diff
@@ -84,7 +84,7 @@ export const Settings = ({
         }
       });
     } else {
-      const notStripePromise = {};
+      const notStripePromise = new Promise(resolve => setTimeout(resolve, 700));
```
oleorhagen, line 87: from nothing to `700` -_-

tests/e2e_tests/integration/00-setup.spec.ts
```diff
@@ -87,6 +87,12 @@ test.describe('Test setup', () => {
       ]);
       const page = await context.newPage();
       await page.goto(`${baseUrl}ui/#/settings`);
+      const isVisible = await page.isVisible(`text=/Change email/i`);
+      if (!isVisible) {
+        console.log('settings may not be loaded - move around');
+        await page.goto(`${baseUrl}ui/#/help`);
```
oleorhagen, line 93: Q: Why `help`, then `settings`?

.gitlab-ci.yml
```diff
@@ -204,17 +203,17 @@ test:prep:
 test:staging-deployment:chrome:
   extends: .template:test:staging-deployment
   script:
-    - npm run test
+    - npm run test-ci
```
oleorhagen, line 206: :raised_hands:

oleorhagen (approved):

src/js/components/settings/settings.js
mzedel: just anything that's not nothing - usually the stripe request should be faster (they are in a `Promise.race` 🏎️  after this) and at these times it still is considered to not be perceived by a user 👍 (although 700 is on the upper end of what we use in the codebase)

tests/e2e_tests/integration/00-setup.spec.ts
tranchitella: I suppose to trigger a refetch.

tranchitella (approved):
Looks good.

tests/e2e_tests/integration/00-setup.spec.ts
oleorhagen: So the `help` part could've been pretty much anything?

mzedel: the help pages are the most static area of the ui so I think no additional requests to the backend, navigating there is basically a UI only thing... afterwards the step to the settings gives the stripe request another chance to complete and show the settings - this should be just the fallback for the 700 timeout from above 🤞

PR #2840, opened 2022-06-17
fix: fixed an issue that caused the device list to crash on group selection if a device was selected at the time

mzedel:
Ticket: None
Changelog: Title


mzedel:
cherry-pick to:
 * 3.3.x

PR #2844, opened 2022-06-22
fix: version file retrieval, parsing & action access

mzedel:
Ticket: None
Changelog: None

### DIFF
--- a/src/js/actions/appActions.js
+++ b/src/js/actions/appActions.js
@@ -178,8 +178,12 @@ export const setVersionInfo = info => (dispatch, getState) =>
     })
   );
 
+const versionRegex = new RegExp(/\d+\.\d+/);
 const getLatestRelease = thing => {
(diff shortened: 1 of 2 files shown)


src/js/actions/appActions.js
```diff
@@ -178,8 +178,12 @@ export const setVersionInfo = info => (dispatch, getState) =>
     })
   );
 
+const versionRegex = new RegExp(/\d+\.\d+/);
```
oleorhagen, line 181: Q: Why not define it in the scope where it's used only?

oleorhagen (approved):

src/js/actions/appActions.js
mzedel: to squeeze out the precious performance improvement of only initializing this once on load and not on call to the function... 😅

oleorhagen: :laughing: :light_rail:

PR #2924, opened 2022-07-18
feat: made use of per user settings

mzedel:
Ticket: None
Changelog: None


src/js/actions/userActions.js
```diff
@@ -571,21 +571,33 @@ export const saveGlobalSettings =
...
+      const userSettings = getUserSettingsSelector(getState());
+      const updatedSettings = {
+        ...userSettings,
+        ...settings,
+        onboarding: {
+          ...userSettings.onboarding,
+          ...settings.onboarding
+        }
+      };
+      return GeneralApi.post(`${useradmApiUrl}/settings/me`, updatedSettings, { headers: { etag: result[result.length - 1] } }).then(() =>
```
tranchitella, line 596: This is wrong: you should use "If-Match: ETAG", where ETAG is the previous value. However, you can insert the header only if you have a value for ETAG. In other words, you cannot send "undefined".

tranchitella (changes requested):

tranchitella (approved):
I tested it locally and works as expected.

PR #2934, opened 2022-07-20
MEN-5453

mzedel:


src/js/components/settings/integrations.js
```diff
@@ -1,32 +1,119 @@
...
+      label={`${title} connection string`}
+      multiline
+      onChange={updateConnectionConfig}
+      value={value}
+    />
+  );
+};
+
+const providerConfigMap = {
+  amazon: ConnectionDetailsInput,
```
tranchitella, line 92: This should be iot-core

tranchitella (changes requested):

tranchitella:
@mzedel still not working as expected, the UI is sending:
`{"provider":"amazon","credentials":{"type":"aws","access_key_id":"A","secret_access_key":"B","endpoint_url":"https://mender-test-hub.azure-devices.net","device_policy_ARN":"C"}}`

It should send:
`{"provider":"iot-hub","credentials":{"type":"aws","access_key_id":"A","secret_access_key":"B","endpoint_url":"https://mender-test-hub.azure-devices.net","device_policy_ARN":"C"}}`, instead.

mzedel:
@tranchitella can you double check you tried it with the latest version? there should not be any trace of `amazon` in the src folder any more...

tranchitella:
@mzedel I'm sorry, I messed up my local branch, you are right. I still have an issue, though: despite correct data being returned by the end-point, the UI is not populating the fields with the existing values:

image

Any idea why this is happening?

mzedel:
@tranchitella yes you're right - the config destructuring and the casing for the config were not aligned with the endpoint - should now be aligned... I also added a simple URL validation for the endpoint, unfortunately it won't catch non-trusted domains (btw thanks a lot for the example data!)...

tranchitella (approved):

PR #2938, opened 2022-07-26
chore: increased chances of failing early on incompatible build hosts

mzedel:
unfortunately that seems to only work with somewhat newer node/ npm versions (v14/v8 worked, v12/v6 didn't), but it might be better than not having it...


merlin-northern (approved):
thank you!

PR #2987, opened 2022-08-01
MEN-5624 - downloads

mzedel:
in order to make the browser pass the JWT to the download subdomains, the JWT is stored in a cookie on the upper `.mender.io` domain and thus made available to the different subdomains (setting them for the subdomains only is not possible for security reasons)... to minimize the increased security risk, that cookie expires shortly after creation...
furthermore, the additional copyable options (curl + gitlab) should be considered as examples of course...


src/js/components/help/downloads.js
```diff
@@ -0,0 +1,324 @@
...
+    paths:
+      - ${filename}
+`;
+};
+
+const tools = [
+  {
+    id: 'mender',
+    packageId: 'mender-client',
+    title: 'Mender Client package',
```
tranchitella, line 109: Maybe "Debian package"?

```suggestion
    title: 'Mender Client Debian package',
```

src/js/components/help/downloads.js
```diff
@@ -0,0 +1,324 @@
...
+  {
+    id: 'mender-artifact',
+    title: 'Mender Artifact',
+    getLocations: defaultLocationFormatter,
+    canAccess: () => true,
+    osList: ['darwin', 'linux']
+  },
+  {
+    id: 'mender-binary-delta',
+    title: 'Mender binary delta',
```
tranchitella, line 124:
```suggestion
    title: 'Mender Binary Delta generator and Update Module',
```

src/js/components/help/downloads.js
```diff
@@ -0,0 +1,324 @@
...
+    id: 'mender-cli',
+    title: 'Mender CLI',
+    getLocations: defaultLocationFormatter,
+    canAccess: () => true,
+    osList: ['darwin', 'linux']
+  },
+  {
+    id: 'mender-configure-module',
+    packageId: 'mender-configure',
+    title: 'Mender configure',
```
tranchitella, line 139:
```suggestion
    title: 'Mender Configure',
```

src/js/components/help/downloads.js
```diff
@@ -0,0 +1,324 @@
...
+    getLocations: multiArchLocationFormatter,
+    canAccess: ({ tenantCapabilities }) => tenantCapabilities.hasDeviceConfig,
+    osList: defaultOSVersions,
+    archList: ['all']
+  },
+  {
+    id: 'mender-connect',
+    title: 'Mender connect',
```
tranchitella, line 147:
```suggestion
    title: 'Mender Connect',
```

src/js/components/help/downloads.js
```diff
@@ -0,0 +1,324 @@
...
+    getLocations: multiArchLocationFormatter,
+    canAccess: () => true,
+    osList: defaultOSVersions,
+    archList: defaultArchitectures
+  },
+  {
+    id: 'mender-convert',
+    title: 'Mender convert',
```
tranchitella, line 155:
```suggestion
    title: 'Mender Convert',
```

src/js/components/help/downloads.js
```diff
@@ -0,0 +1,324 @@
...
+      {
+        location: `https://github.com/mendersoftware/mender-convert/archive/refs/tags/${getVersion(versionInfo, 'mender-convert')}.zip`,
+        title: 'Mender convert'
+      }
+    ],
+    canAccess: ({ tenantCapabilities }) => tenantCapabilities.hasDeviceConnect
+  },
+  {
+    id: 'mender-gateway',
+    title: 'Mender gateway',
```
tranchitella, line 166:
```suggestion
    title: 'Mender Gateway',
```

src/js/components/help/downloads.js
```diff
@@ -0,0 +1,324 @@
...
+    getLocations: multiArchLocationFormatter,
+    isPublic: false,
+    canAccess: ({ isEnterprise }) => isEnterprise,
+    osList: defaultOSVersions,
+    archList: defaultArchitectures
+  },
+  {
+    id: 'monitor-client',
+    packageId: 'mender-monitor',
+    title: 'Monitor-client',
```
tranchitella, line 176: To be consistent with the other names above...

```suggestion
    title: 'Mender Monitor',
```

src/js/components/help/downloads.js
```diff
@@ -0,0 +1,324 @@
...
+  return [
+    {
+      location: `https://downloads.${location}/${id}/${getVersion(versionInfo, id)}/${id}-${getVersion(versionInfo, id)}.tar.xz`,
+      title
+    }
+  ];
+};
+
+const defaultCurlDownload = ({ location, tokens }) => {
+  let headers = `-H cookie: JWT=${getToken()}`;
```
tranchitella, line 77:
```suggestion
  let headers = `-H "Cookie: JWT=${getToken()}"`;
```

src/js/components/help/downloads.js
```diff
@@ -0,0 +1,324 @@
...
+  if (tokens.length) {
+    headers = `-H Authorization: Bearer ${tokens[0]}`;
```
tranchitella, line 79:
```suggestion
    headers = `-H "Authorization: Bearer ${tokens[0]}"`;
```

src/js/components/help/downloads.js
```diff
@@ -0,0 +1,324 @@
...
+  const filename = location.substring(location.lastIndexOf('/') + 1);
+  return `
+download:mender-tools:
+  image: curlimages/curl
+  stage: download
+  variables:
+    ${tokens.length ? `MENDER_TOKEN: ${tokens}` : `MENDER_JWT: ${getToken()}`}
+  script:
+    - if [ -n "$MENDER_TOKEN" ]; then
+    - curl -H Authorization: Bearer $MENDER_TOKEN -LO ${location}
```
tranchitella, line 94:
```suggestion
    - curl -H "Authorization: Bearer $MENDER_TOKEN" -LO ${location}
```

src/js/components/help/downloads.js
```diff
@@ -0,0 +1,324 @@
...
+
+  const handleToggle = event => {
+    const anchor = anchorEl ? null : event?.currentTarget.parentElement;
+    setAnchorEl(anchor);
+    setCurrentLocation(event?.target.getAttribute('value'));
+  };
+
+  const handleSelection = useCallback(
+    ({ target: { value } }) => {
+      const option = copyOptions[value];
```
tranchitella, line 211: I don't know why, but it always copies the curl command. I could never get the gitlab job...

src/js/components/help/downloads.js
```diff
@@ -0,0 +1,324 @@
...
+  }
+  return `curl ${headers} -LO ${location}`;
+};
+
+const defaultGitlabJob = ({ location, tokens }) => {
+  const filename = location.substring(location.lastIndexOf('/') + 1);
+  return `
+download:mender-tools:
+  image: curlimages/curl
+  stage: download
```
tranchitella, line 89: This is a nice idea; FYI, we'll cover this (and the other CI/CD templates) in the "application updates" epic I'm currently working on. My only objection is that the gitlab job can be useful for a subset of the downloadable packages, not all of them.

src/js/components/help/downloads.js
```diff
@@ -0,0 +1,323 @@
...
+import { ArrowDropDown, ExpandMore, FileDownloadOutlined as FileDownloadIcon, Launch } from '@mui/icons-material';
+
+import { setSnackbar } from '../../actions/appActions';
+import { getCurrentUser, getDocsVersion, getIsEnterprise, getTenantCapabilities } from '../../selectors';
+import Time from '../common/time';
+import { getToken } from '../../auth';
+import { detectOsIdentifier } from '../../helpers';
+import Tracking from '../../tracking';
+
+// wget --auth-no-challenge --user "$HOSTED_MENDER_EMAIL" --password "$HOSTED_MENDER_PASSWORD" https://downloads.customer.mender.io/content/hosted/mender-monitor/debian/1.2.0/mender-monitor_1.2.0-1%2Bdebian%2Bbuster_all.deb
```
lluiscampos, line 16: Leftover comment?

src/js/components/help/downloads.js
```diff
@@ -0,0 +1,323 @@
...
+  }
+  return [{ location: `https://downloads.${location}/${id}/${getVersion(versionInfo, id)}/linux/${id}`, title }];
+};
+
+const multiArchLocationFormatter = ({ tool, versionInfo }) => {
+  const location = getToolLocation(tool);
+  const { archList, id, osList, packageId } = tool;
+  return osList.reduce((accu, os) => {
+    const osArchitectureLocations = archList.map(arch => ({
+      location: `https://downloads.${location}/repos/debian/pool/main/m/${id}/${encodeURIComponent(
```
lluiscampos, line 55: The `/m/` in the path is just the first letter of the package name (aka `${id}`). This works today because all our packages start by `m` (_m_ ender-connect, _m_ ender-client, ...). To be better future ready we should deduce it from `id`.

src/js/components/help/downloads.js
```diff
@@ -0,0 +1,323 @@
...
+  {
+    id: 'mender-cli',
+    title: 'Mender CLI',
+    getLocations: defaultLocationFormatter,
+    canAccess: () => true,
+    osList: ['darwin', 'linux']
+  },
+  {
+    id: 'mender-configure-module',
+    packageId: 'mender-configure',
```
lluiscampos, line 138: We shortly discuss this on our call, but just as a heads-up.

With these structures we have the limitation of one component = one package. For example `mender-configure` component publishes three packages: `mender-configure`, `...-demo`, `...-timezone`; or `mender` also produces two packages. While these extra packages that we have today have limited value for this Downloads section in the GUI, we might need others in the future.

mzedel: this should now be supported (I added `demo` + `timezone` for `mender-configure` + `dev` for `mender`)

src/js/components/help/downloads.js
mzedel: mandatory: https://martinfowler.com/bliki/Yagni.html - it will now be deducted 👌

src/js/components/help/downloads.js
mzedel: leftover reminder to add `wget` in the download options... - fixed now 👍

mzedel:
@lluiscampos, @tranchitella do you think having update modules listed in here would be beneficial for a client developer?
Personally I'm not convinced, since they require even more additional manual steps, but having a central place does have benefits nonetheless...

tranchitella:
> @lluiscampos, @tranchitella do you think having update modules listed in here would be beneficial for a client developer? Personally I'm not convinced, since they require even more additional manual steps, but having a central place does have benefits nonetheless...

No, I think update modules are just too much.

lluiscampos:
I vote also for leaving them out.

michaelatmender (approved):

PR #2989, opened 2022-08-01
MEN-5753 - SSO setup

mzedel:
in lieu of sending the location of the idp document to the backend, the UI will try to parse the uploaded or pasted document as XML as early input validation...


tranchitella (changes requested):
I have a suggestion: when setting up the Service Provider application in your IdP, you need to specify two values:
- ACS URL
- Entity ID

Additionally, to log in to Mender, you need to know the SSO login URL, e.g. https://docker.mender.io/api/management/v1/useradm/auth/sso/UUID/login

See screenshot (from the Google admin console):
image

Would it be possible to expose these URLs/values in easy-to-copy-to-the-clipboard text fields?

mzedel:
@tranchitella like so?

Screenshot 2022-08-03 at 15 03 01

unless @michaelatmender objects of course!

tranchitella (approved):
It looks good, I only have three extra requests:
- Remove the double slash from the URLs
- Order as follows: Entity ID, ACS URL, login URL
- Rename "Start URL" to "Login URL"

tranchitella (approved):

PR #2994, opened 2022-08-03
MEN-5516

mzedel:


mzedel:
@michaelatmender now this should be ready for a last review round and I don't expect more rebases/ changes to the existing commits to come in the next few days 👌

src/js/themes/Mender/light.js
```diff
@@ -1,89 +1,135 @@
...
+  },
+  secondary: {
+    lighter: '#8e577b',
+    main: '#5d0f43'
+  },
+  tooltip: {
+    text: grey[50]
+  },
+  text: {
+    primary: 'rgba(10, 10, 11, 0.78)'
```
michaelatmender, line 39: a `text.hint` value of something like `rgba(0, 0, 0, 0.5)` is needed here to ensure the "connector" arrow lines are visible in the update pauses step when creating a deployment. It's also used to give the device list header text a lighter color

michaelatmender (changes requested):
I only found one small thing left to adjust that affected the light theme

michaelatmender (approved):

PR #3001, opened 2022-08-11
MEN-5104

mzedel:
needs https://github.com/mendersoftware/useradm-enterprise/pull/518 to work...

### DIFF
--- a/src/js/components/common/forms/__snapshots__/passwordinput.test.js.snap
+++ b/src/js/components/common/forms/__snapshots__/passwordinput.test.js.snap
@@ -394,137 +394,62 @@ label[data-shrink=false]+.MuiInputBase-formControl .emotion-3:focus::-ms-input-p
   color: #d32f2f;
 }
 
-.emotion-9 {
-  display: -webkit-inline-box;
-  display: -webkit-inline-flex;
-  display: -ms-inline-flexbox;
-  display: inline-flex;
-  -webkit-align-items: center;
-  -webkit-box-align: center;
-  -ms-flex-align: center;
-  align-items: center;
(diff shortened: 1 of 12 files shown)


merlin-northern:
I gave it a go, says `Password too weak` when I want to create user without password.

merlin-northern:
> I gave it a go, says `Password too weak` when I want to create a user without password. The rest of comments are on slack -- we need to decide how we create the users and how we pass the identity providers and OAuth vs SAML users.

ok I checked a bit and this one sends:

```json
{
  "email": "cooluser@gmail.com",
  "login": "cooluser@gmail.com"
}
```

which currently will not work we expect login to be an object (https://docs.mender.io/api/#management-api-user-administration-and-authentication-schemas-usernew). I will follow-up with the team today on how we want to create the users from the UI point of view.

merlin-northern:
cutting the long story short: please send just emails:

```
{
  "email": "msmith@samltest.id"
}
```

tranchitella (approved):

merlin-northern (commented):
I tested it with useradm-enterprise latest changes, but it still shows "Password too weak", let's you create the user without password, but the error is visible right before the dialog disappears, perhaps around here: https://github.com/mendersoftware/gui/blob/master/src/js/components/common/forms/form.js#L104 we could not show the too_weak error if password is empty?

mzedel:
@merlin-northern should now be fixed! I'll bring it to staging so we can give it a go 😃

merlin-northern:
> @merlin-northern should now be fixed! I'll bring it to staging so we can give it a go 😃

yes it is lovely now!

PR #3007, opened 2022-08-16
feat: get the device twin/shadow from the first integration which supports it

tranchitella:


src/js/components/devices/device-details/devicetwin.js
```diff
@@ -130,7 +131,13 @@ export const DeviceTwin = ({ device, getDeviceTwin, integrations, setDeviceTwin
   const [isSync, setIsSync] = useState(true);
   const editorRef = useRef(null);
 
-  const integration = integrations.find(integration => integration.provider === externalProvider.provider);
+  var integrationIds = Object.keys(device.twinsByIntegration ?? {}).filter(n => n != 'twinError');
```
mzedel, line 134: will the device twin really be populated by a `twinError`?

src/js/components/devices/device-details/devicetwin.js
```diff
@@ -130,7 +131,13 @@ export const DeviceTwin = ({ device, getDeviceTwin, integrations, setDeviceTwin
...
+  if (integrationIds.length == 0) {
+    integrationIds = integrations.filter(integration => EXTERNAL_PROVIDER[integration.provider]?.deviceTwin).map(integration => integration.id);
+  }
+
+  const integration = integrations.find(integration => integration.id === integrationIds[0]);
```
mzedel, line 139: No. This won't work once we have anyone trying out how 2 integrations interact, or someone actually uses 2 integrations...

src/js/components/devices/device-details/devicetwin.js
```diff
@@ -130,7 +131,13 @@ export const DeviceTwin = ({ device, getDeviceTwin, integrations, setDeviceTwin
   const [isSync, setIsSync] = useState(true);
   const editorRef = useRef(null);
 
-  const integration = integrations.find(integration => integration.provider === externalProvider.provider);
+  var integrationIds = Object.keys(device.twinsByIntegration ?? {}).filter(n => n != 'twinError');
+  if (integrationIds.length == 0) {
+    integrationIds = integrations.filter(integration => EXTERNAL_PROVIDER[integration.provider]?.deviceTwin).map(integration => integration.id);
+  }
```
mzedel, line 137: with both remaining providers offering device twin capabilities this should no longer be needed...

mzedel (changes requested):
Choosing the first would be an option as a fallback, but not when there is a real possibility of using multiple integrations, then it just provokes bug reports IMO...

src/js/components/devices/device-details/devicetwin.js
tranchitella: no, it is populated by the UI

src/js/components/devices/device-details/devicetwin.js
tranchitella: Honestly, I'm thinking about limiting the "cloud-based" integrations (Azure/AWS) to one per tenant.

src/js/components/devices/device-details/devicetwin.js
tranchitella: We'll soon have Webhooks, which do NOT support device twin capabilities.

mzedel: ah ok, I did not expect Webhooks to be handled in that area...

src/js/components/devices/device-details/devicetwin.js
mzedel: if we change it to show up for the proper twin, it will still work if it is one per tenant, but until then this will too likely cause problems.

src/js/components/devices/device-details/devicetwin.js
tranchitella: Web Hooks will be "yet another integration", but without support for the device state (twin/shadow).
We expect to configure webhooks using the same "integrations" API end-points.

tranchitella:
> Choosing the first would be an option as a fallback, but not when there is a real possibility of using multiple integrations, then it just provokes bug reports IMO...

Generally speaking, the device state end-points return the (integration ID, state) mapping.
Therefore, the UI can know the list of integration IDs to use. However, it is limited to one integration ID in the current implementation, and my idea was to limit the UI to the first one. Do you believe we should show the component multiple times, one per device integration which supports the device state APIs? The practical use-case is not very probable (multiple integrations for the same device), though.

mzedel:
> Generally speaking, the device state end-points return the (integration ID, state) mapping.
Then we should use this mapping.
> Therefore, the UI can know the list of integration IDs to use. However, it is limited to one integration ID in the current implementation, and my idea was to limit the UI to the first one.
If it is limited to one in the backend - this is ok, but as long as this is not the case - we have to show multiple in the UI or risk confusing users/ bug reports.
> Do you believe we should show the component multiple times, one per device integration which supports the device state APIs?
Yes - just like we do for integrations... (as a user I would expect this to show up for or otherwise assume we merge the different data sources into our twin)
> The practical use-case is not very probable (multiple integrations for the same device), though.
I disagree, if one provider has better integrations with a certain functionality, I will use it for that use case and use another for the next use case... (say configuring build integration with Azure and supply level integration with AWS)

tranchitella:
@mzedel ok, let me play with it and see if I can make it work with multiple device twins.
If I fail, we'll create a task.

mzedel (approved):

PR #3012, opened 2022-08-17
test: e2e tests for SAML SSO signup

merlin-northern:
Ticket: MEN-5651

### DIFF
--- a/tests/e2e_tests/integration/07-saml.spec.ts
+++ b/tests/e2e_tests/integration/07-saml.spec.ts
@@ -1,0 +1,243 @@
+import axios from 'axios';
+import * as fs from 'fs';
+import * as https from 'https';
+import jwtDecode from 'jwt-decode';
+
+import test, {expect} from '../fixtures/fixtures';
+
+const timeoutFourSeconds = 4000;
+const timeoutOneSecond = 1000;
+const samlSettings = {
+  'credentials': {
+    'chromium': {
+      'login': 'morty',
+      'password': 'panic',
+      'email': 'msmith@samltest.id'
+    },
+    'firefox': {
+      'login': 'rick',
+      'password': 'psych',
+      'email': 'rsanchez@samltest.id'
+    },
+    'webkit': {
+      'login': 'sheldon',
+      'password': 'bazinga',
+      'email': 'scooper@samltest.id'
+    },
+  },
+  'idp_url': 'https://samltest.id/saml/idp'
+};
+let metadataId;
+let jwt;
+
+test.describe('SAML Login', () => {
+  test.use({storageState: 'storage.json'});
+  test.describe('SAML login via sso/id/login', () => {
+    test.afterAll(async ({baseUrl, browserName}, testInfo) => {
+      console.log(`Finished ${testInfo.title} with status ${testInfo.status}. Cleaning up.`);
+      const response = await axios({
+        url: `${baseUrl}api/management/v1/useradm/users?email=${encodeURIComponent(samlSettings.credentials[browserName].email)}`,
+        method: 'GET',
+        headers: {
+          Authorization: `Bearer ${jwt}`,
+          'Content-Type': 'application/json'
+        },
+        httpsAgent: new https.Agent({
+          rejectUnauthorized: false
+        })
+      });
+
+      const responseStatus = await response.status;
+      if (responseStatus != 200 || response.data.length < 1) {
+        console.log(`${samlSettings.credentials[browserName].email} does not exist.`);
+      } else {
+        expect(response.status).toBe(200);
+        const userId = response.data[0]['id'];
+        const responseDelete = await axios({
+          url: `${baseUrl}api/management/v1/useradm/users/${userId}`,
+          method: 'DELETE',
+          headers: {
+            Authorization: `Bearer ${jwt}`,
+            'Content-Type': 'application/json'
+          },
+          httpsAgent: new https.Agent({
+            rejectUnauthorized: false
+          })
+        });
+        expect(response.status).toBe(200);
+        console.log(`removed user ${samlSettings.credentials[browserName].email}/${userId}.`);
+      }
+    });
+
+    // Setups the SAML/SSO login with samltest.id Identity Provider
+    test('Set up SAML', async ({environment, context, baseUrl, page, password, username}) => {
+      test.skip(environment !== 'staging');
+      test.setTimeout(320000);
+      const storage = await context.storageState();
+      jwt = storage['cookies'].find(cookie => cookie.name === 'JWT').value;
+
+      const idpResponse = await axios({
+        url: samlSettings.idp_url,
+        method: 'GET'
+      });
+      expect(idpResponse.status).toBe(200);
+      const metadata = idpResponse.data;
+
+      console.log(`running with ${baseUrl}`);
+      console.log(`IdP metadata len=${metadata.length} loaded and uploading`);
+
+      await page.goto(`${baseUrl}ui/settings/organization-and-billing`);
+      await expect(page).toHaveURL(`${baseUrl}ui/settings/organization-and-billing`);
+      // Check input[type="checkbox"]
+      await page.locator('input[type="checkbox"]').check();
+      // Click text=input with the text editor
+      await page.locator('text=input with the text editor').click();
+
+      // Click .view-lines
+      await page.locator('.view-lines').click();
+
+      console.log('typing metadata');
+      await page.locator('[aria-label="Editor content\\;Press Alt\\+F1 for Accessibility Options\\."]').type(metadata.replace(/(?:\r\n|\r|\n)/g, ''))
+      console.log('typing metadata done.');
+      // The screenshot saves the view of the typed metadata
+      await page.screenshot({'path': 'saml-edit-saving.png'});
+      // Click text=Save >> nth=1
+      await page.locator('text=Save').nth(1).click();
+      console.log('typing metadata done. waiting for Entity ID to be present on page.');
+      await page.waitForSelector('text=Entity ID');
+      console.log('waiting over continuing.');
+
+      const response = await axios({
+        url: `${baseUrl}api/management/v1/useradm/sso/idp/metadata`,
+        method: 'GET',
+        headers: {
+          Authorization: `Bearer ${jwt}`,
+          'Content-Type': 'application/json'
+        },
+        httpsAgent: new https.Agent({
+          rejectUnauthorized: false
+        })
+      });
+      expect(response.status).toBe(200);
+      metadataId = response.data[0]['id'];
+      console.log(`got metadata id: ${metadataId}`)
+
+      await page.goto(`${baseUrl}ui/settings/organization-and-billing`);
+      await page.locator('text=View metadata in the text editor').click();
+      // Click text=Download file
+      const [download] = await Promise.all([
+        page.waitForEvent('download'),
+        page.locator('text=Download file').click()
+      ]);
+      const downloadTargetPath = await download.path();
+      expect(downloadTargetPath).toBeTruthy();
+      const spMetadaData = fs.readFileSync(downloadTargetPath, 'utf8');
+      console.log(`downloaded SP metadata length=${spMetadaData.length}`)
+
+      await page.goto(`${baseUrl}ui/settings/organization-and-billing`);
+
+      const expectedLoginUrl = `${baseUrl}api/management/v1/useradm/auth/sso/${metadataId}/login`;
+      const findResults = await page.locator('span', {'hasText': expectedLoginUrl})
+      await expect(findResults).toHaveText(expectedLoginUrl);
+      console.log(`got span with login url: ${await findResults.count()}`)
+
+      const expectedAcsUrl = `${baseUrl}api/management/v1/useradm/auth/sso/${metadataId}/acs`;
+      const findResultsAcs = await page.locator('span', {'hasText': expectedAcsUrl})
+      await expect(findResultsAcs).toHaveText(expectedAcsUrl);
+      console.log(`got span with acs url: ${await findResultsAcs.count()}`)
+
+      const expectedSpMetaUrl = `${baseUrl}api/management/v1/useradm/sso/sp/metadata/${metadataId}`;
+      const findResultsSpMeta = await page.locator('span', {'hasText': expectedSpMetaUrl})
+      await expect(findResultsSpMeta).toHaveText(expectedSpMetaUrl);
+      console.log(`got span with SP metadata url: ${await findResultsSpMeta.count()}`)
+
+      const serviceProviderMetadata = spMetadaData.replaceAll('Signed="true"', 'Signed="false"')
+      fs.writeFileSync('fixtures/service_provider_metadata.xml', serviceProviderMetadata);
+
+      await page.goto('https://samltest.id/upload.php');
+      await page.waitForSelector('input[id=uploader]');
+      const handle=await page.$('input[id="uploader"]');
+      await handle.setInputFiles('fixtures/service_provider_metadata.xml');
+      // Click input:has-text("Upload")
+      await page.locator('input:has-text("Upload")').click();
+      await expect(page).toHaveURL('https://samltest.id/upload.php');
+
+      console.log('uploaded file, making screen shot, after waiting 4s');
+      await page.waitForTimeout(timeoutFourSeconds);
+      // Let's save the image after the upload
+      await page.screenshot({path: 'saml-uploaded.png'});
+    });
+
+    // Creates a user with login that matches Identity privder (samltest.id) user email
+    test('Creates a user without a password', async ({ environment, browserName, context, baseUrl, page }) => {
+      test.skip(environment !== 'staging');
+      const response = await axios({
+        url: `${baseUrl}api/management/v1/useradm/users?email=${encodeURIComponent(samlSettings.credentials[browserName].email)}`,
+        method: 'GET',
+        headers: {
+          Authorization: `Bearer ${jwt}`,
+          'Content-Type': 'application/json'
+        },
+        httpsAgent: new https.Agent({
+          rejectUnauthorized: false
+        })
+      });
+
+      const responseStatus = await response.status;
+      if (responseStatus != 200 || response.data.length < 1) {
+        await page.goto(`${baseUrl}ui/settings/user-management`);
+        // Click text=Create new user
+        await page.locator('text=Create new user').click();
+        // Click [placeholder="Email"]
+        await page.locator('[placeholder="Email"]').click();
+        // Fill [placeholder="Email"]
+        await page.locator('[placeholder="Email"]').fill(samlSettings.credentials[browserName].email);
+        // Click text=Create user
+        await page.locator('text=Create user').click();
+        await page.waitForTimeout(timeoutOneSecond);
+        await expect(page.locator(`text=${samlSettings.credentials[browserName].email}`)).toHaveText(samlSettings.credentials[browserName].email);
+        console.log(`${samlSettings.credentials[browserName].email} created.`);
+      } else {
+        console.log(`${samlSettings.credentials[browserName].email} already exists.`);
+      }
+    });
+
+    // This test calls auth/sso/${id}/login, where id is the id of the identity provider
+    // and verifies that login is successful.
+    test('User can login via sso/login endpoint', async ({environment, browserName, baseUrl, browser}) => {
+      test.skip(environment !== 'staging');
+      console.log(`logging in via ${baseUrl}api/management/v1/useradm/auth/sso/${metadataId}/login (credentials set:${browserName})`);
+      const context = await browser.newContext()
+      const page = await context.newPage()
+      await page.goto(`${baseUrl}api/management/v1/useradm/auth/sso/${metadataId}/login`);
+      await page.waitForTimeout(timeoutFourSeconds);
+      // This screenshot saves the view right after the first redirection
+      await page.screenshot({path: 'saml-redirected.png'});
+
+      // Click input[name="j_username"]
+      await page.locator('input[name="j_username"]').click();
+
+      // Fill input[name="j_username"]
+      await page.locator('input[name="j_username"]').fill(samlSettings.credentials[browserName].login);
+
+      // Click input[name="j_password"]
+      await page.locator('input[name="j_password"]').click();
+
+      // Fill input[name="j_password"]
+      await page.locator('input[name="j_password"]').fill(samlSettings.credentials[browserName].password);
+
+      // Click button:has-text("Login")
+      await page.locator('button:has-text("Login")').click();
+      await page.waitForSelector('text=Accept');
+      // This screen shot saves the summary of the data that will be sent in assertion
+      await page.screenshot({path: 'saml-logging-in.png'});
+
+      // Click text=Accept
+      await page.locator('text=Accept').click();
+      // confirm we have logged in successfully
+      await page.screenshot({path: 'saml-logging-in-accept.png'});
+      await page.waitForSelector('text=License information');
+      await page.waitForSelector(`text=${samlSettings.credentials[browserName].email}`);
+    });
+  });
+});


mzedel:
Yeah the usage of commands and test should be good, I would say the commands should all be part of the tests though the UI... if we can get the idp metadata I would also `fetch` this from an address, just in case it gets changed/ updated/ times out otherwise... also: I thought this was going to run on staging, so you might have to add something like https://github.com/mendersoftware/gui/blob/master/tests/e2e_tests/integration/99-settings.spec.ts#L96 
FYI, you can use `npx playwright codegen staging.hosted.mender.io` to record your interactions - not perfect, but a good starting point IMO... https://playwright.dev/docs/codegen shows what you can expect from running that command 😉

merlin-northern:
> Yeah the usage of commands and test should be good, I would say the commands should all be part of the tests though the UI... if we can get the idp metadata I would also `fetch` this from an address, just in case it gets changed/ updated/ times out otherwise... also: I thought this was going to run on staging, so you might have to add something like https://github.com/mendersoftware/gui/blob/master/tests/e2e_tests/integration/99-settings.spec.ts#L96 FYI, you can use `npx playwright codegen staging.hosted.mender.io` to record your interactions - not perfect, but a good starting point IMO... https://playwright.dev/docs/codegen shows what you can expect from running that command 😉

thanks. roger all that. fixed.

tests/e2e_tests/integration/07-saml.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      const request = await axios({
+        url: `${baseUrl}api/management/v1/useradm/sso/idp/metadata`,
+        method: 'POST',
+        headers: {
+          Authorization: `Bearer ${jwt}`,
+          'Content-Type': 'application/samlmetadata+xml'
+        },
+        data: metadata,
+       httpsAgent: new https.Agent({
+          rejectUnauthorized: false
```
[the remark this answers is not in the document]
merlin-northern: this is running against staging, so I could remove the acceptance of anything :) but it also runs against local setups (i.e.: `TEST_ENVIRONMENT=localhost`) and without it it will not that easily so I vote: keep it.

mzedel:
> thanks. roger all that. fixed.

😅 not quite how I meant it... I would make the settings through the UI - to also test that this is working correctly... (in a separate test in the outer describe section, similar to what's happening in the settings.spec for the password change maybe...) I'll happily handle this if you like, but would also gladly see you work more with playwright - so feel free to continue or pass it along as you please!

merlin-northern:
> > thanks. roger all that. fixed.
> 
> 😅 not quite how I meant it... I would make the settings through the UI - to also test that this is working correctly... (in a separate test in the outer describe section, similar to what's happening in the settings.spec for the password change maybe...) I'll happily handle this if you like, but would also gladly see you work more with playwright - so feel free to continue or pass it along as you please!

aha! ok, let me fix it today.

merlin-northern:
> > thanks. roger all that. fixed.
> 
> 😅 not quite how I meant it... I would make the settings through the UI - to also test that this is working correctly... (in a separate test in the outer describe section, similar to what's happening in the settings.spec for the password change maybe...) I'll happily handle this if you like, but would also gladly see you work more with playwright - so feel free to continue or pass it along as you please!

ok, fixed strikes back. is it better now? how adamant are you to have two tests in that case?

tests/e2e_tests/integration/07-saml.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+      const jwt=storage['cookies'].find(cookie => cookie.name === 'JWT').value;
+
+      const idp_response = await axios({
+        url: saml_settings['idp_url'],
+        method: 'GET'
+      });
+
+      if (idp_response.status !== 200) {
+        throw 'expected 201';
+      }
```
mzedel, line 32:
```suggestion
  expect(idp_responose.status).toBe(200);
```
or `expect(idp_responose.status).toBe(201);` - not quite sure here 😉

tests/e2e_tests/integration/07-saml.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+      test.skip(environment !== 'staging');
+      test.setTimeout(240000);
+      const saml_settings={
+        'login':'morty',
+        'password':'panic',
+        'email':'msmith@samltest.id',
+        'idp_url':'https://samltest.id/saml/idp'
+      };
+      const timeout_4s=4000;
+      const timeout_1s=1000;
```
mzedel, line 21: since these are constant, they should best be outside the test definition (being strict, they would be a fixture, but since they are unlikely to be reused across this test outside the test is fine).
also: this is JS/TS land, we do camelCase when we define with our things 😛 (or: your PR made me realize, that the linting seems to not be enforced in these tests... 🤯)

tests/e2e_tests/integration/07-saml.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+      // Click text=input with the text editor
+      await page.locator('text=input with the text editor').click();
+      // Press Tab
+      await page.locator('text=SAML metadataImport from a fileEnter to Rename, ⇧Enter to PreviewCancelSave').press('Tab');
+      // Press Tab
+      await page.locator('text=SAML metadataImport from a file >> div[role="presentation"]').press('Tab');
+      // Press Tab
+      await page.locator('text=Import from a file').press('Tab');
+      // Press Tab
+      await page.locator('text=SAML metadataImport from a file >> button').nth(1).press('Tab');
```
mzedel, line 53: This seems very brittle, whenever the layout changes in an unrelated way, these interactions will most likely break... Keep in mind: the test generation you use only generates a starting point for the test, but you should still think that all the steps are useful.

tests/e2e_tests/integration/07-saml.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+      await page.locator('input[name="userfile"]').click();
+      // Upload m
+      await page.locator('input[name="userfile"]').setInputFiles('fixtures/service_provider_metadata.xml');
+      // Click input:has-text("Upload")
+      await page.locator('input:has-text("Upload")').click();
+      await expect(page).toHaveURL('https://samltest.id/upload.php');
+
+      console.log('uploaded file, making screen shot, after waiting 4s');
+      await page.waitForTimeout(timeout_4s);
+      await page.screenshot({path: 'saml-uploaded.png'});
```
mzedel, line 132: this & the other screenshots look like they were used for debug purposes - do you think we need them on regular runs (mind you: we get a video on every first retry...)?

tests/e2e_tests/integration/07-saml.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+      }
+
+      const metadata = idp_response.data;
+
+      console.log(`running with ${baseUrl}`);
+      console.log(`IdP metadata len=${metadata.length} loaded and uploading`);
+
+      await page.goto(`${baseUrl}ui/settings/organization-and-billing`);
+
+      await expect(page).toHaveURL(`${baseUrl}ui/settings/organization-and-billing`);
```
mzedel, line 41: just a sidenote: while I can see the benefit of not having to worry about which calls might return async - most of your `expect` calls don't expect async results, so `await`ing them is most likely not needed...

tests/e2e_tests/integration/07-saml.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+          rejectUnauthorized: false
+        })
+      });
+
+      if (sp_meta_response.status !== 200) {
+        throw 'expected 201';
+      }
+
+      const service_provider_metadata=sp_meta_response.data.replaceAll('Signed="true"','Signed="false"')
+      fs.writeFileSync('fixtures/service_provider_metadata.xml', service_provider_metadata);
```
mzedel, line 117: this should be done through the UI too... you can either copy it or download the file through the UI...

tests/e2e_tests/integration/07-saml.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+      await expect(page.locator(`text=${saml_settings['email']}`)).toHaveText(saml_settings['email']);
+
+      console.log(`logging in via ${baseUrl}api/management/v1/useradm/auth/sso/${id}/login`);
+      // enter valid username and password
+      await page.goto(`${baseUrl}api/management/v1/useradm/auth/sso/${id}/login`);
+      await page.waitForTimeout(timeout_4s);
+      await page.screenshot({path: 'saml-redirected.png'});
+
+      // Double click text=morty
+      await page.locator('text=morty').dblclick();
```
mzedel, line 154: what does the double click here do?

tests/e2e_tests/integration/07-saml.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+
+      const expected_acs_url=`${baseUrl}api/management/v1/useradm/auth/sso/${id}/acs`;
+      const find_results_acs = await page.locator('span',{'hasText' : expected_acs_url})
+      await expect(find_results_acs).toHaveText(expected_acs_url);
+      console.log(`got span with acs url: ${await find_results_acs.count()}`)
+
+      const expected_sp_meta_url=`${baseUrl}api/management/v1/useradm/sso/sp/metadata/${id}`;
+      const find_results_sp_meta = await page.locator('span',{'hasText' : expected_sp_meta_url})
+      await expect(find_results_sp_meta).toHaveText(expected_sp_meta_url);
+      console.log(`got span with SP metadata url: ${await find_results_sp_meta.count()}`)
```
mzedel, line 98: this seems to duplicate the metadata checks the final login process does... + if you want to use the obtained values, they are available in the UI too...

tests/e2e_tests/integration/07-saml.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+      // enter valid username and password
+      await page.goto(`${baseUrl}api/management/v1/useradm/auth/sso/${id}/login`);
+      await page.waitForTimeout(timeout_4s);
+      await page.screenshot({path: 'saml-redirected.png'});
+
+      // Double click text=morty
+      await page.locator('text=morty').dblclick();
+
+      // Click input[name="j_username"]
+      await page.locator('input[name="j_username"]').click();
```
mzedel, line 157: my guess would be this selector is from the samltest login, in this case we have to be extra careful with the selectors as they might change (not sure if they are generated, but their next deployment might be `k_username`), so maybe something like `[placeholder=username]` would work (if there is a placeholder displayed)

tests/e2e_tests/integration/07-saml.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+      // Fill [placeholder="Email"]
+      await page.locator('[placeholder="Email"]').fill(saml_settings['email']);
+      // Click text=Create user
+      await page.locator('text=Create user').click();
+      await page.waitForTimeout(timeout_1s);
+      await expect(page.locator(`text=${saml_settings['email']}`)).toHaveText(saml_settings['email']);
+
+      console.log(`logging in via ${baseUrl}api/management/v1/useradm/auth/sso/${id}/login`);
+      // enter valid username and password
+      await page.goto(`${baseUrl}api/management/v1/useradm/auth/sso/${id}/login`);
```
mzedel, line 149: I _think_ between these two steps it would be good to clear the cookies (or even better: separate them into another test) in order to not accidentally rely on existing info in the session storage (like a JWT token)...

mzedel (changes requested):
some more comments to go over... the general direction is good, but from a user perspective it's very unlikely to pull up the API while using the UI - the UI does that for you and any follow up step relies on any prior obtained value to be correct...
While splitting the test up in multiple `test('<name>', async ({...}) => { ... })` calls might not be  strictly needed, it would help isolating areas where the integration might fail...

mzedel:
also: you can adjust the `.gitlab-ci.yml` a bit (mainly change the `stage` to `test` in `.template:test:staging-deployment`, remove the `only` limitation and the delay + remove the `only` limitation in the `test:prep` stage) to see if the tests run in the pipeline

tests/e2e_tests/integration/07-saml.spec.ts
merlin-northern: I added a check that we actually have the user email on the page, it is impossible to get redirected to a page that contains this username after going to the sso/login page. I will explore the tests separation.

tests/e2e_tests/integration/07-saml.spec.ts
merlin-northern: this is the samltes.id login page, there are no placeholders there. I do not think they change it very often. I can do more elaborate search for fields, do you think it is necessary?

tests/e2e_tests/integration/07-saml.spec.ts
merlin-northern: not entirely: these links you want to have to copy-paste to your identity provider,  so this checks that we present them correctly.

tests/e2e_tests/integration/07-saml.spec.ts
merlin-northern: all it does is: it does not hurt. :) removed.

tests/e2e_tests/integration/07-saml.spec.ts
merlin-northern: changed as requested.

tests/e2e_tests/integration/07-saml.spec.ts
merlin-northern: I believe toHaveURL needs await:

```js
  /**
   * Ensures the page is navigated to the given URL.
   *
   * ```js
   * await expect(page).toHaveURL(/.*checkout/);
   * ```
   *
   * @param urlOrRegExp Expected URL string or RegExp.
   * @param options
   */
  toHaveURL(urlOrRegExp: string|RegExp, options?: {
    /**
     * Time to retry the assertion for. Defaults to `timeout` in `TestConfig.expect`.
     */
    timeout?: number;
  }): Promise<void>;
```

tests/e2e_tests/integration/07-saml.spec.ts
merlin-northern: I am thinking all the time. replaced with single click.

tests/e2e_tests/integration/07-saml.spec.ts
merlin-northern: moved up.

tests/e2e_tests/integration/07-saml.spec.ts
merlin-northern: fixed everywhere. heh, yeah, throw what you dont really expect ;-)

merlin-northern:
> also: you can adjust the `.gitlab-ci.yml` a bit (mainly change the `stage` to `test` in `.template:test:staging-deployment`, remove the `only` limitation and the delay + remove the `only` limitation in the `test:prep` stage) to see if the tests run in the pipeline

okey!

tests/e2e_tests/integration/07-saml.spec.ts
mzedel: yes, but the links the UI shows are taken straight from the metadata just like you show them here - if either the link we use to login (that could be copied and used to login the newly created user) or any of the values shown in the ui (that are derived from the metadata) don't align, the login won't work. The link creation in the UI itself is tested through unit tests already.

tests/e2e_tests/integration/07-saml.spec.ts
mzedel: what might be more important is the search for reliably stable selectors... 👌  I'll take a look to see if there might be  other selectors that might fit...

tests/e2e_tests/integration/07-saml.spec.ts
mzedel: keep in mind: every interaction that is not required, is still executed and the completion awaited, so execution time...

merlin-northern:
> some more comments to go over... the general direction is good, but from a user perspective it's very unlikely to pull up the API while using the UI - the UI does that for you and any follow up step relies on any prior obtained value to be correct... While splitting the test up in multiple `test('<name>', async ({...}) => { ... })` calls might not be strictly needed, it would help isolating areas where the integration might fail...

I will try to separate the tests.

tests/e2e_tests/integration/07-saml.spec.ts
mzedel: that's why I wrote _most of your `expect` calls_ 😉 - OTOH since you `.goto()` in the line before you can likely remove the assertion entirely.

tests/e2e_tests/integration/07-saml.spec.ts
merlin-northern: I believe the input name is pretty stable -- if you change that you will have to change the code that handles the form on the other side. I would keep them (I could not find more reliable one)

tests/e2e_tests/integration/07-saml.spec.ts
merlin-northern: okey, no problema, so you vote for removing those three blocks?

tests/e2e_tests/integration/07-saml.spec.ts
merlin-northern: point taken. this was me being a bit fast on the trigger when I recorded the session, and the forgot to remove.

tests/e2e_tests/integration/07-saml.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+      await page.locator('button:has-text("Login")').click();
+      await page.waitForSelector('text=Accept');
+
+      await page.screenshot({path: 'saml-logging-in.png'});
+
+      // Click text=Accept
+      await page.locator('text=Accept').click();
+      // confirm we have logged in successfully
+      await page.waitForSelector('text=License information');
+      await page.waitForSelector('text=msmith@samltest.id');
```
merlin-northern, line 169: `${saml_settings['email']}` -- fixed already, but waiting for the next push.

merlin-northern:
@mzedel ready for the final round. The tests are separated, the login now uses a new incognito window, camelCase if everywhere, global test variables are global. I am struggling with .gitlab-ci.yml to make it run in this pipeline, apparently without success so far.

tests/e2e_tests/integration/07-saml.spec.ts
mzedel: Yes, that is e.g. how server side rendering works - my guess would be the saml test code is in PHP, where such a change happens in a single place... But we can keep it and fix if it breaks...

mzedel:
> also: you can adjust the `.gitlab-ci.yml` a bit (mainly change the `stage` to `test` in `.template:test:staging-deployment`, remove the `only` limitation and the delay + remove the `only` limitation in the `test:prep` stage) to see if the tests run in the pipeline


To be more explicit:
- in `.template:test:staging-deployment`:
  - change the `stage` to `test`
  - remove the `only` limitation
  - remove the delay 
- in the `test:prep`:
  - remove the `only` limitation

tests/e2e_tests/integration/07-saml.spec.ts
mzedel: My point is, that on case of a successful run we won't care about the screenshots and in case of a failure we have a video that likely contains the image... But feel free to keep them if they help you!

merlin-northern:
> > also: you can adjust the `.gitlab-ci.yml` a bit (mainly change the `stage` to `test` in `.template:test:staging-deployment`, remove the `only` limitation and the delay + remove the `only` limitation in the `test:prep` stage) to see if the tests run in the pipeline
> 
> To be more explicit:
> 
> * in `.template:test:staging-deployment`:
>   
>   * change the `stage` to `test`
>   * remove the `only` limitation
>   * remove the delay
> * in the `test:prep`:
>   
>   * remove the `only` limitation

did all of that and some more too, but I missed the only in test:prep.

merlin-northern:
@mzedel I had to add afterAll: to clean up the users:
* we can have only one user of given email in Mender -- so we have to cleanup
* we have to delete the user since there are 3 tests running
I hope it is enough. If not I have another idea, but perhaps this one will work.

merlin-northern:
@mzedel changes I had to make:
* firefox: all broken the meta upload to saml: I had to go with [handle](https://github.com/mendersoftware/gui/pull/3012/files#diff-43e6c191ca50ee3da88db714f9f87b2528568ef31a7d9b9ccebd7a198707b1f9R159)
* webkit: failed with timeouts: had tno [increase](https://github.com/mendersoftware/gui/pull/3012/files#diff-43e6c191ca50ee3da88db714f9f87b2528568ef31a7d9b9ccebd7a198707b1f9R75)
* users: had to use [three of them](https://github.com/mendersoftware/gui/pull/3012/files#diff-43e6c191ca50ee3da88db714f9f87b2528568ef31a7d9b9ccebd7a198707b1f9R75) (and [support](https://github.com/mendersoftware/gui/pull/3012/files#diff-43e6c191ca50ee3da88db714f9f87b2528568ef31a7d9b9ccebd7a198707b1f9R11)) from samltest.id, there was no good other way

.gitlab-ci.yml
```diff
@@ -202,16 +196,22 @@ test:prep:
...
+  variables:
+    TEST_SAML_CREDENTIALS: "firefox"
   script:
     - npm run test-ci -- --browser=firefox
 
 test:staging-deployment:webkit:
   extends: .template:test:staging-deployment
+  variables:
+    TEST_SAML_CREDENTIALS: "webkit"
   script:
```
mzedel, line 215: These values are already provided OOTB by playwright, please use them. (for examples you can search the codebase for `webkit` or check [here](https://github.com/mendersoftware/gui/blob/master/tests/e2e_tests/integration/05-deviceDetails.spec.ts#L40))

tests/e2e_tests/fixtures/fixtures.ts
```diff
@@ -50,6 +51,10 @@ const test = (process.env.TEST_ENVIRONMENT === 'staging' ? nonCoveredTest : cove
     const environment = process.env.TEST_ENVIRONMENT ? process.env.TEST_ENVIRONMENT : 'localhost';
     await use(environment);
   },
+  samlCredentials: async ({}, use) => {
+    const samlCredentials = process.env.TEST_SAML_CREDENTIALS ? process.env.TEST_SAML_CREDENTIALS : 'chrome';
+    await use(samlCredentials);
+  },
   username: async ({ environment }, use) => {
```
mzedel, line 58: see above...

tests/e2e_tests/integration/07-saml.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+          headers: {
+            Authorization: `Bearer ${jwt}`,
+            'Content-Type': 'application/json'
+          },
+          httpsAgent: new https.Agent({
+            rejectUnauthorized: false
+          })
+        });
+        expect(response.status).toBe(200);
+        console.log(`removed user ${samlSettings['credentials'][samlCredentials]['email']}/${userId}.`);
```
mzedel, line 68: Nitpick:
```suggestion
        console.log(`removed user ${samlSettings.credentials[samlCredentials].email}/${userId}.`);
```
- smells too much like python and a lack of comfort otherwise 😝 (also the other occurrences)

tests/e2e_tests/integration/07-saml.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+      await page.screenshot({path: 'saml-redirected.png'});
+
+      // Click input[name="j_username"]
+      await page.locator('input[name="j_username"]').click();
+
+      // Fill input[name="j_username"]
+      await page.locator('input[name="j_username"]').fill(samlSettings['credentials'][samlCredentials]['login']);
+
+      // Press Tab
+      await page.locator('input[name="j_username"]').press('Tab');
```
mzedel, line 224: unless this is needed, please remove it, as it only increases reliance on the page layout.

mzedel (changes requested):

tests/e2e_tests/integration/07-saml.spec.ts
merlin-northern: removed. nice catch, I thought I removed all of them.

tests/e2e_tests/integration/07-saml.spec.ts
merlin-northern: right! fixed everywhere, with login, password and idp_url.

tests/e2e_tests/fixtures/fixtures.ts
merlin-northern: fixed!

.gitlab-ci.yml
merlin-northern: aha! I knew they must be there -- forgot to check, sorry for that.

tests/e2e_tests/integration/07-saml.spec.ts
merlin-northern: thanks :) they really do help me :) I just like a good pic :)

tests/e2e_tests/integration/07-saml.spec.ts
merlin-northern: roger that. and I know it is in php :)

mzedel (approved):
👌 Thank you! Now just make sure to remove the adjustments to `.gitlab-ci.yml` once the staging tests ran through 😉

PR #3013, opened 2022-08-17
MEN-5770 - fix: removed effect cascade in order to fix an issue when opening a device as a RO user

mzedel:
Ticket: MEN-5770
Changelog: Title


mzedel:
the `device.id` debounce is already happening in the outer component (in `authorized-devices.js`), so this shouldn't be required...

tranchitella (approved):

PR #3019, opened 2022-08-18
fix: adjusted saml configuration limitation to enterprise

mzedel:
Ticket: None
Changelog: None


src/js/components/settings/organization/organization.js
```diff
@@ -134,21 +136,23 @@ export const Organization = ({
...
-        {isResettingSSO && !isConfiguringSSO && (
-          <>
-            <Button onClick={onCancelSSOSettings}>Cancel</Button>
-            <Button onClick={onSaveSSOSettings} disabled={!hasSingleSignOn} variant="contained">
-              Save
-            </Button>
-          </>
-        )}
-      </div>
+      {isEnterprise && (
```
tranchitella, line 139: One request from @alfrunes, would it be possible to hide it unless version is "next" similarly to what we have with the IoT Core integration?

tranchitella (approved):

PR #3027, opened 2022-08-24
fix: moved announcement section of header to its own row

michaelatmender:
Ticket: None
Changelog: None


mzedel (changes requested):
👍 that should work better, just the _Lorem ipsum_ needs to go

PR #3031, opened 2022-08-25
Ci/lint all the things

mzedel:


mzedel:
needs some more cleanup for the duplicate strings...

tests/e2e_tests/.eslintrc
```diff
@@ -0,0 +1,15 @@
+{
+  "parser": "@typescript-eslint/parser",
+  "extends": ["eslint:recommended", "plugin:@typescript-eslint/eslint-recommended", "plugin:@typescript-eslint/recommended"],
+  "rules": {
+    "prettier/prettier": 2,
```
merlin-northern, line 5: I heard of:

```js
"prettier/prettier": ["error"]
```

what does `2` do?

tests/e2e_tests/.eslintrc
```diff
@@ -0,0 +1,15 @@
...
+    "consistent-this": ["error", "self"],
+    "import/no-named-as-default": 0,
+    "no-console": 0,
+    "no-multiple-empty-lines": 2,
```
merlin-northern, line 9: can we be explicit here? suggested: `"max": 2` also: could we have `"maxEOF":1` and `"maxBOF":0` as I believe this are the max we use.

tests/e2e_tests/.eslintrc
```diff
@@ -0,0 +1,15 @@
...
+    "no-prototype-builtins": 0,
```
merlin-northern, line 10: dont you think we should have it?

tests/e2e_tests/.eslintrc
```diff
@@ -0,0 +1,15 @@
...
+    "quotes": [2, "single", { "allowTemplateLiterals": true }],
```
merlin-northern, line 11: quotes rule doe snot have three options, only two, what does `2` do here?

merlin-northern: do we consider `avoidEscape` ?

tests/e2e_tests/.eslintrc
```diff
@@ -0,0 +1,15 @@
...
+    "sonarjs/cognitive-complexity": 0,
```
merlin-northern, line 12: I think that we should have it. all the rest of the codebase has some sort of complexity restrictions. `["error", 8]` ?

tests/e2e_tests/.eslintrc
```diff
@@ -0,0 +1,15 @@
...
+    "sonarjs/no-nested-template-literals": 0
```
merlin-northern, line 13: I would try to have it in, in case you want to go crazy with the backticks.

tests/e2e_tests/integration/07-saml.spec.ts
```diff
@@ -64,13 +63,13 @@ test.describe('SAML Login', () => {
             rejectUnauthorized: false
           })
         });
-        expect(response.status).toBe(200);
+        expect(responseDelete.status).toBe(200);
```
merlin-northern, line 66: I went back and forth with it, and one of those times I just missed it. thanks!

tests/mockData.js
```diff
@@ -31,6 +31,24 @@ const testDate = new Date(TEST_SESSION_DATETIME);
 export const mockDate = new Date(testDate.setMilliseconds(testDate.getMilliseconds() + 900));
 
 export const defaultPassword = 'mysecretpassword!123';
+export const defaultCreationDate = '2019-01-01T06:25:00.000Z';
```
merlin-northern, line 34: sorry for being picky, 01-01 is not a good mock date, can we get a day number greater than 12?

tests/mockData.js
```diff
@@ -458,50 +476,50 @@ export const permissionSets = [
...
+      { action: 'http', object: { type: 'DELETE', value: commonEndpoints.artifactDetails } },
       { action: 'http', object: { type: 'GET', value: '^/api/management/v1/deployments/artifacts/[^/]+/download' } }
     ]
   },
   {
     name: defaultPermissionSets.ReadUsers.value,
-    action: 'Read',
-    object: 'User management',
+    action: uiPermissionsById.read.title,
+    object: permissionSetObjectTypes.userManagement,
```
merlin-northern, line 493: in a review process I actually went to `src/js/constants/userConstants.js` to check them.

tests/e2e_tests/.eslintrc
mzedel: "off" or 0 - turn the rule off
"warn" or 1 - turn the rule on as a warning (doesn’t affect exit code)
"error" or 2 - turn the rule on as an error (exit code is 1 when triggered)

merlin-northern: could we be explicit? it is not easy to find what the numbers mean.

tests/e2e_tests/.eslintrc
mzedel: that will be enforced from prettier already 👍

tests/e2e_tests/.eslintrc
mzedel: hmm nah, I find the readability of the alternative too bad to justify the rule 🙃

tests/e2e_tests/.eslintrc
mzedel: hmm looks like the number of options has changed in one of the previous version... we need escapes for some of the texts in the GUI...

tests/e2e_tests/.eslintrc
mzedel: we need this in e.g. the device connection snippet...

merlin-northern: I see. sorry for being annoying and challenging every line.

tests/mockData.js
mzedel: we could, but I'm quite sure that would cause quite a few more changes for assertions on rendered content etc. ...

tests/e2e_tests/.eslintrc
mzedel: ? https://eslint.org/docs/latest/user-guide/configuring/rules - this is an eslint config file after all...

tests/e2e_tests/.eslintrc
mzedel: that rule unfortunately doesn't really work with react components and is more targeted at plain JS - so no.

tests/e2e_tests/.eslintrc
mzedel: ah no, the options are counted after the severity 🤦 - so all good here 👌

tests/e2e_tests/.eslintrc
mzedel: on second thought, you're right - changed it in the outer config!

merlin-northern (approved):
thanks for answers.

PR #3035, opened 2022-08-29
chore: fixed password change input + let uploaded filename become release name

mzedel:


tranchitella (approved):
Tested locally and it works fine.

PR #3041, opened 2022-08-31
feat: added possibility to add extra confirmation step on deployment creation

mzedel:
for now this uses a 10 device threshold + an organization wide setting to determine if the request should be shown - the threshold could be replaced with a percentage or we could just rely on the organization setting...

@michaelatmender I changed the dialog approach to the confirmation overlay we use in other parts of the UI to try to avoid the dialog nested in the slide out, but would also be happy to change it to the dialog 👌
Screenshot 2022-08-31 at 10 27 15

### DIFF
--- a/src/js/components/common/confirm.js
+++ b/src/js/components/common/confirm.js
@@ -41,7 +41,7 @@ const confirmationType = {
   }
 };
 
-export const Confirm = ({ action, cancel, classes = '', style = {}, type }) => {
+export const Confirm = ({ action, cancel, classes = '', message = '', style = {}, type }) => {
   const [className, setClassName] = useState('fadeIn');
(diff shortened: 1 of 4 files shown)


michaelatmender (commented):
I think my only comment is whether we need the 10 device limit. Possibly better to just have the on/off switch apply to all deployments.

mzedel:
> I think my only comment is whether we need the 10 device limit. Possibly better to just have the on/off switch apply to all deployments.

Thanks for looking at this @michaelatmender! 👌 I removed the 10 device limit (rereading the ticket for this, I shouldn't have brought it in in the first place 🙈)

michaelatmender (approved):

PR #3082, opened 2022-09-02
MEN-5721 - webhook configuration

mzedel:
this should be very easy to extract into a separate settings page, but deviates a bit from the design [MEN-5716](https://tracker.mender.io/browse/MEN-5716) while waiting for the next iteration:
- no details for a certain webhook activity (only device ids are returned)
- no pagination on webhook activity (due to lack of pagination information - [MEN-5855](https://tracker.mender.io/browse/MEN-5855)) - the first 50 events get returned
- only supports having a single webhook/ integration (so no pagination there either)
- can't disable an integration (only remove + recreate)

still needs tests to be added...

### DIFF
--- a/src/js/components/common/detailstable.js
+++ b/src/js/components/common/detailstable.js
@@ -1,0 +1,29 @@
+import React from 'react';
+
+// material ui
+import { Table, TableBody, TableCell, TableHead, TableRow } from '@mui/material';
+
+export const DetailsTable = ({ columns, items, onItemClick }) => {
+  return (
+    <Table className="margin-bottom">
+      <TableHead>
+        <TableRow>
+          {columns.map(({ key, title }) => (
+            <TableCell key={key}>{title}</TableCell>
+          ))}
+        </TableRow>
+      </TableHead>
+      <TableBody>
+        {items.map((item, index) => (
+          <TableRow className="clickable" hover key={item.id || index} onClick={() => onItemClick(item)}>
+            {columns.map(column => (
+              <TableCell key={column.key}>{column.render(item, column.extras)}</TableCell>
+            ))}
+          </TableRow>
+        ))}
+      </TableBody>
+    </Table>
+  );
+};
+
+export default DetailsTable;
--- a/src/js/components/settings/user-management/userlist.js
+++ b/src/js/components/settings/user-management/userlist.js
@@ -1,72 +1,69 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 
 // material ui
-import { Chip, Table, TableBody, TableCell, TableHead, TableRow } from '@mui/material';
+import { Chip } from '@mui/material';
 import { ArrowRightAlt as ArrowRightAltIcon, Check as CheckIcon } from '@mui/icons-material';
 
 import Time, { RelativeTime } from '../../common/time';
 import { twoFAStates } from '../../../constants/userConstants';
+import DetailsTable from '../../common/detailstable';
 
 const columnData = [
-  { id: 'email', disablePadding: false, label: 'Email', enterpriseOnly: false },
-  { id: 'created_ts', disablePadding: false, label: 'Date created', enterpriseOnly: false },
-  { id: 'updated_ts', disablePadding: false, label: 'Last updated', enterpriseOnly: false },
-  { id: 'roles', disablePadding: false, label: 'Role', enterpriseOnly: true },
-  { id: 'actions', disablePadding: false, label: 'Manage', enterpriseOnly: false }
+  {
+    key: 'email',
+    disablePadding: false,
+    title: 'Email',
+    enterpriseOnly: false,
+    render: user => (
+      <>
+        <span>{user.email}</span>
+        {user.tfa_status === twoFAStates.enabled && (
+          <Chip className="margin-left-small" icon={<CheckIcon titleAccess={`2FA ${twoFAStates.enabled}`} />} label="2FA" size="small" variant="outlined" />
+        )}
+      </>
+    )
+  },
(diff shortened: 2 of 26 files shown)


src/js/components/settings/webhooks/configuration.js
```diff
@@ -0,0 +1,103 @@
...
+        </Button>
+      </div>
+    </>
+  );
+};
+
+export const WebhookCreation = ({ adding, onCancel, ...props }) => (
+  <Drawer anchor="right" open={adding} PaperProps={{ style: { minWidth: 600, width: '50vw' } }}>
+    <div className="flexbox center-aligned margin-bottom-small space-between">
+      <h3>Add a webhook</h3>
```
oleorhagen, line 93: Q: Why not use `Typography`?

src/js/components/settings/webhooks/webhooks.js
```diff
@@ -0,0 +1,173 @@
...
+  createIntegration,
+  deleteIntegration,
+  docsVersion,
+  events,
+  getIntegrations,
+  getWebhookEvents,
+  webhook = { ...emptyWebhook },
+  webhooks
+}) => {
+  const [adding, setAdding] = useState(false);
```
oleorhagen, line 45: Holy moly, a lot of bool state

tests/__mocks__/organizationHandlers.js
```diff
@@ -155,6 +155,19 @@ export const organizationHandlers = [
     }
     return res(ctx.status(200));
   }),
+  rest.get(`${iotManagerBaseURL}/events`, (req, res, ctx) => {
```
oleorhagen, line 158: Nit: `req` is never used

mzedel: yep, but I need to get to the 2nd and 3rd argument and there's no named arguments in JS... just object destructuring (if an object would be passed in here)

src/js/components/settings/webhooks/configuration.js
mzedel: while the Typography allows to more centrally manage the theming of anything typography I feel it dilutes the connection to the markup, since you can write `<Typography variant="h2" />` but define a different mapping to the proper html `<h2 />` in the theme… so you might end up writing `h2` but getting `h6` or `span`… if you know the entire codebase you would know, but if multiple people work on theme + components… the connection gets fuzzy too easily… 🤷‍♂️ not an ideal position for me to have I think…

src/js/components/settings/user-management/userlist.js
```diff
@@ -1,72 +1,69 @@
...
-            return accu;
-          }
-          accu.push(
-            <TableCell key={column.id} padding={column.disablePadding ? 'none' : 'normal'}>
-              {column.label}
-            </TableCell>
-          );
+const UserList = ({ editUser, isEnterprise, roles, users }) => {
+  const columns = useMemo(
+    () =>
```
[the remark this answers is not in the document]
mzedel: yep, since the shared table component can't really know what to optimize/ with what to limit rerenderings...

src/js/components/common/detailstable.js
```diff
@@ -0,0 +1,29 @@
...
+      <TableHead>
+        <TableRow>
+          {columns.map(({ key, title }) => (
+            <TableCell key={key}>{title}</TableCell>
+          ))}
+        </TableRow>
+      </TableHead>
+      <TableBody>
+        {items.map((item, index) => (
+          <TableRow className="clickable" hover key={item.id || index} onClick={() => onItemClick(item)}>
```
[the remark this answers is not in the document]
mzedel: `index` might not be the same for an object (say if the list order is changed), so you might end up not updating the correct list item...

tranchitella (changes requested):
It works fine, but I detected an issue which I believe we should fix. When editing the webhook, the UI tries to send back the secret as "<omitted>" which triggers a 400:

`malformed request body: value error: '\u003comitted\u003e' is not a hexadecimal string`

The API expects the secret together with the URL, therefore I believe the secret text field should not be type password and we should leave it to the user to manually remove the "<omitted>" and replace it with the correct value when editing the URL, or eventually leave it empty.

What do you think, @mzedel

mzedel:
Thanks for taking a look @tranchitella!

> The API expects the secret together with the URL, therefore I believe the secret text field should not be type password and we should leave it to the user to manually remove the `<omitted>` and replace it with the correct value when editing the URL, or eventually leave it empty.

The secret should already be disabled on edits and will now be omitted from any update requests to an existing webhook integration (+ same for the description)

> EDIT: as we are hiding AWS IoT Core behind the `next` version, we should do the same for the Webhooks.

This should now be the case 👌 

Also the UI will check for additional events if the number of displayed events is the page limit (and also show the pagination controls)...

tranchitella:
> The API expects the secret together with the URL, therefore I believe the secret text field should not be type password and we should leave it to the user to manually remove the `<omitted>` and replace it with the correct value when editing the URL, or eventually leave it empty.
> 
> What do you think, @mzedel?

This still doesn't work: if you don't set a secret, the backend removes the value.
I believe the right way (according to the current APIs implementation) is make the secret editable and visible, with `"<omitted>"` (coming from the backend) as default value.

tranchitella:
> > This still doesn't work: if you don't set a secret, the backend removes the value. I believe the right way (according to the current APIs implementation) is make the secret editable and visible, with `"<omitted>"` (coming from the backend) as default value.
> 
> @tranchitella I believe the current APIs implementation is wrong then... if a user were to update the URL without supplying a secret this shouldn't be overwritten... if a user were to supply e.g. an empty secret, this should be overwritten (although it will likely be rejected as non-hexadecimal)... mostly because the URL is independent of the secret... I'm open to make the secret editable at a later point, but editing URL only should be possible.

Noted, colleagues in the backend team will follow-up then. @alfrunes FYI.

tranchitella (approved):

PR #3089, opened 2022-09-09
Fix staging saml tests - next round

mzedel:


mzedel:
https://gitlab.com/Northern.tech/Mender/gui/-/pipelines/638900157 should end up green

tranchitella (approved):

PR #3092, opened 2022-09-13
MEN-5653 - fix: switched device connection dialog to delta based device addition check

mzedel:
Ticket: MEN-5653
Changelog: None


merlin-northern (approved):
I tested and both the MEN-5653 as well as the location refreshes are fixed.

PR #3110, opened 2022-09-14
SEC-579 - more login & added rbac related e2e tests

mzedel:


mzedel:
successful run on staging is [here](https://gitlab.com/Northern.tech/Mender/gui/-/pipelines/641597366)

tranchitella (approved):

PR #3132, opened 2022-09-16
Chore/improve e2e tests

mzedel:


mzedel:
this is both a follow up to the SAML url changes + adjustments to accommodate some slightly different selector behaviour + timings on webkit runs
a successful run against staging is [here](https://gitlab.com/Northern.tech/Mender/gui/-/pipelines/644172454)

tranchitella (approved):

PR #3140, opened 2022-09-20
feat: prevented long configuration values from unlimited growth

mzedel:
Ticket: None
Changelog: None

### DIFF
--- a/src/js/components/common/configurationobject.js
+++ b/src/js/components/common/configurationobject.js
@@ -16,18 +16,24 @@ const useStyles = makeStyles()(theme => ({
   }
 }));
 
-const ValueColumn = ({ value, setSnackbar }) => {
+const cutoffLength = 100;
+const ValueColumn = ({ value = '', setSnackbar }) => {
   const [tooltipVisible, setTooltipVisible] = useState(false);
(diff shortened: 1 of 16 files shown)


src/js/components/common/configurationobject.js
```diff
@@ -16,26 +16,32 @@ const useStyles = makeStyles()(theme => ({
...
+      if (isComponent) {
         copyable = value.props.value;
       }
       copy(copyable);
       setSnackbar('Value copied to clipboard');
     }
   };
+  let shownValue = value;
+  if (!isComponent) {
+    shownValue = <div title={value}>{value.length > cutoffLength ? `${value.substring(0, cutoffLength - 3)}...` : value}</div>;
```
oleorhagen, line 35: Q: Why the post `...` ? :thinking:

oleorhagen (approved):
:raised_back_of_hand:

src/js/components/common/configurationobject.js
mzedel: it's commonly treated as a sign for "more content" if it was left to CSS to handle this via overflow that would be the default "ellipsis" so `...`

PR #3146, opened 2022-09-22
fix: made onboarding completion notification more forgiving for users not following the intended flow

mzedel:
Ticket: None
Changelog: None


lluiscampos (approved):
My use case works fine now :clap:

PR #3153, opened 2022-09-22
MEN-5929 - fix: made device state error messages integration provider dependent

mzedel:
Ticket: MEN-5929
Changelog: None

### DIFF
--- a/src/js/actions/deviceActions.js
+++ b/src/js/actions/deviceActions.js
@@ -11,7 +11,7 @@ import AppConstants from '../constants/appConstants';
 import DeviceConstants from '../constants/deviceConstants';
 
 import { attributeDuplicateFilter, deepCompare, extractErrorMessage, getSnackbarMessage, mapDeviceAttributes } from '../helpers';
-import { getIdAttribute, getTenantCapabilities, getUserCapabilities, getUserSettings } from '../selectors';
+import { getDeviceTwinIntegrations, getIdAttribute, getTenantCapabilities, getUserCapabilities, getUserSettings } from '../selectors';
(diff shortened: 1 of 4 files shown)


src/js/actions/deviceActions.js
```diff
@@ -1014,14 +1014,18 @@ export const setDeviceTags = (deviceId, tags) => dispatch =>
       .then(() => Promise.resolve(dispatch({ type: DeviceConstants.RECEIVE_DEVICE, device: { ...device, tags } })));
   });
 
-export const getDeviceTwin = deviceId => (dispatch, getState) => {
+export const getDeviceTwin = (deviceId, integration) => (dispatch, getState) => {
```
lluiscampos, line 1017: Should the call at [line 487](https://github.com/mendersoftware/gui/pull/3153/files#diff-74f29c7fa70ecbddf5cd2b41184645057374578e6b26e9a4ff4ff49d3cabb157R487) (and the signature for `getDeviceInfo`, and then the callers too) be updated?

mzedel: Thank you very much for catching line 487!! and regarding the signature https://northern-tech.slack.com/archives/C0162LSV0A2/p1663861094911339?thread_ts=1663855342.895419&cid=C0162LSV0A2 - to me device twin (or more broadly digital twin) are common enough that I wouldn't let this be taken by any one company...

lluiscampos: I didn't mean the name but the fact that (I assumed) `getDeviceInfo` would need then to take a new parameter `integration` -> in that sense of a signature change. Anyway, your rework showed the limits of my "intuitive javascript" coding level.

lluiscampos (approved):

PR #3161, opened 2022-09-30
fix: fixed links to debian packages with a different package name.

mzedel:
Ticket: None
Changelog: None

### DIFF
--- a/src/js/components/help/__snapshots__/help.test.js.snap
+++ b/src/js/components/help/__snapshots__/help.test.js.snap
@@ -744,7 +744,7 @@ exports[`Help Component static components renders Downloads correctly 1`] = `
   white-space: nowrap;
 }
 
-.emotion-107 {
+.emotion-75 {
   -webkit-user-select: none;
   -moz-user-select: none;
   -ms-user-select: none;
@@ -761,7 +761,7 @@ exports[`Help Component static components renders Downloads correctly 1`] = `
   font-size: 1.25rem;
 }
 
-.emotion-123 {
+.emotion-91 {
   display: -webkit-inline-box;
   display: -webkit-inline-flex;
   display: -ms-inline-flexbox;
@@ -832,28 +832,28 @@ exports[`Help Component static components renders Downloads correctly 1`] = `
   cursor: pointer;
 }
 
-.emotion-123::-moz-focus-inner {
+.emotion-91::-moz-focus-inner {
   border-style: none;
 }
 
-.emotion-123.Mui-disabled {
+.emotion-91.Mui-disabled {
   pointer-events: none;
   cursor: default;
 }
 
 @media print {
-  .emotion-123 {
+  .emotion-91 {
     -webkit-print-color-adjust: exact;
     color-adjust: exact;
   }
 }
 
-.emotion-123.Mui-disabled {
+.emotion-91.Mui-disabled {
   opacity: 0.38;
   pointer-events: none;
 }
 
-.emotion-123 .MuiChip-avatar {
+.emotion-91 .MuiChip-avatar {
   margin-left: 5px;
   margin-right: -6px;
   width: 24px;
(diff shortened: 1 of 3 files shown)


src/js/components/help/downloads.js
```diff
@@ -51,7 +51,9 @@ const defaultLocationFormatter = ({ os, tool, versionInfo }) => {
 const osArchLocationReducer = ({ archList, location = downloadLocations.public, packageId, id, osList, versionInfo }) =>
   osList.reduce((accu, os) => {
     const osArchitectureLocations = archList.map(arch => ({
-      location: `${location}/repos/debian/pool/main/${id[0]}/${id}/${encodeURIComponent(`${packageId}_${getVersion(versionInfo, id)}-1+${os}_${arch}.deb`)}`,
+      location: `${location}/repos/debian/pool/main/${id[0]}/${packageId || id}/${encodeURIComponent(
+        `${packageId}_${getVersion(versionInfo, id)}-1+${os}_${arch}.deb`
+      )}`,
```
lluiscampos, line 56: Sorry, this is a bit more tricky than anticipated (after all, we are reverse-engineering the APT repositories pool). Two things still missing:
* one "recipe" can have multiple "packages". For example `mender-client` recipe produces `mender-client` and `mender-client-dev`, but both end up in the same recipe directory. So basically URL for the `-dev` one will be: `https://downloads.mender.io/repos/debian/pool/main/m/mender-client/mender-client-dev_3.3.0...`
* on top of that, some packages are architecture agnostic and need to have `_all` suffix. This applies to `mender-client-dev`: `.../main/m/mender-client/mender-client-dev_3.3.0-1%2Bdebian%2Bbuster_all.deb`

src/js/components/help/__snapshots__/help.test.js.snap
```diff
@@ -1798,7 +1798,7 @@ exports[`Help Component static components renders Downloads correctly 1`] = `
                     class="MuiSvgIcon-root MuiSvgIcon-fontSizeMedium MuiChip-deleteIcon MuiChip-deleteIconMedium MuiChip-deleteIconColorDefault MuiChip-deleteIconOutlinedColorDefault emotion-6"
                     data-testid="ArrowDropDownIcon"
                     focusable="false"
-                    value="https://downloads.mender.io/repos/debian/pool/main/m/mender/mender-client-dev_3.3.0-1%2Bubuntu%2Bfocal_amd64.deb"
+                    value="https://downloads.mender.io/repos/debian/pool/main/m/mender-client-dev/mender-client-dev_3.3.0-1%2Bubuntu%2Bfocal_amd64.deb"
```
lluiscampos, line 1801: After the fixes pointed in the previous comment, these links should look like:
```suggestion
                    value="https://downloads.mender.io/repos/debian/pool/main/m/mender-client/mender-client-dev_3.3.0-1%2Bubuntu%2Bfocal_all.deb"
```

src/js/components/help/__snapshots__/help.test.js.snap
```diff
@@ -1798,7 +1798,7 @@ exports[`Help Component static components renders Downloads correctly 1`] = `
...
                     viewBox="0 0 24 24"
```
lluiscampos, line 1802: Could we have a similar test for `mender-configure-module` links? This can also be a tricky component to get the links right so it would be good to have a test for it.

lluiscampos (changes requested):
See comments below.

src/js/components/help/downloads.js
mzedel: the `all` suffix is already used for monitor & configure, but I was assuming the architecture would be the same for all the packages... would you think it's only limited to the `-dev` package that this diverges to `all` or would you expect this to happen more often?

src/js/components/help/__snapshots__/help.test.js.snap
mzedel: => https://github.com/mendersoftware/gui/pull/3161/commits/9a2b27ef9ac5ab92b74f669e84190c1ee2ccf4c7#diff-8386ac26f9a9fce64463470bb86eac18f22b4c80f450109de912be9a7b5182ffR6-R9 👌

src/js/components/help/downloads.js
lluiscampos: As a rule of thumb we should not assume that all packages from one recipe share architecture dependencies. The `-dev` here is a coincidence: `mender-client`(recipe) produces a package with binaries (iow architecture dependent) and another with dev files (architecture independent). But instead of `-dev` could be an extra package with examples, or documentation, or something else that is architecture independent.

In mender-configure we have example packages that are architecture independent - in this case the main package is also architecture independent but it could very well not be the case.

lluiscampos (approved):
To the extend I can review it, looks good now! Nicely done :+1:

PR #3194, opened 2022-10-05
chore: updated webkit terminal reference to align with new xterm version

mzedel:


oleorhagen (approved):
interesting update :p

PR #3199, opened 2022-10-07
fix: fixed an issue that prevented config of users using SSO to log in

mzedel:
- this also aligns the SSO check on user config with the updated sso info from the api

Ticket: None
Changelog: Title

### DIFF
--- a/src/js/components/settings/user-management/userdefinition.js
+++ b/src/js/components/settings/user-management/userdefinition.js
@@ -9,7 +9,7 @@ import { makeStyles } from 'tss-react/mui';
 import { UserRolesSelect } from './userform';
 import { uiPermissionsById, uiPermissionsByArea } from '../../../constants/userConstants';
 import { TwoColumnData } from '../../common/configurationobject';
-import { OAuth2Providers } from '../../login/oauth2providers';
+import { genericProvider, OAuth2Providers } from '../../login/oauth2providers';
 import { mapUserRolesToUiPermissions } from '../../../actions/userActions';
 
 const useStyles = makeStyles()(theme => ({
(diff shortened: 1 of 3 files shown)


mzedel:
FYI @lluiscampos

src/js/components/settings/user-management/userdefinition.js
```diff
@@ -20,8 +20,15 @@ const useStyles = makeStyles()(theme => ({
   widthLimit: { maxWidth: 500 }
 }));
 
+export const getUserSSOState = user => {
+  const { sso = [] } = user;
+  const isOAuth2 = !!sso.length;
+  const provider = isOAuth2 ? OAuth2Providers.find(provider => sso.some(({ kind }) => kind.includes(provider.id))) ?? genericProvider : null;
+  return { isOAuth2, provider };
+};
+
```
merlin-northern, line 29: just out of curiosity, how are you dealing with unit tests? does this function have units? also: usually I would expect some test being enhanced to catch this case.

merlin-northern (approved):
lgtm. however I was not able to test it in action.

src/js/components/settings/user-management/userdefinition.js
mzedel: https://northern.tech.gitlab.io/-/Mender/gui/-/jobs/3141899067/artifacts/coverage/lcov-report/js/components/settings/user-management/userdefinition.js.html#L23 <- this function seems to be called 40 times during the unit tests (since it's used in 2 components already)... although the fallback to the `genericProvider` is not caught by itself 😕, but I'll give it a go once it's on staging 😉

mzedel:
@mender-test-bot cherry-pick to:
 * 3.4.x

PR #3204, opened 2022-10-11
fix(e2e-tests): unify client image tags to avoid inconsistencies

mzedel:
Ticket: None
Changelog: None


oleorhagen:
wrong, unfortunately

PR #3210, opened 2022-10-14
MEN-5744 - feat: made it possible to upload multiple files in parallel

mzedel:
+ allowed aborting them individually

Ticket: MEN-5744
Changelog: Title
Signed-off-by: Manuel Zedel <manuel.zedel@northern.tech>

PR #3217, opened 2022-10-18
fix: fixed an issue that would prevent uploads to devices

mzedel:
Ticket: None
Changelog: None


mzedel:
this fixes a regression introduced with #3210 (since there was no unit test for the device file upload 😱) + it was caught by the e2e tests 🎉

tranchitella (approved):

PR #3219, opened 2022-10-20
MEN-5875

mzedel:


mzedel:
this results in the overview of:
Screenshot 2022-10-21 at 09 27 23

for the use case from [Mender Hub](https://hub.mender.io/t/design-proposal-uefi-capsule-support/5278/11)
Screenshot 2022-10-21 at 09 25 50

and when going all overboard with a list of software:
Screenshot 2022-10-21 at 09 26 05

src/js/helpers.js
```diff
@@ -469,13 +469,24 @@ export const getSnackbarMessage = (skipped, done) => {
...
+  return Object.entries(attributes).reduce(
+    (accu, item) => {
+      if (softwareKeys.some(key => item[0].startsWith(key))) {
+        accu.software.push(item);
+      } else {
+        accu.nonSoftware.push(item);
+      }
+      return accu;
+    },
+    { software: [], nonSoftware: [] }
```
oleorhagen, line 488: Nice!

oleorhagen (approved):
Took the liberty to have a quick look :)

Looks nice. Last commit was a mouthful too much this morn.

I have a question about the UX though. Isn't the levels sort of hard to discern from eachother?

kacf:
Haha, I love the "overboard" case!

Looks really nice to me, exactly what I had in mind for the grouping.

One visual issue that could perhaps be improved, though I'm unsure what would be the best method: It's a little bit hard to see that for example `directory` is under `System filesystem`, but `uefi-firmware` isn't. Part of it is because the indentation is very light, but I'm not necessarily convinced that increasing it is the best way, because that will make it uglier. But maybe there could be a dividing line, or a box or something, just to make it a bit clearer visually?

In any case, you have my approval of the UI, but I'm not using the Approve button since I haven't reviewed the code. 🙂

mzedel:
@michaelatmender the last push makes this now look like:
Screenshot 2022-10-21 at 16 27 23

tranchitella (approved):

PR #3224, opened 2022-10-31
MEN-5918, MEN-5919, MEN-5920

mzedel:


tranchitella:
This is how the dashboard looks like this:

image

At the bottom of the page, the box "Create a new deployment..." doesn't have top margin and looks weird.
Additionally, I do have a pending deployment but the dashboard doesn't get it until I click on "Deployments", then back to the dashboard. At this point, the dashboard looks like this:

image

Once the deployment is completed, the dashboard looks like this:

image

At this point it is good, though I'd still add a bit of margin-bottom to the dashboard as "See all deployments" is too attached to the bottom border of the browser's window.

A final comment: I don't get why there is a lot of white space between the first row of widgets and the software distribution ones.

tranchitella (approved):
I tested it again and I couldn't spot any issues.
I also skimmed through the code, however I didn't read every single line, and it makes sense.
Green light from me, thanks.

mzedel:
@michaelatmender I merged this to staging, so it should be easier to test it & I'll address any issues you come across in a separate PR asap - I hope that's ok!

michaelatmender:
> @michaelatmender I merged this to staging, so it should be easier to test it & I'll address any issues you come across in a separate PR asap - I hope that's ok!

Yes, that's great - thank you!

PR #3275, opened 2022-11-16
QA-493: ci: Update jobs tags for new runners

lluiscampos:


lluiscampos:
> no, we need the CPUs there.

Sure, let's discuss it and define a tag for your requirements.

Do not just close this PR because I am going to deprecate the  `slave` kind of tags.

mzedel:
> Do not just close this PR because I am going to deprecate the `slave` kind of tags.

last info I had on this was that they would be kept for backwards compatibility... 🤷‍♂️ - the PR as is would break the GUI pipelines.

lluiscampos:
Sorry if that was miss-leading. I will keep the old tags for backwards compatibility while maintain releases that have these in the pipelines. But eventually (whenever 3.3.x LTS goes end of life) I will remove them.

I can offer a shinny AMD machines for you :money_with_wings: Let me run some tests with 8 and 16 cores and see how the run times go before deciding.

> the PR as is would break the GUI pipelines.

You mean the release branches or something else that I am missing? What exactly would break?

mzedel:
> You mean the release branches or something else that I am missing? What exactly would break?

every pipeline that runs unit tests (so master + the release branches this lands in)

mzedel (approved):

PR #3279, opened 2022-11-18
chore: add Mender logo to repo and fix image in README file

michaelatmender:
Changelog: None
Ticket: None


mzedel (approved):
ok for me, but instead we could also reuse the svg from `src/assets/img/loginlogo.svg` (so `Mender logo and set a width there in the svg e.g. `<svg viewBox="0 0 145 36" width="200"  xmlns="http://www.w3.org/2000/svg">` - since the png looks slightly blurred on my machine...

mzedel (approved):

PR #3282, opened 2022-11-24
MEN-6061 - fix: reverted react-idle-timer to known working version to try fix unexpected log out issue

mzedel:
Ticket: MEN-6061
Changelog: Title


tranchitella (approved):
Honestly, I don't know if this really fixes the issue but it is worth a try.

mzedel:
> Honestly, I don't know if this really fixes the issue but it is worth a try.

same here... I base this purely on [the issue in the upstream repo](https://github.com/SupremeTechnopriest/react-idle-timer/issues/293)... so far I unfortunately have no luck reproducing it

PR #3284, opened 2022-11-29
MEN-6034 - tabbed device details

mzedel:


tranchitella:
@mzedel there is a strange label "Device actions" here:
image

Why?

tranchitella:
Below Monitoring and Troubleshooting's title there is a paragraph starting with the info icon.
These paragraphs are not aligned with each other, and when moving between the tabs you can detect it.
I know, this is a tiny issue, but still...

mzedel:
Thanks a lot for checking so quick @tranchitella! I'll adjust the alignment issue right away, but:

> @mzedel there is a strange label "Device actions" 

that should be the title of the floating action button... that _should_ be on top of the slide out, right next to the label 🧐

tranchitella:
@mzedel alignments are good, but I don't see how this label should be positioned on top of the floating menu.

image

and...

image

mzedel:
@tranchitella the screenshot pointed me in the right direction... Chrome 108 seems to render this differently - I had 107 and it looked as intended, with 108 there seems to be something changed... I'm looking into it!

mzedel:
@tranchitella now it should work 🤞

tranchitella (approved):
Confirmed, it works fine... and the tabs are lovely.
EDIT: I retested after the force-push, but couldn't spot any difference...

PR #3316, opened 2022-12-06
MEN-5847 - allow delta artifact generation + deployments configuration

mzedel:


tranchitella:
A few observations:
- In OS, we have the banner about "Automatic delta artifacts generation is not enabled in your account". Shouldn't we simply omit the banner and hide the feature?
- in ENT, the banner correctly shows up; however, once I set up the configuration with the following API call:
```
curl -X PUT http://mender-deployments:8080/api/internal/v1/deployments/tenants/TENANT_ID/config -d '{"delta": {"enabled": true, "binary_delta_limits": {"jobs_in_parallel": {"max": 2}, "queue_length": {"max": 4}}, "binary_delta": {"timeout": 3600}}}' -H "Content-Type: application/json"
```
... the UI continues to show the banner, and it automatically saves the configuration (?!?). I see in the backend/database that I know have values for settings for values I didn't include in the curl call above:
```
{
	"_id" : ObjectId("63a3ec00750a0af98d711671"),
	"delta" : {
		"enabled" : true,
		"binary_delta" : {
			"timeout" : 0,
			"xdelta_args" : {
				"disable_checksum" : false,
				"disable_external_decompression" : false,
				"compression_level" : 6,
				"source_window_size" : 0,
				"input_window_size" : 0,
				"compression_duplicates_window" : 0,
				"instruction_buffer_size" : 0
			}
		},
		"binary_delta_limits" : {
			"jobs_in_parallel" : {
				"max" : 2
			},
			"queue_length" : {
				"max" : 4
			}
		}
	},
	"etag" : "b6126661-a446-41a4-8e22-fad2a42f6153"
}
```

However, no matter what I did, the interface never replaced the banner with the form.

tranchitella:
With this diff, things are slightly better and the form shows up:
```
diff --git a/src/js/actions/deploymentActions.js b/src/js/actions/deploymentActions.js
index 0d399e787..d95f941f3 100644
--- a/src/js/actions/deploymentActions.js
+++ b/src/js/actions/deploymentActions.js
@@ -292,21 +292,21 @@ export const getDeploymentsConfig = () => (dispatch, getState) =>
       binaryDelta: {
         ...oldConfig.binaryDelta,
         timeout: binary_delta.timeout,
-        duplicatesWindow: binary_delta.xdelta_args.compression_duplicates_window,
-        compression: binary_delta.xdelta_args.compression_level,
-        disableChecksum: binary_delta.xdelta_args.disable_checksum,
-        disableDecompression: binary_delta.xdelta_args.disable_external_decompression,
-        inputWindow: binary_delta.xdelta_args.input_window_size,
-        instructionBuffer: binary_delta.xdelta_args.instruction_buffer_size,
-        sourceWindow: binary_delta.xdelta_args.source_window_size
+        duplicatesWindow: binary_delta.xdelta_args?.compression_duplicates_window,
+        compression: binary_delta.xdelta_args?.compression_level,
+        disableChecksum: binary_delta.xdelta_args?.disable_checksum,
+        disableDecompression: binary_delta.xdelta_args?.disable_external_decompression,
+        inputWindow: binary_delta.xdelta_args?.input_window_size,
+        instructionBuffer: binary_delta.xdelta_args?.instruction_buffer_size,
+        sourceWindow: binary_delta.xdelta_args?.source_window_size
       },
       binaryDeltaLimits: {
         ...oldConfig.binaryDeltaLimits,
         timeout: binary_delta_limits.timeout ?? limitDefault,
-        sourceWindow: binary_delta_limits.xdelta_args_limits.source_window_size ?? limitDefault,
-        inputWindow: binary_delta_limits.xdelta_args_limits.input_window_size ?? limitDefault,
-        duplicatesWindow: binary_delta_limits.xdelta_args_limits.compression_duplicates_window ?? limitDefault,
-        instructionBuffer: binary_delta_limits.xdelta_args_limits.instruction_buffer_size ?? limitDefault
+        sourceWindow: binary_delta_limits.xdelta_args_limits?.source_window_size ?? limitDefault,
+        inputWindow: binary_delta_limits.xdelta_args_limits?.input_window_size ?? limitDefault,
+        duplicatesWindow: binary_delta_limits.xdelta_args_limits?.compression_duplicates_window ?? limitDefault,
+        instructionBuffer: binary_delta_limits.xdelta_args_limits?.instruction_buffer_size ?? limitDefault
       }
     };
     return Promise.resolve(dispatch({ type: DeploymentConstants.SET_DEPLOYMENTS_CONFIG, config }));
diff --git a/src/js/components/settings/artifactgeneration.js b/src/js/components/settings/artifactgeneration.js
index 557670498..8804da821 100644
--- a/src/js/components/settings/artifactgeneration.js
+++ b/src/js/components/settings/artifactgeneration.js
@@ -171,10 +171,10 @@ export const ArtifactGenerationSettings = ({ deltaConfig, deltaEnabled, deltaLim
 const actionCreators = { getDeploymentsConfig, saveDeltaDeploymentsConfig };

-  const { binaryDelta = {}, binaryDeltaLimits = {}, enabled } = state.deployments.config ?? {};
+  const { binaryDelta = {}, binaryDeltaLimits = {}, hasDelta } = state.deployments.config ?? {};
   return {
     deltaConfig: binaryDelta,
-    deltaEnabled: enabled,
+    deltaEnabled: hasDelta,
     deltaLimits: binaryDeltaLimits
   };
 };
```

However:
- the form saves automatically when you open it, updating the settings in the db
- if some values are missing (e.g. Source window size), the UI sets those value to zero, both in the interface and in the DB; we should keep them "null" if no value is selected, and not provide them to the backend as configuration to skip those options when generating the binary delta
- All values default to 10, which is strange (and again, they should default to null)
- The label for Compression level overlaps the field; only that one, the other fields are okay.
image

tranchitella (changes requested):
See my comments.

tranchitella (commented):
It looks good, but I still have two issues:
- if the backend returns an empty configuration object, the UI still tries to auto-save the settings, showing a tooltip with the 404 not found error (it needs a fix: https://github.com/mendersoftware/deployments-enterprise/pull/747, without it it returns 500)
- Entering "10" in the numeric fields is tricky, because when you enter "1" using Chrome it automatically replaces it with "5" (the min value).

mzedel:
> * if the backend returns an empty configuration object, the UI still tries to auto-save the settings

that is an issue with the backend. the configuration the backend returns is a lie, as in the boolean values can not have a nullish value... the options we have there are to either fix the backend to return the values + defaults as they are used in the backend, or omit all fields in the ui that are not initialized in the backend with a corresponding value...

> Entering "10" in the numeric fields is tricky, because when you enter "1" using Chrome it automatically replaces it with "5" (the min value).

on it!

tranchitella:
@mzedel still experiencing an auto-save when the delta generation is disabled despite the default config:
https://github.com/mendersoftware/deployments-enterprise/pull/749

tranchitella:
> curl -X PUT http://mender-deployments:8080/api/internal/v1/deployments/tenants/TENANT_ID/config -d '{"delta": {"enabled": true, "binary_delta_limits": {"jobs_in_parallel": {"max": 2}, "queue_length": {"max": 4}}, "binary_delta": {"timeout": 3600}}}' -H "Content-Type: application/json"



> * Entering "10" in the numeric fields is tricky, because when you enter "1" using Chrome it automatically replaces it with "5" (the min value).

This one is solved, it works fine now!

tranchitella:
@mzedel it works fine when I set the parameters the first time. However, when I try editing them, the screen reports the wrong defaults (10 for all of them) and does not show the current values.

image

mzedel:
> the screen reports the wrong defaults (10 for all of them) 

if you don't like that default the backend has to send one (basically fix https://github.com/mendersoftware/deployments-enterprise/pull/749), until then the UI will always keep using the ones defined in the UI - the alternative that I'll have to switch to is that no input will be shown for values that are not supported (set) in the backend response... 🤷‍♂️

> does not show the current values.

will look into that when I'll align this to the properly reported backend limits...

tranchitella:
> @tranchitella to better explain the problems with the backend implementation I added [4f294ad](https://github.com/mendersoftware/gui/commit/4f294ad80db8599aaf47e4e5262d7073346dadd4) - there might be value in properly verifying backend responses in the long run based on the api spec to prevent these things from happening...

What about this? https://github.com/mendersoftware/deployments-enterprise/pull/755

mzedel:
> What about this? [mendersoftware/deployments-enterprise#755](https://github.com/mendersoftware/deployments-enterprise/pull/755)

Thanks @tranchitella, better, but unfortunately still lacks the boolean settings - as an API consumer knowing that these exist, there is no way to interact with the functionality without an assumed value, especially with a UI connected to it... The UI will _always_ have to assume the shown checkbox setting is a deliberate choice by the user even if they haven't touched it. As such I _think_ the api specs are faulty, as the boolean settings should be required and thus also returned on config retrieval...

mzedel:
> What about this? [mendersoftware/deployments-enterprise#755](https://github.com/mendersoftware/deployments-enterprise/pull/755)

Thanks @tranchitella, better, but unfortunately still lacks the boolean settings - as an API consumer knowing that these exist, there is no way to interact with the functionality without an assumed value, especially with a UI connected to it... The UI will _always_ have to assume the shown checkbox setting is a deliberate choice by the user even if they haven't touched it. As such I _think_ the api specs are faulty, as the boolean settings should be required and thus also returned on config retrieval...

tranchitella:
@mzedel please see my commit

tranchitella (approved):
Looks good, thank you!

PR #3332, opened 2022-12-19
MEN-6035 - device systems

mzedel:


tranchitella:
With this PR, I constantly get the scrambled columns:

image

tranchitella:
I have some doubts about the implementation, mainly because of lack of details in the tickets description and acceptance criteria. Let's start from the concept that a device is identified as gateway if the inventory attribute `mender_is_gateway` is set. The initial idea was to leverage something similar to aggregate the devices together, for example having an inventory attribute `mender_system_id` with a common value for all the devices belonging to a system, including the gateway. At this point, the UI could filter by `mender_system_id=value`, where `value` is the value set for the gateway. Similarly, when expanding a device belonging to a system, given the presence of `mender_system_id` we know it belongs to a System and we can filter by `mender_system_id=value && mender_is_gateway=true` to find out the corresponding gateway and link to it.

Does it make sense?

mzedel:
> Does it make sense?

yes, afaik the values we have in other parts of the system are: `mender_gateway=<gatewayDeviceId>` for the link to the gateway system & `mender_is_gateway=true` to identify gateway devices

tranchitella:
> > Does it make sense?
> 
> yes, afaik the values we have in other parts of the system are: `mender_gateway=<gatewayDeviceId>` for the link to the gateway system & `mender_is_gateway=true` to identify gateway devices

The problem I have with `mender_gateway=<gatewayDeviceId>` is that the device ID is not easy to obtain on the client side. I know it can be obtained from the JWT token, but if you think about a scenario where you want to configure the gateway and the system devices before the authorization, it can be tricky.

mzedel:
> > > Does it make sense?
> > 
> > 
> > yes, afaik the values we have in other parts of the system are: `mender_gateway=<gatewayDeviceId>` for the link to the gateway system & `mender_is_gateway=true` to identify gateway devices
> 
> The problem I have with `mender_gateway=<gatewayDeviceId>` is that the device ID is not easy to obtain on the client side. I know it can be obtained from the JWT token, but if you think about a scenario where you want to configure the gateway and the system devices before the authorization, it can be tricky.

Then I'll leave the PR as is until that is clear... 🤷‍♂️

tranchitella:
@mzedel my proposal:
* https://github.com/mendersoftware/mender-gateway/pull/77
* https://github.com/mendersoftware/mender-docs/pull/1985

src/js/actions/deviceActions.js
```diff
@@ -1061,3 +1059,74 @@ export const setDeviceTwin = (deviceId, integration, settings) => (dispatch, get
...
+  };
+
+export const getGatewayDevices = deviceId => (dispatch, getState) => {
+  const state = getState();
+  let device = state.devices.byId[deviceId];
+  const { attributes = {} } = device;
+  const { mender_gateway_system_id } = attributes;
+  const filters = [
+    { ...emptyFilter, key: 'mender_is_gateway', value: true, scope: 'system' },
+    { ...emptyFilter, key: 'mender_gateway_system_id', value: mender_gateway_system_id, scope: 'system' }
```
tranchitella, line 1117: Scope is `inventory`, it cannot be `system`. At the same time, the `mender_is_gateway` value is a string, not a boolean; the reason is that inventory attributes only support strings.

```suggestion
    { ...emptyFilter, key: 'mender_is_gateway', value: 'true', scope: 'inventory' },
    { ...emptyFilter, key: 'mender_gateway_system_id', value: mender_gateway_system_id, scope: 'inventory' }
```

kacf:
@mzedel: I think I need a bit more context to answer. Let's talk f2f maybe? (Or maybe one of the others can answer better than me)

oleorhagen:
> @kacf, @lluiscampos, @oleorhagen the ping for a review is pretty much just for [this](https://github.com/mendersoftware/gui/blob/db2ac70899d3cc3d4901757dba9c3d35daa3de1a/src/js/components/devices/dialogs/__snapshots__/connecttogatewaydialog.test.js.snap#L455-L457) part, since we need to provide the user with an easy way to connect additional devices to the gateway, but I'm not sure if the `--demo` flags in there won't defeat the purpose...

I'm not sure I understand the question. Like if it'll work with the demo flag?

mzedel:
> > @kacf, @lluiscampos, @oleorhagen the ping for a review is pretty much just for [this](https://github.com/mendersoftware/gui/blob/db2ac70899d3cc3d4901757dba9c3d35daa3de1a/src/js/components/devices/dialogs/__snapshots__/connecttogatewaydialog.test.js.snap#L455-L457) part, since we need to provide the user with an easy way to connect additional devices to the gateway, but I'm not sure if the `--demo` flags in there won't defeat the purpose...
> 
> I'm not sure I understand the question. Like if it'll work with the demo flag?

yep, basically the command in the link will show up once you would click on this button to [connect more devices to a gateway](https://www.figma.com/file/toVZN96xho52InU7VdvDRd/Device-systems-(via-gateway-devices)?node-id=49%3A130&t=dvbkxINf7sICltlo-0) - the device would be pointed to by the `--server-ip`... - so maybe one of the demo flags causes issues there... or there is a more appropriate way to connect devices to a gateway 🤷‍♂️

oleorhagen:
> > > @kacf, @lluiscampos, @oleorhagen the ping for a review is pretty much just for [this](https://github.com/mendersoftware/gui/blob/db2ac70899d3cc3d4901757dba9c3d35daa3de1a/src/js/components/devices/dialogs/__snapshots__/connecttogatewaydialog.test.js.snap#L455-L457) part, since we need to provide the user with an easy way to connect additional devices to the gateway, but I'm not sure if the `--demo` flags in there won't defeat the purpose...
> > 
> > 
> > I'm not sure I understand the question. Like if it'll work with the demo flag?
> 
> yep, basically the command in the link will show up once you would click on this button to [connect more devices to a gateway](https://www.figma.com/file/toVZN96xho52InU7VdvDRd/Device-systems-(via-gateway-devices)?node-id=49%3A130&t=dvbkxINf7sICltlo-0) - the device would be pointed to by the `--server-ip`... - so maybe one of the demo flags causes issues there... or there is a more appropriate way to connect devices to a gateway man_shrugging

Hmm, the first flag causes `demo` components to be downloaded, which I'm thinking is fine, and the second one is .. Hmm, not sure. But what is the context here. Is this a link which the user will copy and run on his device?

mzedel:
> > > > @kacf, @lluiscampos, @oleorhagen the ping for a review is pretty much just for [this](https://github.com/mendersoftware/gui/blob/db2ac70899d3cc3d4901757dba9c3d35daa3de1a/src/js/components/devices/dialogs/__snapshots__/connecttogatewaydialog.test.js.snap#L455-L457) part, since we need to provide the user with an easy way to connect additional devices to the gateway, but I'm not sure if the `--demo` flags in there won't defeat the purpose...
> > > 
> > > 
> > > I'm not sure I understand the question. Like if it'll work with the demo flag?
> > 
> > 
> > yep, basically the command in the link will show up once you would click on this button to [connect more devices to a gateway](https://www.figma.com/file/toVZN96xho52InU7VdvDRd/Device-systems-(via-gateway-devices)?node-id=49%3A130&t=dvbkxINf7sICltlo-0) - the device would be pointed to by the `--server-ip`... - so maybe one of the demo flags causes issues there... or there is a more appropriate way to connect devices to a gateway man_shrugging
> 
> Hmm, the first flag causes `demo` components to be downloaded, which I'm thinking is fine, and the second one is .. Hmm, not sure. But what is the context here. Is this a link which the user will copy and run on his device?

yep...
here you can see the button for the user to click highlighted: 
https://www.figma.com/file/toVZN96xho52InU7VdvDRd/Device-systems-(via-gateway-devices)?node-id=49%3A130&t=dvbkxINf7sICltlo-0
and here you can see the dialog this command is going to be shown in so that a user can copy it:
https://www.figma.com/file/toVZN96xho52InU7VdvDRd/Device-systems-(via-gateway-devices)?node-id=53%3A377&t=7Zm9dDgonQLtccLs-0

lluiscampos:
> @kacf, @lluiscampos, @oleorhagen the ping for a review is pretty much just for [this](https://github.com/mendersoftware/gui/blob/db2ac70899d3cc3d4901757dba9c3d35daa3de1a/src/js/components/devices/dialogs/__snapshots__/connecttogatewaydialog.test.js.snap#L455-L457) part, since we need to provide the user with an easy way to connect additional devices to the gateway, but I'm not sure if the `--demo` flags in there won't defeat the purpose...

Assuming
* That the device running `mender-gateway` was installed using `--demo` flag (as it is instructed from `Promote device to gateway`) and
* That `--server-ip` there points to the Gateway device IP address

Then it will work as expected :+1: 

Explanation follows,

When `mender-gateway` is installed as a demo, it uses the `docker.mender.io` Mender Server demo certificates ([these](https://github.com/mendersoftware/mender-gateway/tree/52442cc0ceb4eedb8fee3ab36e537ee4f327527b/examples/cert)), which are also distributed with the `mender-client` deb package.

Then, a client configured with `mender setup --demo` (sic, flag is deprecated) and `mender setup --server-ip` it will trust the demo certificate, set `ServerUrl` to `docker.mender.io` and set an entry like `1.2.3.4      docker.mender.io s3.docker.mender.io` into `/etc/hosts`, in effect making the client send all the requests to the gateway.

I verified this whole setup now using hosted.mender.io and it worked :+1:

oleorhagen:
> > @kacf, @lluiscampos, @oleorhagen the ping for a review is pretty much just for [this](https://github.com/mendersoftware/gui/blob/db2ac70899d3cc3d4901757dba9c3d35daa3de1a/src/js/components/devices/dialogs/__snapshots__/connecttogatewaydialog.test.js.snap#L455-L457) part, since we need to provide the user with an easy way to connect additional devices to the gateway, but I'm not sure if the `--demo` flags in there won't defeat the purpose...
> 
> Assuming
> 
>     * That the device running `mender-gateway` was installed using `--demo` flag (as it is instructed from `Promote device to gateway`) and
> 
>     * That `--server-ip` there points to the Gateway device IP address
> 
> 
> Then it will work as expected +1
> 
> Explanation follows,
> 
> When `mender-gateway` is installed as a demo, it uses the `docker.mender.io` Mender Server demo certificates ([these](https://github.com/mendersoftware/mender-gateway/tree/52442cc0ceb4eedb8fee3ab36e537ee4f327527b/examples/cert)), which are also distributed with the `mender-client` deb package.
> 
> Then, a client configured with `mender setup --demo` (sic, flag is deprecated) and `mender setup --server-ip` it will trust the demo certificate, set `ServerUrl` to `docker.mender.io` and set an entry like `1.2.3.4 docker.mender.io s3.docker.mender.io` into `/etc/hosts`, in effect making the client send all the requests to the gateway.
> 
> I verified this whole setup now using hosted.mender.io and it worked +1

Awesome, thanks!

tranchitella:
How to test it:

```
export TENANT_TOKEN="..."
# gateway device
mender-stress-test-client run \
    --tenant-token "$TENANT_TOKEN" \
    --debug --count=1 \
    --mac-address-prefix ff \
    --inventory-attribute device_type:test \
    --inventory-attribute mender_is_gateway:true \
    --inventory-attribute "mender_gateway_system_id:mysystem01" \
    --inventory-attribute "ipv4_wlan0:192.168.1.110/24" \
    --auth-interval 5 \
    --inventory-interval 5 \
    --update-interval 5 \
    --deployment-time 5
# system devices
mender-stress-test-client run \
    --tenant-token "$TENANT_TOKEN" \
    --debug --count=2 \
    --mac-address-prefix fe \
    --inventory-attribute device_type:test \
    --inventory-attribute "mender_gateway_system_id:mysystem01" \
    --inventory-attribute "ipv4_wlan0:192.168.1.111/24" \
    --auth-interval 5 \
    --inventory-interval 5 \
    --update-interval 5 \
    --deployment-time 5
```

tranchitella:
@mzedel please see my commit

tranchitella (approved):
It works fine; however, it still has the issue that when you click on a device in the device system tab (or the mender gateway icon on a child device), the URL changes but the device details sidebar doesn't change the content. Not sure if it is related to this PR, though.

PR #3339, opened 2022-12-22
MEN-6088 - feat: added information about reporting limits + attributes in use

mzedel:
Ticket: MEN-6088
Changelog: Title


tranchitella (commented):
Thanks, this is exactly what we were thinking about. However, why is it visible in Enterprise only? Shouldn't we provide the same functionality in Open Source? @kjaskiewiczz what do you think?

mzedel:
> However, why is it visible in Enterprise only?

it should be limited to multi-tenant + having reporting only (like the rest of the organization tab in the settings), mostly because unlike the global settings and this is more informational, much like the device limit (+ the OS version is a little less comfortable in other areas too)

tranchitella (approved):
Thanks, makes sense.

PR #3398, opened 2023-01-03
MEN-6130 - allow force deployment

mzedel:


mzedel:
passing pipeline _should_ be [here](https://gitlab.com/Northern.tech/Mender/gui/-/pipelines/737163362)

tranchitella (approved):
Lovely, the code looks code and it works like a charm!

PR #3401, opened 2023-01-05
MEN-6085

mzedel:


tranchitella:
@mzedel looks good, thank you; however, as what we set as "defaults" in the versioning is what we suggest to the user as best practice, what about this change?

```
diff --git a/src/js/components/artifacts/dialogs/addartifact.js b/src/js/components/artifacts/dialogs/addartifact.js
index 0cd441784..81d9da8be 100644
--- a/src/js/components/artifacts/dialogs/addartifact.js
+++ b/src/js/components/artifacts/dialogs/addartifact.js
@@ -184,7 +184,7 @@ export const AddArtifactDialog = ({
     isValidDestination: false,
     name: '',
     selectedDeviceTypes: [],
-    softwareName: 'single-file',
+    softwareName: '',
     softwareVersion: '',
     type: uploadTypes.mender.key
   });
diff --git a/src/js/components/artifacts/dialogs/artifactinformationform.js b/src/js/components/artifacts/dialogs/artifactinformationform.js
index 940bcf399..568ca4694 100644
--- a/src/js/components/artifacts/dialogs/artifactinformationform.js
+++ b/src/js/components/artifacts/dialogs/artifactinformationform.js
@@ -24,11 +24,13 @@ export const ReleaseTooltip = () => (
   </div>
 );
 
+const defaultVersion = '1.0.0';
+
 export const VersionInformation = ({ creation = {}, onRemove, updateCreation }) => {
   const { file, fileSystem: propFs, name, softwareName: propName, softwareVersion: version = '', type } = creation;
   const [fileSystem, setFileSystem] = useState(propFs);
-  const [softwareName, setSoftwareName] = useState(propName);
-  const [softwareVersion, setSoftwareVersion] = useState(version || name);
+  const [softwareName, setSoftwareName] = useState(propName || name.replace('.', '-'));
+  const [softwareVersion, setSoftwareVersion] = useState(version || defaultVersion);
 
   useEffect(() => {
     updateCreation({ finalStep: true });

```

It uses the release name (replacing dots with dashes) as software name, and suggests a default version `1.0.0` to show the user how we expect to set version numbers.

Adding @kjaskiewiczz  and @lluiscampos to the loop, what do you think guys?

kjaskiewiczz:
I think using the release name as software name, and version 1.0.0 as defaults is a very good idea.

tranchitella (approved):

lluiscampos:
I think it is a good idea to not use `single-file` (which is also the update module name) as the default software name :+1:

PR #3406, opened 2023-01-17
fix: fix filterable attributes bar behavior when count is bigger than limit

kjaskiewiczz:
Ticket: [MEN-6212](https://tracker.mender.io/browse/MEN-6212)


kjaskiewiczz:
not sure that's the right solution, but at least it's very cheap

mzedel (approved):
As a preventive measure this might be ok, but it masks a broken backend and should be fixed there.

tranchitella (approved):

tranchitella:
> As a preventive measure this might be ok, but it masks a broken backend and should be fixed there.

@mzedel and @kjaskiewiczz, here is the fix in the backend which prevents to return more attributes than the limit:
https://github.com/mendersoftware/reporting/pull/91

tranchitella (changes requested):
Let's drop this one and fix it in the backend.

PR #3407, opened 2023-01-18
feat: move the reporting limits to the global settings

tranchitella:
Changelog: None
Ticket: None


kjaskiewiczz (approved):
I'm not an expert here, but I've checked it locally (together with https://github.com/mendersoftware/reporting/pull/91) and it was working fine.
I think it fits better in the global settings section, and it's nice to see actual usage and limit.
Thanks a lot!

mzedel (changes requested):
No. This is not actionable and thus was deliberately not put to the settings section.

mzedel (approved):

PR #3413, opened 2023-01-26
MEN-6051 - device deployment history + reset

mzedel:


tranchitella:
When I open the deployments history and click outside of the side panel I see this error.

image

The URL of the API call is wrong:
https://localhost/api/management/v1/deployments/deployments/devices/undefined?page=1&per_page=10

Another minor issue I am experiencing is that when I click the "DELETE DEVICE DEPLOYMENT HISTORY" and I confirm the operation clicking on the green checkmark, the confirmation message/prompt doesn't disappear.

Finally: are we sure we want to call this button "DELETE DEVICE DEPLOYMENT HISTORY"? We are not deleting the entries (if not "logically"). Maybe we should use a different wording? "RESET DEVICE DEPLOYMENT HISTORY", for example? @michaelatmender ?

mzedel:
Thanks for taking a look @tranchitella ! I'll fix the first two issues, for the naming I switched from reset to deletion to align with the API & docs here https://docs.mender.io/api/development/#management-api-deployments-delete-device-deployments-history

tranchitella:
You are right, we should probably change the title to replace "Delete" with "Reset".
Are you okay with this change @mzedel and @kjaskiewiczz?

michaelatmender (changes requested):
Just a couple of issues when testing:

- Reset history button doesn't seem to have any effect
- Deployments in the 'queued to start' pending status don't show (possibly because the device doesn't know about the deployment yet? May be unavoidable. but thought I'd mention it)

tranchitella:
> * Deployments in the 'queued to start' pending status don't show (possibly because the device doesn't know about the deployment yet? May be unavoidable. but thought I'd mention it)

This is okay: the backend will create the device deployment object once the device checks in.
Nothing we can fix in the UI.

tranchitella:
> Just a couple of issues when testing:
> * Reset history button doesn't seem to have any effect

About this one: entries which are reset/deleted have a `deleted` property with a timestamp. Maybe we should mark them in a special way (e.g. strikethrough, or an icon) to make it clear that they will be ignored when looking up the next update?

tranchitella:
> You are right, we should probably change the title to replace "Delete" with "Reset". Are you okay with this change @mzedel and @kjaskiewiczz?

Done:
* https://github.com/mendersoftware/deployments/pull/818
* https://github.com/mendersoftware/deployments-enterprise/pull/765

mzedel:
would this:
Screenshot 2023-01-30 at 12 54 11
be an option @michaelatmender & @tranchitella ?

tranchitella:
> would this: Screenshot 2023-01-30 at 12 54 11 be an option @michaelatmender & @tranchitella ?

It looks very nice!

michaelatmender:
> would this: be an option @michaelatmender & @tranchitella ?

Yes, that's a nice solution for now! This wasn't considered in the initial UX, but if necessary we can revisit it later. Looks good!

tranchitella (approved):
It works perfectly, well done!

PR #3421, opened 2023-01-30
Chore/public route test

mzedel:
since the date in the footer is based on the environment date, we just need to check that the returned value is in line with the environment (and not the current date by accident)

### DIFF
--- a/src/js/components/app.js
+++ b/src/js/components/app.js
@@ -3,7 +3,9 @@ import { useIdleTimer, workerTimers } from 'react-idle-timer';
 import { connect } from 'react-redux';
 import { useLocation, useNavigate } from 'react-router-dom';
 
+import { CssBaseline } from '@mui/material';
 import { ThemeProvider, createTheme } from '@mui/material/styles';
+import withStyles from '@mui/styles/withStyles';
 import { makeStyles } from 'tss-react/mui';
 
 import Cookies from 'universal-cookie';
@@ -17,7 +19,6 @@ import { PrivateRoutes, PublicRoutes } from '../config/routes';
 import { onboardingSteps } from '../constants/onboardingConstants';
 import ErrorBoundary from '../errorboundary';
 import { toggle } from '../helpers';
-import { WrappedBaseline } from '../main';
 import { getOnboardingState, getUserSettings } from '../selectors';
 import { dark as darkTheme, light as lightTheme } from '../themes/Mender';
 import Tracking from '../tracking';
@@ -33,6 +34,34 @@ import Uploads from './uploads';
 const activationPath = '/activate';
 const timeout = 900000; // 15 minutes idle time
 const cookies = new Cookies();
+
+const reducePalette =
+  prefix =>
+  (accu, [key, value]) => {
+    if (value instanceof Object) {
+      return {
+        ...accu,
+        ...Object.entries(value).reduce(reducePalette(`${prefix}-${key}`), {})
+      };
+    } else {
+      accu[`${prefix}-${key}`] = value;
+    }
+    return accu;
+  };
+
+const cssVariables = ({ palette }) => {
+  const muiVariables = Object.entries(palette).reduce(reducePalette('--mui'), {});
+  return {
+    '@global': {
+      ':root': {
+        ...muiVariables,
+        '--mui-overlay': palette.grey[400]
+      }
+    }
+  };
+};
+
+const WrappedBaseline = withStyles(cssVariables)(CssBaseline);
 
 const useStyles = makeStyles()(() => ({
   public: {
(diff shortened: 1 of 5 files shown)


src/js/components/app.test.js
```diff
@@ -58,16 +61,25 @@ describe('App Component', () => {
...
       </Provider>
     );
-    await act(async () => jest.advanceTimersByTime(900500));
+    const { rerender } = render(ui);
+    await act(async () => jest.advanceTimersByTime(timeout + 500));
+    cookies.get.mockReturnValue('');
+    await waitFor(() => rerender(ui));
+    await waitFor(() => expect(screen.queryByText(/Version:/i)).not.toBeInTheDocument());
+    expect(screen.queryByText(/Northern.tech/i)).toBeInTheDocument();
+    expect(screen.queryByText(`© ${mockDate.getFullYear()} Northern.tech AS`)).toBeInTheDocument();
```
merlin-northern, line 83: just for my peace of mind, `testDate` from `mockData` is set as the current date of the test run, right?

src/js/components/app.test.js
```diff
@@ -58,16 +61,25 @@ describe('App Component', () => {
...
-    render(
+    window.localStorage.getItem.mockReturnValueOnce('false');
+    const ui = (
       <Provider store={store}>
         <App />
       </Provider>
     );
-    await act(async () => jest.advanceTimersByTime(900500));
+    const { rerender } = render(ui);
+    await act(async () => jest.advanceTimersByTime(timeout + 500));
```
merlin-northern, line 78: in general this seems to a a long wait. two questions:
* why are we waiting that long?
* why are we adding a very small fraction of that wait here?

merlin-northern (approved):
lgtm, my-learning questions and some nit comments.

src/js/components/app.js
```diff
@@ -34,6 +35,34 @@ const activationPath = '/activate';
...
+    return accu;
+  };
+
+const cssVariables = ({ palette }) => {
+  const muiVariables = Object.entries(palette).reduce(reducePalette('--mui'), {});
+  return {
+    '@global': {
+      ':root': {
+        ...muiVariables,
+        '--mui-overlay': palette.grey[400]
```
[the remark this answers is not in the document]
mzedel: the 400 is basically using the definition... defined in [here](https://github.com/mendersoftware/gui/blob/master/src/js/themes/Mender/light.js#L8-L19)
using a constant representation would be `palette.grey['400']`, but then I would think `palette.grey.400` should work, but that won't work in javascript

as to why not to define this with a more granular name: the number approach is supposed to target the different shades of a color (most often grey, due to the popularity of using shades of grey), since the naming depends so much on subjective perception/ the display you are using/ ... in CSS there's e.g. `dimgrey` and `darkgrey` - guess which is darker?

src/js/components/app.test.js
mzedel: Yes, that date is used to mock the environment/ browser date, so what the footer renders is a `new Date().getFullYear()` and these should be equal - unless the mock doesn't work... and keeping it at a live date would break many other parts of the test that either show a date we need to compare or use the date as a seed for a request e.g.

src/js/components/app.test.js
mzedel: the `timeout` is what is used in the UI to do the auto logout when a user is idle - instead of waiting for that time to pass, the `advanceTimersByTime` forwards the time in the test run by the defined amount, so we skip ahead by just a bit more than the timeout and cause the idle check to trigger => this should get us logged out

src/js/components/app.js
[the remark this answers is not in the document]
merlin-northern: I bet dimgrey is darker!

src/js/components/app.test.js
merlin-northern: not to mention runs right before midnight on Ner Years Eve.
nice.

src/js/components/app.test.js
merlin-northern: right! I was missing some parts. thanks.

merlin-northern (approved):
thanks for being patient and answering my newbie questions.

mzedel:
> thanks for being patient and answering my newbie questions.

thank you for asking @merlin-northern! and for nagging me to add this in the first place!

src/js/components/app.js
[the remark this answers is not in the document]
mzedel: so... on my display: yes!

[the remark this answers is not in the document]
merlin-northern: ha! you see, not only I know my colour consts, I also am familiar with your display :>

PR #3425, opened 2023-01-30
MEN-5921 - dashboard refactoring

mzedel:


tranchitella:
@mzedel it is super nice and it works fine. The only issue I have is that it doesn't work correctly with dynamic groups: if you configure the widget to work on a dynamic group, instead of using the dynamic group filter, it queries the aggregation end-point filtering by device group's name using the dynamic group name.

tranchitella (approved):
Thanks

PR #3452, opened 2023-02-01
chore: adjusted e2e test runner to account for load related failures

mzedel:


lluiscampos (approved):
Okey. But can you give more context on what are the failures so that I can follow-up? How is that moving the jobs back to public runners fixes it?

mzedel:
Since the switch to our runners the OS e2e test setup with a running client started failing (https://gitlab.com/Northern.tech/Mender/gui/-/pipelines?page=1&scope=all&ref=master should show the difference to the weeks before). Sometimes the job would pass on a rerun (e.g. [here](https://gitlab.com/Northern.tech/Mender/gui/-/jobs/3696434039)), sometimes after several reruns (unfortunately I can't filter for these in Gitlab)) The biggest hint I saw was that it was happening especially often during the dependabot CI runs yesterday, when the runners were loaded (or at least many pipelines were waiting for a runner to become available).
Strangely enough I tried using the more powerful runner variants (e.g. [here](https://gitlab.com/Northern.tech/Mender/gui/-/jobs/3696068968)) and they would also fail (but OTOH I would also have to wait for a runner to be available so I assume they were not running isolated either).
I could live with the added waiting time for jobs to run, but the test instability due to the runners is not an option.

some more example failures:
- https://gitlab.com/Northern.tech/Mender/gui/-/jobs/3699825545
- https://gitlab.com/Northern.tech/Mender/gui/-/jobs/3697784142
- https://gitlab.com/Northern.tech/Mender/gui/-/jobs/3697275084
- https://gitlab.com/Northern.tech/Mender/gui/-/jobs/3696453793
- https://gitlab.com/Northern.tech/Mender/gui/-/jobs/3696432532
- https://gitlab.com/Northern.tech/Mender/gui/-/jobs/3696416727
- https://gitlab.com/Northern.tech/Mender/gui/-/jobs/3696437251
- https://gitlab.com/Northern.tech/Mender/gui/-/jobs/3696581926

PR #3474, opened 2023-02-10
fix(e2e-tests): enabled addons on tenant creation

mzedel:


tranchitella (approved):
I'm really sorry @mzedel, I didn't know you are creating tenants using the CLI when running acceptance tests otherwise I'd have fixed it by myself.

PR #3491, opened 2023-02-22
chore: tried to distinguish system & os updates

mzedel:


tranchitella (approved):
Nice, thank you very much!

PR #3498, opened 2023-02-28
SEC-706 - chore: added license check for direct dependencies

mzedel:
+ adjusted license to be timelessly relevant


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright 2017 Northern.tech AS
+Copyright Northern.tech AS
```
lluiscampos, line 1: We need to keep the year as we have for the rest of the repositories. Additionally, we should add the `mendertesting` test to verify that the year is up to date.

I know there are good arguments for removing it, but we need first to adopt such a policy (across all Mender repos or across all NT) and then follow-up accordingly everywhere.

mzedel: This aligns with the best practice in the frontend space (so it should be just fine for this repository) - if that seems not to be relevant, then use this as an example of me aligning us closer with [CFEngine](https://github.com/cfengine/core/blob/master/LICENSE) or our company wide used [libntech](https://github.com/cfengine/libntech/blob/master/LICENSE).
To follow our principle of continuous improvement I would rather suggest to adjust `mendertesting` and follow up with the other Mender repos - now would be a very good time, given that the last release is just out!

kacf: @mzedel: I got the license directly from Thomas back in the day. The two projects do not have the same license requirements because CFEngine uses copyright assignment and Mender doesn't.

I don't know how important the year in this file, but please do not change this without having discussed it with Thomas first. I am not a lawyer so I cannot really answer it myself.

mzedel: @kacf then I'll check with Thomas for this.
As to the different handling in CFEngine and here - that shouldn't apply, since we use libntech as part of the product already - if it matters we should probably halt the 5.3.1 release and fix it...

kacf: My gut feeling is that you're probably right, it can be removed. But I don't want to do this without having checked first.

lluiscampos: I also think we can remove it (and I really would like to do so - here and everywhere) but it needs to be confirm by the legal department :)

mzedel: after checking with the legal department: we can't remove it, but having a year and sticking with it is enough - so I added the copyright year for all files based on when they were first added to git...

tranchitella (approved):

kacf (approved):

PR #3537, opened 2023-03-02
test: Disable tests that keep failing in staging.

kacf:
Ticket: QA-579

### DIFF
--- a/tests/e2e_tests/integration/07-saml.spec.ts
+++ b/tests/e2e_tests/integration/07-saml.spec.ts
@@ -47,7 +47,10 @@ test.describe('SAML Login', () => {
     });
 
     // Setups the SAML/SSO login with samltest.id Identity Provider
-    test('Set up SAML', async ({ context, environment, baseUrl, page }) => {
+    test('Set up SAML', async ({ browserName, context, environment, baseUrl, page }) => {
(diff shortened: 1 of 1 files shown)


mzedel (approved):
Excellent thank you!

tests/e2e_tests/integration/07-saml.spec.ts
```diff
@@ -48,6 +48,9 @@ test.describe('SAML Login', () => {
 
     // Setups the SAML/SSO login with samltest.id Identity Provider
     test('Set up SAML', async ({ context, environment, baseUrl, page }) => {
+      // QA-579
```
mzedel, line 51: ah I guess I was a bit too excited there - the `browserName` is not yet available to check for... =>
```suggestion
    test('Set up SAML', async ({ browserName, context, environment, baseUrl, page }) => {
      // QA-579
```

mzedel (changes requested):

tests/e2e_tests/integration/07-saml.spec.ts
kacf: Good catch, fixed!

mzedel (approved):

PR #3541, opened 2023-03-08
ME-59 - feat: added possibility to create never expiring PATs

mzedel:
- also added a warning about the risks involved

Ticket: ME-59
Changelog: Title


tranchitella:
@mzedel it works in combination with https://github.com/mendersoftware/useradm/pull/354
However, we cannot (easily) omit the expiration date. I'm setting the end of the century, which the UI renders as expiration in "77 years". What do you suggest?

tranchitella (approved):

PR #3542, opened 2023-03-09
Fix/firefox staging

mzedel:


mzedel:
passing firefox run [here](https://gitlab.com/Northern.tech/Mender/gui/-/jobs/3903302104)

tranchitella (approved):

PR #3543, opened 2023-03-09
chore: let staying logged in rely on backend expiration handling

mzedel:


tranchitella (approved):
Pre-approving, I tested it (with the change) and it works.

tranchitella (approved):

PR #3564, opened 2023-03-13
QA-579

mzedel:
pipeline with passing staging tests is [here](https://gitlab.com/Northern.tech/Mender/gui/-/pipelines/963019029)


tranchitella (approved):
Mainly based on the fact that there is a green pipeline.
I didn't check the actual diff, I trust your judgement.

PR #3594, opened 2023-03-21
chore: update legal agreement links and footer color

michaelatmender:
Ticket: None
Changelog: None


mzedel (approved):
just the one failing test needs fixing... I guess I was a bit overly specific in the check for the footer content + sorry about that!

PR #3597, opened 2023-03-28
MEN-6347 - the releases section part

mzedel:


tranchitella:
@mzedel got this

image

src/js/actions/appActions.js
```diff
@@ -82,14 +93,8 @@ export const parseEnvironmentInfo = () => (dispatch, getState) => {
       trackerCode: trackerCode || state.app.trackerCode
     };
     environmentFeatures = {
-      hasAddons: stringToBoolean(features.hasAddons),
-      hasAuditlogs: stringToBoolean(features.hasAuditlogs),
-      hasMultitenancy: stringToBoolean(features.hasMultitenancy),
-      hasDeviceConfig: stringToBoolean(features.hasDeviceConfig),
-      hasDeviceConnect: stringToBoolean(features.hasDeviceConnect),
-      hasMonitor: stringToBoolean(features.hasMonitor),
+      ...featureFlags.reduce((accu, flag) => ({ ...accu, [flag]: stringToBoolean(features[flag]) }), {}),
```
tranchitella, line 96: neat!

src/js/components/releases/releases.js
```diff
@@ -0,0 +1,314 @@
...
+
+  const onTagSelectionChanged = ({ selection }) => setReleasesListState({ selectedTags: selection });
+
+  const availableTabs = useMemo(
+    () =>
+      tabs.reduce((accu, tab) => {
+        if (tab.isApplicable({ features, tenantCapabilities })) {
+          accu.push(tab);
+        } else {
+          accu.push({ ...tab, disabled: true });
```
tranchitella, line 86: Just a confirmation: if the tab is not supported, we render it as "disabled" (click and nothing happens". Right?

tranchitella:
Sorry, another small thing: the icon to sort the columns is misaligned:

image

src/js/components/releases/releases.js
mzedel: yes... but to give a bit more context we could show a tooltip and track any clicks on it to gauge interest 🙃
(same goes with the tag functionality that's hidden behind the `$HAVE_RELEASE_TAGS`)

mzedel:
@tranchitella thanks a lot for the nice review! the issues _should_ now be addressed 🤞

src/js/components/releases/releases.js
michaelatmender: I think it would be better to allow the user to click it, but show placeholder text - similar to the empty states we have elsewhere. I should have specified in the figma files, sorry! I can take care of it. It shouldn't block this PR though

src/js/components/releases/releases.js
```diff
@@ -0,0 +1,323 @@
...
+        </Tabs>
+        <div>
+          {canUpload && (
+            <>
+              <Button
+                ref={uploadButtonRef}
+                color="secondary"
+                onClick={onUploadClick}
+                startIcon={<CloudUpload fontSize="small" />}
+                style={{ marginTop: 30, minWidth: 164 }}
```
michaelatmender, line 107: Can we remove `marginTop: 30`? It messes up the layout a bit

michaelatmender:
Just a couple of small details I found when playing around with it: 

If you delete a Release, the next Release in the list automatically opens in the slide panel view - it would be better just to close the slide panel instead, I believe.

The interaction with deleting a Release has some overlapping buttons/labels:
delete Release
 - it might be necessary to move the confirmation to a small dialog - in future if we allow deleting multiple releases from the list view, it would work there too!

src/js/components/releases/releaseslist.js
```diff
@@ -0,0 +1,106 @@
...
+  if (repoRef.current?.lastChild?.lastChild) {
+    const element = repoRef.current.lastChild.lastChild;
+    const anchor = { left: element.offsetLeft + element.offsetWidth / 2, top: element.offsetTop + element.offsetParent?.offsetTop + element.offsetHeight };
+    onboardingComponent = getOnboardingComponentFor(onboardingSteps.ARTIFACT_INCLUDED_ONBOARDING, { ...onboardingState, artifactIncluded }, { anchor });
+    onboardingComponent = getOnboardingComponentFor(onboardingSteps.DEPLOYMENTS_PAST_COMPLETED, onboardingState, { anchor }, onboardingComponent);
+  }
+
+  const potentialTotal = searchTerm ? searchTotal : total;
+  return (
+    <div style={{ maxWidth: '70%' }}>
```
michaelatmender, line 83: I think we can replace the 70% maxWidth with a px - it doesn't work great on smaller screens with this limit. We can go pretty wide: I suggest 1600px to give it full width on small and medium screens, but still keeping it readable on large screens :)

michaelatmender (changes requested):
See my comments - just a couple of small visual things, and UX quirks that were missed in the initial designs

src/js/components/releases/releases.js
mzedel: with a `marginTop: 12` it aligned with the tab underline - would that also be ok?

michaelatmender: Perfect!

michaelatmender (approved):

PR #3598, opened 2023-03-29
ME-64 - fix: made device group check rely on per device information for more reliable error messages

mzedel:
- also made device quick action setup slightly less repetitive

Ticket: 
Changelog: Title

### DIFF
--- a/src/js/components/devices/expanded-device.js
+++ b/src/js/components/devices/expanded-device.js
@@ -190,6 +190,7 @@ const tabs = [
 
 export const ExpandedDevice = ({
   abortDeployment,
+  actionCallbacks,
   applyDeviceConfig,
   columnSelection,
   decommissionDevice,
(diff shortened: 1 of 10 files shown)


tranchitella:
@mzedel minor issue when playing with the PR:
image

src/js/components/devices/expanded-device.js
```diff
@@ -285,16 +281,6 @@ export const ExpandedDevice = ({
...
   const scrollToMonitor = () => setDetailsTab('monitor');
 
-  const onCreateDeploymentClick = () => navigate(`/deployments?open=true&deviceId=${deviceId}`);
-
-  const actionCallbacks = {
-    onAddDevicesToGroup,
-    onAuthorizationChange,
-    onDeviceDismiss,
-    onRemoveDevicesFromGroup,
-    onPromoteGateway: onMakeGatewayClick,
```
tranchitella, line 295: How does this end up in the actionCallbacks after your change?

mzedel: Argh I missed that, thanks a lot for catching this!

PR #3659, opened 2023-04-25
chore: reverted reliance on reporting

mzedel:
this _should_ make the UI use reporting as an optional component again...

FYI @alfrunes

### DIFF
--- a/src/js/actions/deviceActions.js
+++ b/src/js/actions/deviceActions.js
@@ -82,7 +82,7 @@ export const getGroups = () => (dispatch, getState) =>
       const ungroupedDevices = promises[promises.length - 1] || [];
       const result = ungroupedDevices[ungroupedDevices.length - 1] || {};
       if (!result.total) {
-        return;
+        return Promise.resolve();
       }
       return Promise.resolve(
         dispatch({


src/js/components/devices/device-details/devicesystem.js
```diff
@@ -71,6 +73,7 @@ export const DeviceSystem = ({
         <TwoColumnData config={{ 'Server IP': deviceIp }} compact setSnackbar={setSnackbar} />
       </DeviceDataCollapse>
       <DeviceDataCollapse className={classes.container} title="System for this gateway">
+        <EnterpriseNotification isEnterprise={hasFullFiltering} benefit="see devices connected to your gateway device for easy access" />
```
tranchitella, line 76: Didn't get this one: what is it?

tranchitella (approved):

src/js/components/devices/device-details/devicesystem.js
mzedel: Once we have reporting, all plans will be able to use all filtering options, until then at least users on the basic plan won't be able to see devices from systems (since retrieving them relies on `$ne`) - unlikely to be shown often, but was part of the reverted commit...

tranchitella: FYI: systems are Enterprise only as they rely on Mender Gateway.

mzedel: Right, I had the feeling in the back of my head, but was misled by the code, since the rest just relied on less explicit checks for this => https://github.com/mendersoftware/gui/pull/3664

PR #3704, opened 2023-05-26
feat: raised visibility of existing features to help increase awareness

mzedel:
Ticket: None
Changelog: None

### DIFF
--- a/src/js/__snapshots__/main.test.js.snap
+++ b/src/js/__snapshots__/main.test.js.snap
@@ -261,6 +261,21 @@ exports[`Main Component renders correctly 1`] = `
             </span>
           </div>
         </a>
+        <a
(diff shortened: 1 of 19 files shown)


tranchitella:
@mzedel I tried reviewing this PR but I'm not sure I understand what you wanted to achieve.
For example, running an open-source server gives me this screen:

image

Is it correct? We want to show the whole audit logs view plus the Enterprise notification?

mzedel:
that was my initial suggestion to make it visible for everyone - but considering the examples linked in [MEN-6457](https://northerntech.atlassian.net/browse/MEN-6457) I think it's better to close this and focus on the new approach 👌

mzedel:
In any case: thanks a lot for looking into this!

PR #3705, opened 2023-05-29
:warning:  draft - PoC for non-root user

oldgiova:
Ticket: SEC-849

* using `nginx` user
* using configurable port, set as `9080` by default.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,16 +10,38 @@ FROM base AS build
 COPY . ./
 RUN npm run build
 
-FROM nginx:1.23.4-alpine
+FROM nginx:1.23.4-bullseye
 EXPOSE 8080
 WORKDIR /var/www/mender-gui/dist
 ARG GIT_COMMIT_TAG
 ENV GIT_COMMIT_TAG="${GIT_COMMIT_TAG:-local_local}"
+ARG NGINX_PORT
+ENV NGINX_PORT=${NGINX_PORT:-9080}
 
-COPY ./entrypoint.sh /entrypoint.sh
+RUN apt-get update && apt-get install -y --no-install-recommends \
+  wget \
+  && apt-get clean \
+  && rm -rf /var/lib/apt/lists/*
+
 COPY httpd.conf /etc/nginx/nginx.conf
-COPY --from=build /usr/src/app/dist .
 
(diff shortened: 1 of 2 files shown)


mzedel (changes requested):
no - we had a related discussion e.g. here: https://hub.mender.io/t/problems-with-mender-helm-chart-on-openshift-privileged-ports/5041

the main point is, that this would most likely break existing installations + there is a somewhat supported image directly from nginx (to create a custom built one on top of ours)...

oldgiova:
> no - we had a related discussion e.g. here: https://hub.mender.io/t/problems-with-mender-helm-chart-on-openshift-privileged-ports/5041
> 
> the main point is, that this would most likely break existing installations + there is a somewhat supported image directly from nginx (to create a custom built one on top of ours)...

Thanks for the review. Let's try a different approach: preparing the field for a possible future switch: I added a new listen stanza with unprivileged port alongside the old one:
```
  server {
    listen 80;
    ...
  }
  server {
    listen 9080;
    ...
  }
```
This way we have backward compatibility. Unfortunately we still cannot safely switch to rootless containers unless breaking existing setup.
This modification is to do alongside a new `mender-helm` Chart version with `gui.httpPort: 8090`, still root.

From now on we can start distributing the new Helm Chart to the customers with the default `gui.httpPort: 8090` port. In the future we could drop the `listen 80` nginx stanza and switch to `nginx` users.

mzedel:
> Thanks for the review. Let's try a different approach: preparing the field for a possible future switch: I added a new listen stanza with unprivileged port alongside the old one

Please no. The duplication will just invite things to diverge (e.g. in the CSP)... also the default port has to stay at 80 if we want to keep backwards compatibility.

Instead of doing the modifications to the nginx setup ourselves I would much rather use the upstream unprivileged nginx image to not have to care about any changes to the setup...
Also as of today we still have docker-compose based installations as default (also the only approach we test), so we can't rely on changing a helm config yet.

I would go so far as to publish an additional image based on the unprivileged nginx via an extra stage in the Dockerfile, but sacrificing the compatibility of the majority of installations for the requirements of few seems bad if the alternative would be any downstream user building their image with ours as base...

oldgiova:
Superseded by https://github.com/mendersoftware/gui/pull/3709

PR #3738, opened 2023-06-01
chore: Update google analytics to GA4

michaelatmender:
Ticket: MEN-6503
Changelog: None


michaelatmender:
Also: https://github.com/mendersoftware/saas/pull/994 and 
https://github.com/mendersoftware/mender-kustomize/pull/2

src/js/tracking.js
```diff
@@ -79,9 +79,6 @@ class Tracker {
     } else {
       data = this.currentPageView;
     }
-    if (this.initialized && this.trackingEnabled) {
-      ReactGA.pageview(data);
-    }
```
mzedel, line 84: based on https://github.com/codler/react-ga4#example this likely has to become:
```
ReactGA.send({ hitType: "pageview", page: data, title: "Some Title" });
```
in order for us to keep tracking pageviews...

mzedel (changes requested):

mzedel:
and the [license checking file](https://github.com/mendersoftware/gui/blob/master/tests/licenses/directDependencies.csv#L81) needs to be adjusted to the MIT license and the `github.com/codler/react-ga4` location

src/js/tracking.js
michaelatmender: I understood that pageviews should be automatically tracked now, unless we do something custom:
https://developers.google.com/analytics/devguides/migration/measurement/virtual-pageviews#comparison_of_ua_and_ga4_virtual_pageviews
And (the _don't do this_ image) in https://medium.com/@vergil333/google-analytics-4-in-react-a250005c2455

mzedel: oh nice! I wasn't aware of this, then sorry about the noise 🙈

mzedel (approved):

PR #3741, opened 2023-06-02
MEN-6487 - fix: added more granular check for device troubleshooting feature

mzedel:
Ticket: MEN-6487
Changelog: Title


michaelatmender (changes requested):
Testing this, and I still see the File Transfer UI even as a user with no troubleshooting/connect permissions

michaelatmender:
> > Testing this, and I still see the File Transfer UI even as a user with no troubleshooting/connect permissions
> 
> yes, since read access to devices should still allow you to download files from a device...

Got it! That makes sense. Perhaps we shouldn't show the 'upload' tab in this case? That can be a separate ticket though

mzedel:
Thanks for the review @michaelatmender - I adjusted the dialog to take this into account 👌

michaelatmender:
> Thanks for the review @michaelatmender - I adjusted the dialog to take this into account 👌
Almost there! It still loads the file upload view by default for Read-only users: 
image

Otherwise, all good

michaelatmender (approved):

PR #3752, opened 2023-06-07
chore(e2e-tests): slightly simplified log in test + separated log out test

mzedel:


tests/e2e_tests/integration/01-login.spec.ts
```diff
@@ -52,6 +47,14 @@ test.describe('Login', () => {
       expect(loginVisible).toBeTruthy();
     });
 
+    test('Logs out using UI', async ({ loggedInPage: page }) => {
+      await page.waitForSelector(loggedInText);
+      // now we can log out
+      await page.click('.header-dropdown', { force: true });
+      await page.click(`text=/Log out/i`, { force: true });
+      await page.waitForSelector('text=/log in/i', { timeout: 7000 });
```
merlin-northern, line 55: I would check for `Log in with:` and `const reloadTimeoutMillis = 7000`

tests/e2e_tests/integration/01-login.spec.ts
```diff
@@ -34,16 +36,9 @@ test.describe('Login', () => {
...
+      await page.waitForSelector(loggedInText);
       await page.evaluate(() => localStorage.setItem(`onboardingComplete`, 'true'));
       await context.storageState({ path: 'storage.json' });
-      // now we can log out
-      await page.click('.header-dropdown', { force: true });
-      await page.click(`text=Log out`, { force: true });
-      const loginButton = page.getByRole('button', { name: /log in/i });
-      await loginButton.waitFor({ timeout: 7000 });
-      const loginVisible = await loginButton.isVisible();
-      expect(loginVisible).toBeTruthy();
```
merlin-northern, line 46: occasionally what happens to me is: I have a stale JWT in the local store, and there is no other standard way to exit this situation than pressing logout. I would ensure that we are logged out before we login, or even before we start any test case that requires or assumes a logged out user.

tests/e2e_tests/integration/01-login.spec.ts
```diff
@@ -52,6 +47,14 @@ test.describe('Login', () => {
       expect(loginVisible).toBeTruthy();
     });
 
+    test('Logs out using UI', async ({ loggedInPage: page }) => {
+      await page.waitForSelector(loggedInText);
```
merlin-northern, line 51: this will work because this test will wait for a successful login from the other test case?

merlin-northern (commented):
approving not to block, left some comments, trying to be mean, as usual, and as usual sorry for that.

merlin-northern (approved):

tests/e2e_tests/integration/01-login.spec.ts
mzedel: that's what the `page` fixture ensures - a new test context, without any cookies etc. present - also we wouldn't be able to focus/ fill the email & password inputs otherwise 👌

tests/e2e_tests/integration/01-login.spec.ts
merlin-northern: if you have code to remove, do not wait for my permission :)

merlin-northern:
lgtm thanks for explaining.

PR #3753, opened 2023-06-07
chore: added automation to combine dependabot PRs

mzedel:
using https://github.com/github/combine-prs - if accepted & nobody objects I would add this to the other repos too...


alfrunes (approved):

oleorhagen:
Nice! No objections here

PR #3784, opened 2023-06-11
:warning: Test in Progress: support for multiarch oci build

oldgiova:
Ticket: MC-6871

## end goal
To have OCI images multi-arch tested

## Proposed solution
As of now we are exporting built docker images to a tar file and having tested and pushed it on the downstream pipelines. This seems not possible with multi-arch images. 
So to workaround this, and to leverage Gitlab tools, the proposed solution is to use the Gitlab registry as a temporary storage for the downstream pipelines:
1. this PR build the multi-arch image and pushes it to the Gitlab Registry
2. TODO: the downstream pipelines have to be modified to not load the docker image from the Artifact, but pulling from the Gitlab Registry.

@mzedel what do you think?

# Note: incompleted - not to be merged

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -24,6 +24,7 @@ include:
 .export_docker_vars: &export_docker_vars |
   DOCKER_BUILD_TAG=${CI_COMMIT_REF_SLUG:-local}
   DOCKER_BUILD_SERVICE_IMAGE=${DOCKER_REPOSITORY}:${DOCKER_BUILD_TAG}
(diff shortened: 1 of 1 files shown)


mzedel:
hmm interesting, I haven't considered using the gitlab registry as the cache... if we load the image and re-tag it before the test runs it might make it easy 👍
the only issue I could see is tag collisions (e.g. when the dependabot PRs come in, the tags might overlap) - but that should be fixable by using the pipeline ID as tag and then maybe do an image cleanup in the final pipeline stage...

mzedel:
@oldgiova with #4035 in this can be closed, right?

oldgiova:
> @oldgiova with #4035 in this can be closed, right?

Yes

PR #3791, opened 2023-06-13
fix: made search result disappear when clicking outside of it

mzedel:
fyi @chiachenglu 

Reviewer notes:
The search result meant in the PR title is from the global search 👌


michaelntech (approved):

mzedel:
@aleksandrychev FYI

PR #3792, opened 2023-06-13
MEN-6531 - fix: closed release details when onboarding to prevent tooltips from being hidden

mzedel:


merlin-northern (approved):
lgtm. sorry for the huge delay in reviews.

PR #3795, opened 2023-06-15
feat: increased chances of adopting our intended navigation patterns instead of unsupported browser navigation

mzedel:


michaelntech (commented):
Athough it fixes the bug with the unclickable UI, I find it to be worse for usability than the current back button behavior - much of the time the back button has no effect at all, and only sometimes it does what I expect.
I added some comments in the bug ticket

PR #3798, opened 2023-06-21
MEN-3530 - refactored forms

mzedel:
Reviewer notes:
This is primarily a refactoring and should not be noticeable, but the forms that are most affected by the change are:
- login (including 2fa code entry)
- signup
- user creation
- own password & email change

All of this should be well covered by the e2e tests on staging in case something should not work, but a good look would be appreciated!

An easy way to get to a hosted style setup is:
- start an enterprise demo environment
- exec into the gui container
- edit the `env.js` file to say `isHosted: "1"`


alfrunes:
Not sure if it's related to the changes in this PR, but trying to start the compose setup using "Hosted" mode, I get the following error:
```
react-dom.production.min.js:216 TypeError: Cannot read properties of undefined (reading 'charAt')
    at appActions.js:115:38
    at index.js:16:18
    at app.js:105:5
    at Il (react-dom.production.min.js:262:359)
    at t.unstable_runWithPriority (scheduler.production.min.js:18:343)
    at Ui (react-dom.production.min.js:122:325)
    at Tl (react-dom.production.min.js:261:308)
    at react-dom.production.min.js:261:215
    at I (scheduler.production.min.js:16:224)
    at _.port1.onmessage (scheduler.production.min.js:12:346)
```

mzedel:
Thanks for looking into this @alfrunes ! I might have been a bit ambiguous with the instructions there 🙈... Sorry about that!
Only the `isHosted` line had to be adjusted and the rest of the file left as is.. the error message less me to believe that the rest of the file is empty/ at least the `integrationVersion` no longer is present...

alfrunes (changes requested):
Did some testing and found an issue with the button for generating the password does not unlock the submit button even though the form is complete; you have to select and deselect the password form to unlock the submit button. This applies to the user creation and password modification (my profile).

mzedel:
Thanks for looking into and catching this @alfrunes! It should now be working properly...

lluiscampos (changes requested):
When creating a new user, I cannot click `Create user` without writing a password (even after I click the checkbox `Send an email to the user containing a link to reset the password`, which seems to be optional in HM).

image

I did not spot anything else in other areas

mzedel:
> When creating a new user, I cannot click `Create user` without writing a password (even after I click the checkbox `Send an email to the user containing a link to reset the password`, which seems to be optional in HM).
Thank you very much for catching this @lluiscampos! This should now be working as expected, but please have another look...

lluiscampos (approved):
Another thing that I found odd is that when not selecting any role when creating a user the user gets `Admin`. This is pre-existing and might be by design, I just intuitively would expect that with no role selected the new user should default to the less privileged role.

Nothing else that I could spot :green_circle:

mzedel:
Yep, that is by design/ handled in the backend. I could make the role selection default to admin to signal this - but would rather put this in a separate PR 😅

mzedel:
@aleksandrychev this most likely too

PR #3799, opened 2023-06-22
MEN-6553

mzedel:
moved loading indicator in deployment devices list next to pagination controls to prevent layout shift (much like it is for the devices list)

### DIFF
--- a/src/js/actions/deploymentActions.js
+++ b/src/js/actions/deploymentActions.js
@@ -3,7 +3,7 @@ import isUUID from 'validator/lib/isUUID';
 
 import { commonErrorHandler, setSnackbar } from '../actions/appActions';
 import GeneralApi, { apiUrl, headerNames } from '../api/general-api';
-import { SORTING_OPTIONS } from '../constants/appConstants';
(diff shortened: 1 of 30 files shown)


lluiscampos (commented):
I am uncertain about what the issue was/is. Assuming that reported issue is in Deployments pane > Finished tab,

On my local setup, I am still able to see the spinning dots above the table, which quickly disappear making the rows move a bit up (presumably allowing for the reported miss-clicks).

mzedel:
> On my local setup, I am still able to see the spinning dots above the table, which quickly disappear making the rows move a bit up (presumably allowing for the reported miss-clicks).

Thanks a lot for looking into this @lluiscampos! My understanding from the ticket was so far for it to be the deployment report for a single deployment, but thanks to your comment I went across the codebase and aligned the loader usage on lists to always appear like in the device list (so next to the pagination controls).
While doing this I noticed the release list no longer getting populated on artifact creation, so this should now also be fixed + some minor cleanups to constant usage in timeouts... could you have another look please?

lluiscampos (approved):
I haven't reviewed the actual code.

Spinning up the container and playing a bit with it showed to work as expected :+1:

PR #3800, opened 2023-06-23
MEN-6551 - fix: fixed an issue that prevented re-opening the search result for an unchanged search term

mzedel:
- also made search require 2 letters to start, since the search backend doesn't work reliably with a single letter

Ticket: MEN-6551
Changelog: Title

### DIFF
--- a/src/js/components/common/search.js
+++ b/src/js/components/common/search.js
@@ -35,6 +35,9 @@ const endAdornment = (
   </InputAdornment>
 );
 
+// due to search not working reliably for single letter searches, only start at 2
+const MINIMUM_SEARCH_LENGTH = 2;
+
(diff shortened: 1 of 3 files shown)


lluiscampos (commented):
It might be something off with the setup, but it doesn't seem to work as expected on my demo setup. I have two devices and the search for a unique Device ID returns the two devices (I suspect that it returns _all_ the devices)

image

lluiscampos (changes requested):
It might be something off with the setup, but it doesn't seem to work as expected on my demo setup. I have two devices and the search for a unique Device ID returns the two devices (I suspect that it returns _all_ the devices)

image

lluiscampos:
@mzedel You are totally right - the backend returns all the devices

image
image

lluiscampos (approved):

mzedel:
@aleksandrychev this like too

PR #3804, opened 2023-06-27
feat: made use of unified docs link

mzedel:
to allow getting insights into version adoption to steer docs focus

goes along with #3704


aleksandrychev (approved):
lgtm

PR #3806, opened 2023-06-28
chore: made use of id attribute respecting device identity rendering

mzedel:
- this ensures the device identity setting is also reflected in audit log entries (fyi @chiachenglu)


aleksandrychev (approved):
looks good

PR #3807, opened 2023-06-28
Chore/simplify test setup

mzedel:
~depends on #3804 + #3806~

This primarily changes the component tests to run wrapped with a redux store by default, leading to automatic interactions with the msw backend. This also makes it much easier to have smaller components that access the redux store across the codebase, without the extra work of preparing tests to use the store.

 + moved to broader selector usage as well in order to minimize additional state changes that are otherwise raised during test runs


aleksandrychev (approved):
I didn't review all 121 files, but understood intention and it looks good =)

PR #3810, opened 2023-06-29
feat: made roles selection default to admin to signal what a role omission would result in

mzedel:
Ticket: None
Changelog: None


mzedel:
@aleksandrychev FYI

PR #3864, opened 2023-07-05
Fix/rbac parsing

mzedel:


tests/mockData.js
```diff
@@ -697,6 +788,56 @@ const expectedParsedRoles = {
...
+    editable: false,
+    isCustom: true,
+    uiPermissions: {
+      ...emptyUiPermissions,
+      auditlog: [uiPermissionsById.read.value],
+      deployments: [
+        // we can't assign deployment permissions through the default permission sets yet - TODO: adjust once it exists in the backend
+        // uiPermissionsById.manage.value
+        // uiPermissionsById.deploy.value,
+        // uiPermissionsById.read.value
```
mzedel, line 825: @kjaskiewiczz I suspect it might be possible to assign these permissions to https://github.com/mendersoftware/useradm-enterprise/blob/master/model/rbac_permission_sets.go#L27 - but I'm not sure if this is the intention of the `"Deploy"` `DeploymentsUI` constant (I couldn't find where this is used...)

aleksandrychev (approved):
ack

tests/mockData.js
kjaskiewiczz: UI components are not in use atm; they were part of the original design, but we decided to postpone this part of implementation; the idea here was that the UI components will be included in the permission set objects, and based on them, UI can decide what to show/hide;
when it comes to permissions sets for accessing and creating deployments we have this one - https://github.com/mendersoftware/useradm-enterprise/blob/f9c0ef60e5b58adefd3600cb615e1cb63a78068f/model/rbac_permission_sets.go#L18
and for accessing and deploying configuration we have this one:
https://github.com/mendersoftware/useradm-enterprise/blob/f9c0ef60e5b58adefd3600cb615e1cb63a78068f/model/rbac_permission_sets.go#L20

PR #3926, opened 2023-08-02
fix: limited mapwidget visibility while reporting is still pending

mzedel:
Ticket: None
Changelog: None


merlin-northern (approved):
as a review I also checked that the map does work and show devices according to coordinates sent when hasReporting is 1

mzedel:
@mender-test-bot cherry-pick to:
 * 3.6.x

PR #3931, opened 2023-08-09
fix: fixed an issue that could cause the app to freeze when selecting a release

mzedel:
Ticket: None
Changelog: Title

### DIFF
--- a/src/js/components/deployments/deployment-wizard/softwaredevices.js
+++ b/src/js/components/deployments/deployment-wizard/softwaredevices.js
@@ -11,7 +11,7 @@
 //    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { useDispatch } from 'react-redux';
 import { Link } from 'react-router-dom';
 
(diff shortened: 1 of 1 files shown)


src/js/components/deployments/deployment-wizard/softwaredevices.js
```diff
@@ -218,16 +218,22 @@ export const Software = ({ commonClasses, deploymentObject, releaseRef, releases
...
-  const onReleaseInputChange = inputValue => {
-    setIsLoadingReleases(!releases.length);
-    return dispatch(getReleases({ page: 1, perPage: 100, searchTerm: inputValue, searchOnly: true })).finally(() => setIsLoadingReleases(false));
-  };
+  const onReleaseInputChange = useCallback(
+    inputValue => {
+      setIsLoadingReleases(!releases.length);
+      return dispatch(getReleases({ page: 1, perPage: 100, searchTerm: inputValue, searchOnly: true })).finally(() => setIsLoadingReleases(false));
+    },
+    [dispatch, releases.length]
```
tranchitella, line 235: Only for my understanding: why dispatch is listed as dependency?

tranchitella (approved):

src/js/components/deployments/deployment-wizard/softwaredevices.js
mzedel: Because it gets set through the useDispatch hook above (we now have linting rules for this 👌), so it might potentially change over the component lifetime...

PR #3943, opened 2023-08-14
feat(e2e-tests): added test for search functionality

mzedel:
Ticket: None
Changelog: None


aleksandrychev (commented):
cool! would be also nice to check `demoDeviceName` presence in the search results and absence of other devices. but anyway this is good to go

mzedel:
> cool! would be also nice to check `demoDeviceName` presence in the search results and absence of other devices. but anyway this is good to go

Good point @aleksandrychev - added both 👌

aleksandrychev (approved):

PR #3962, opened 2023-08-28
chore: remove unneeded CSP addresses

oldgiova:
Ticket: SEC-867


mzedel (changes requested):
we still need all the `googletagmanager.com` related ones (per https://developers.google.com/tag-platform/security/guides/csp - also because we can't create a nonce/ hash in our setup) - the rest might be outdated now, but OTOH we only found out during usage for the previous GA version... so let's see 🚀

oldgiova:
Closing this one

PR #4028, opened 2023-09-08
MEN-6455

mzedel:
depends on https://github.com/mendersoftware/deployments/pull/918 to proceed


tranchitella:
@mzedel 

I reviewed the code, at first sight it looks good.

However, one thing didn't work as expected:
1. open an artifact
2. edit tags (e.g. add one)
3. the new tag shows up
4. edit tags again
5. the newly added tag disappears, as it shows the previous set of tags.

Also, filtering by tags doesn't work but I didn't have time to check whether it is a backend issue or front-end one.

mzedel:
> However, one thing didn't work as expected:
Thanks for taking a look and catching this @tranchitella! It should now work as intended 👌

> Also, filtering by tags doesn't work but I didn't have time to check whether it is a backend issue or front-end one.
Once the backend is closer I'll be happy to adjust whatever is not aligned, if you don't want to dig in right now!

tranchitella:
I implemented the filter by tag here: https://github.com/mendersoftware/deployments/pull/923

This PR works fine but there are two things which need to be changed:
* Filter by update type is named `update_type`, not `type`.
* My task required to create a new end-point in `v2`, because of the capitalization of the keys in the payload (see MEN-6439 for details).

Can I kindly ask you to make these changes to this PR?

mzedel:
Thanks for checking @tranchitella - the update type & endpoint should be aligned now... As now all of the release retrieval from the UI should happen through the v2 endpoint I had to switch these too to the newer model and hope I caught all references to uppercase `Artifacts` and `Name`... 
but now we depend on https://github.com/mendersoftware/deployments-enterprise/pull/888 to be deployed along with/ before this (FYI @alfrunes)

tranchitella (approved):
Thank you and sorry for the long delay and ping-pong with the deployments PRs.
It works fine, let me merge the backend components first.

PR #4031, opened 2023-09-12
MEN-6516 - chore: aligned troubleshoot tab with updated design

mzedel:


michaelntech (changes requested):
Small issue - the input at the bottom gets focused by default, so every time I open the troubleshoot tab it puts me to the bottom. It is a bit disorienting
image

mzedel:
Thanks for checking this @michaelntech - the auto focus should now be gone with 44352103eeed210a295733c0fbff29c1dc176e33 👌

michaelntech (changes requested):
Just requesting a couple of small style changes:
image

And perhaps an issue with the terminal and file transfer UI showing when not available:
image

mzedel:
these should now be adjusted @michaelntech - however sometimes the backend seems to report devices as connected even when they are offline, which resulted in a similarly confusing state - this should now also be caught but I would very much appreciate another round of testing from you!

michaelntech (approved):
Looks great, thank you!

PR #4032, opened 2023-09-15
MEN-6717 - fix: aligned env checks with diverged setup instructions for on-prem installs

mzedel:
Ticket: MEN-6717
Changelog: None

### DIFF
--- a/src/js/selectors/index.js
+++ b/src/js/selectors/index.js
@@ -351,10 +351,9 @@ export const getTenantCapabilities = createSelector(
   ) => {
     const canDelta = isEnterprise || plan === PLANS.professional.id;
     const hasAuditlogs = isAuditlogEnabled && (!isHosted || isEnterprise || plan === PLANS.professional.id);
-    const hasDeviceConfig = hasAddons || (isDeviceConfigEnabled && (!isHosted || addons.some(addon => addon.name === 'configure' && Boolean(addon.enabled))));
-    const hasDeviceConnect =
-      hasAddons || (isDeviceConnectEnabled && (!isHosted || addons.some(addon => addon.name === 'troubleshoot' && Boolean(addon.enabled))));
-    const hasMonitor = hasAddons || (isMonitorEnabled && (!isHosted || addons.some(addon => addon.name === 'monitor' && Boolean(addon.enabled))));
+    const hasDeviceConfig = hasAddons || isDeviceConfigEnabled || addons.some(addon => addon.name === 'configure' && Boolean(addon.enabled));
+    const hasDeviceConnect = hasAddons || isDeviceConnectEnabled || addons.some(addon => addon.name === 'troubleshoot' && Boolean(addon.enabled));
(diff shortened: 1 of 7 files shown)


src/js/selectors/index.js
```diff
@@ -340,10 +340,9 @@ export const getTenantCapabilities = createSelector(
   ) => {
     const canDelta = isEnterprise || plan === PLANS.professional.id;
     const hasAuditlogs = isAuditlogEnabled && (!isHosted || isEnterprise || plan === PLANS.professional.id);
-    const hasDeviceConfig = hasAddons || (isDeviceConfigEnabled && (!isHosted || addons.some(addon => addon.name === 'configure' && Boolean(addon.enabled))));
-    const hasDeviceConnect =
-      hasAddons || (isDeviceConnectEnabled && (!isHosted || addons.some(addon => addon.name === 'troubleshoot' && Boolean(addon.enabled))));
-    const hasMonitor = hasAddons || (isMonitorEnabled && (!isHosted || addons.some(addon => addon.name === 'monitor' && Boolean(addon.enabled))));
+    const hasDeviceConfig = hasAddons || isDeviceConfigEnabled || addons.some(addon => addon.name === 'configure' && Boolean(addon.enabled));
+    const hasDeviceConnect = hasAddons || isDeviceConnectEnabled || addons.some(addon => addon.name === 'troubleshoot' && Boolean(addon.enabled));
+    const hasMonitor = hasAddons || isMonitorEnabled || addons.some(addon => addon.name === 'monitor' && Boolean(addon.enabled));
```
tranchitella, line 345: I could be wrong but this change will make "hasDeviceConfig" true on HM independently from the addons status.
Am I right? See: https://github.com/mendersoftware/saas/blob/master/kubernetes/mender_stack/gui/mender-gui.yaml#L87

tranchitella (changes requested):

src/js/selectors/index.js
mzedel: then we can remove these from there I guess...

mzedel: https://github.com/mendersoftware/saas/pull/1069

tranchitella: so... if I get it correctly, you use the env variables for OS and rely on add-ons for ENT.
Right?

mzedel: from now on: yes
the only exception being that `HAVE_ADDONS` flag as I'm not sure where it is used (the slack messages from Marcin mentioned the backend tests, but from searching the backend repos I can't find it)...

tranchitella: @mzedel `DEVICEAUTH_HAVE_ADDONS` and `USERADM_HAVE_ADDONS` maybe?

mzedel: these might be what became of it - in the messages he referred to `HAVE_ADDONS` in particular and wanted it added to the UI... but I'd also be happy to remove it and see if something would actually fail - it doesn't seem to be used anymore for customer communication

mzedel: `HAVE_ADDONS` should now also be removed.

tranchitella:
Reject, we don't want to go down this path.

PR #4033, opened 2023-09-15
MEN-6456

mzedel:


michaelntech (changes requested):
I encountered this error when changing the deployment date filters:
image

The error occurs on changing the "to" date. It would also be good if the dates earlier than the "from" date are disabled in the "to" field

tranchitella (approved):
Tested and it works fine; I went through the commits as well and I couldn't spot anything wrong.

PR #4036, opened 2023-09-18
MEN-6718 - fix: fixed an issue that made the link to target devices in a deployment report unreliable

mzedel:
Ticket: MEN-6718
Changelog: Title


tranchitella:
I reviewed the code and tested it locally (enterprise mode) with all the combinations of deployments, it works fine.

tranchitella (approved):

PR #4048, opened 2023-09-26
MEN-6749 - fix: worked around an issue that would render the ui mostly unclickable due to a dangling background

mzedel:
- the drawer expansion relied on redux values that might be changed while the drawer is not rendered, this seems to hit a mui bug that prevents the drawer background from refreshing if it is no longer needed

Ticket: MEN-6749
Changelog: Title


tranchitella (approved):
Oh, this happened a couple of times to me and I couldn't understand why!
Thanks for fixing it.

PR #4090, opened 2023-10-05
MEN-6768 - tenant token warning

mzedel:
more as a RFC...
Screenshot 2023-10-05 at 10 00 28


tranchitella:
I believe it is a good idea. I'd suggest this text:
```Do not share your organization token with others. Treat this token like a password, as it can be used to request authorization for new devices.```

tranchitella (approved):
From my PoV, this is good to go.
Leaving it to @michaelntech if he wants to adjust the UX.

mzedel:
Merging this since it's been a while since this was opened, we still get tokens sent to us and we can re-adjust later in any case.

PR #4092, opened 2023-10-06
MEN-6600 + MEN-6744 - device list resize fixes + name column resizing

mzedel:


tranchitella (approved):
It works perfectly and I cannot spot anything in the diff.

PR #4104, opened 2023-10-16
SEC-947 - chore: reworked jwt usage

mzedel:


tranchitella (approved):
I tested it in combination with the useradm PR, it works fine.
I reviewed the diff and I couldn't spot issues.

PR #4105, opened 2023-10-16
MEN-6789 - fix: made remote terminal slightly wider to accomodate at least 80 chars

mzedel:


TheMeaningfulEngineer (commented):
CE-231
Just checking if I can link it to the Jira ticket post creation.

tranchitella (approved):

PR #4123, opened 2023-10-31
fix: fixed an issue that prevented storing delta artifact configuration changes

mzedel:
Ticket: None
Changelog: None

### DIFF
--- a/src/js/components/settings/global.test.js
+++ b/src/js/components/settings/global.test.js
@@ -48,7 +48,7 @@ const preloadedState = {
 describe('GlobalSettings Component', () => {
   it('renders correctly', async () => {
     const { baseElement } = render(<Global />, { preloadedState });
-    await act(async () => jest.advanceTimersByTime(TIMEOUTS.fiveSeconds));
+    await act(async () => jest.advanceTimersByTime(TIMEOUTS.refreshDefault));
     await waitFor(() => expect(screen.getByText(/xDelta3/i)).toBeVisible());
     const view = baseElement.firstChild.firstChild;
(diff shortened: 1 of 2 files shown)


src/js/components/settings/global.test.js
```diff
@@ -48,7 +48,7 @@ const preloadedState = {
 describe('GlobalSettings Component', () => {
   it('renders correctly', async () => {
     const { baseElement } = render(<Global />, { preloadedState });
-    await act(async () => jest.advanceTimersByTime(TIMEOUTS.fiveSeconds));
+    await act(async () => jest.advanceTimersByTime(TIMEOUTS.refreshDefault));
```
merlin-northern, line 51: is that on purpose: in `src/js/components/settings/artifactgeneration.js` we change to `TIMEOUTS.fiveSeconds` and in here to the default, why? (mainly curious)

merlin-northern (approved):
tested. works. one idle comment, lgtm, thank you.

src/js/components/settings/global.test.js
mzedel: the `fiveSeconds` are for the initialization, to ensure all the requests for the configuration retrieval have settled to not interpret the absence of a config as a call to set the default values and then save these... in the test we have to wait the `fiveSeconds` + a delta to ensure the initialization processing is over and the `refreshDefault` is 10s - so we should be definitely past the `fiveSeconds` + delta time...

merlin-northern: aha! thank you.

PR #4182, opened 2023-11-06
chore(e2e-tests): aligned user creation steps with updated mui & made use of playwright locators

mzedel:


tests/e2e_tests/integration/06-auditlogs.spec.ts
```diff
@@ -44,7 +44,7 @@ test.describe('Auditlogs', () => {
     const expectedPath = path.join(__dirname, '..', 'test-results', 'diffs', 'terminalSecretContent.png');
     const elementHandle = await page.$(selectors.terminalElement);
     await elementHandle.screenshot({ path: expectedPath });
-    await terminalText.fill(secret);
+    await terminalText.pressSequentially(secret);
```
merlin-northern, line 47: I would add a small sleep right before starting typing, if we are faster than the terminal init from the passwd we can still see the first letters.

merlin-northern (approved):
thank you! (one annoying comment)

PR #4200, opened 2023-11-14
MEN-6793 - fix: made use of newly added information about when a device actually started a deployment

mzedel:
Ticket: MEN-6793
Changelog: Title

### DIFF
--- a/src/js/components/deployments/deployment-report/devicelist.js
+++ b/src/js/components/deployments/deployment-report/devicelist.js
@@ -105,7 +105,7 @@ const deviceListColumns = [
     },
     canShow
   },
-  { key: 'started', title: 'Started', render: ({ device: { created } }) => <MaybeTime value={formatTime(created)} />, sortable: false, canShow },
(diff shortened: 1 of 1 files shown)


src/js/components/deployments/deployment-report/devicelist.js
```diff
@@ -105,7 +105,7 @@ const deviceListColumns = [
     },
     canShow
   },
-  { key: 'started', title: 'Started', render: ({ device: { created } }) => <MaybeTime value={formatTime(created)} />, sortable: false, canShow },
+  { key: 'started', title: 'Started', render: ({ device: { started } }) => <MaybeTime value={formatTime(started)} />, sortable: false, canShow },
```
merlin-northern, line 108: I tested but it does not work. (lets discuss privately)

tranchitella: This fix is correct, let's merge it. We'll follow-up with a dedicated PR once we decide how to update the other labels to make them coherent.

tranchitella (approved):

PR #4201, opened 2023-11-14
test: device tags e2e test.

merlin-northern:
Ticket: MEN-6591
Changelog: None


mzedel:
https://github.com/mendersoftware/gui/blob/master/tests/e2e_tests/integration/03-files.spec.ts#L64-L135 covered already there

merlin-northern:
> https://github.com/mendersoftware/gui/blob/master/tests/e2e_tests/integration/03-files.spec.ts#L64-L135 covered already there

wow. but mine was better ;-))
okey then this task needs to be rejected. I will discuss it.

PR #4218, opened 2023-11-27
fix: group names in the autocomplete

tranchitella:
Changelog: None
Ticket: MEN-6866


TheMeaningfulEngineer (approved):
Approving this so it can be merge to staging where I can test it.

PR #4237, opened 2023-12-08
fix: user settings not fetched after login

mineralsfree:
Ticket: MEN-6882
Changelog: None


TheMeaningfulEngineer (commented):
I've tested this, left a comment in [Jira](https://northerntech.atlassian.net/browse/MEN-6882?focusedCommentId=102532).

tranchitella (approved):

PR #4250, opened 2023-12-15
fix: attributes' grouping in the autocomplete widget

tranchitella:
Changelog: None
Ticket: None


merlin-northern:
tested but no luck. discussion is elsewhere.

merlin-northern:
we are almost there, see the private thread.

merlin-northern (approved):
lgtm, tested.

PR #4254, opened 2023-12-19
fix: detection of on-prem addons relying on tenant's information

tranchitella:
Changelog: None
Ticket: MEN-6717

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -31,7 +31,6 @@ cat >/var/www/mender-gui/dist/env.js <<EOF
     hostedAnnouncement: "$ANNOUNCEMENT",
     isDemoMode: "$DEMO",
     features: {
-      hasAddons: "$HAVE_ADDONS",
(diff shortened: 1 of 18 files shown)


MuchoLucho:
Hello @tranchitella, just to be clear, to test this one, if I update the info from the Tenant in MongoDB, should the whole Mender reflect the change? Or what will be the right step-by-step?

tranchitella:
> Hello @tranchitella, just to be clear, to test this one, if I update the info from the Tenant in MongoDB, should the whole Mender reflect the change? Or what will be the right step-by-step?

Yes, it is correct. You can run enterprise on-prem and change the add-ons for the tenant in MongoDB, the UI will reflect the changes. Similarly, you can run OS and you will only see config and troubleshoot but not monitoring.

oldgiova:
Approving to unblock the 3.7 release: "leap of faith"

oldgiova (approved):

PR #4259, opened 2023-12-20
feat: remove noreferrer from links to mender.io and northern.tech

tranchitella:
Ticket: MEN-6903
Changelog: None

### DIFF
--- a/src/js/components/footer.js
+++ b/src/js/components/footer.js
@@ -46,11 +46,14 @@ export const Footer = () => {
   const { classes } = useStyles();
   return (
     <div className={classes.footer}>
-      {targets.map(({ key, target, title }) => (
-        <a className="clickable" href={target} key={key} target="_blank" rel="noopener noreferrer">
-          {title}
-        </a>
-      ))}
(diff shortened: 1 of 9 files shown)


src/js/components/footer.js
```diff
@@ -46,11 +46,14 @@ export const Footer = () => {
   const { classes } = useStyles();
   return (
     <div className={classes.footer}>
-      {targets.map(({ key, target, title }) => (
-        <a className="clickable" href={target} key={key} target="_blank" rel="noopener noreferrer">
-          {title}
-        </a>
-      ))}
+      {targets.map(({ key, target, title }) => {
+        return (
```
merlin-northern, line 50: this return added on purpose, right? I am having doubts.

tranchitella: Yes, I need it to add the comment (as this is not a single value anymore)

merlin-northern: ah right. sorry for the nosie.

merlin-northern (approved):

PR #4260, opened 2023-12-20
fix: expansion of the session token attribute in my profile

tranchitella:
Changelog: None
Ticket: MEN-6886

### DIFF
--- a/src/js/components/common/expandable-attribute.js
+++ b/src/js/components/common/expandable-attribute.js
@@ -57,7 +57,7 @@ export const ExpandableAttribute = ({
       copy(secondary);
       setSnackbar('Value copied to clipboard');
     }
-    if (!expanded) {
+    if (!expanded && !!onExpansion) {
       onExpansion();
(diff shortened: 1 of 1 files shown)


merlin-northern (approved):

src/js/components/common/expandable-attribute.js
```diff
@@ -57,7 +57,7 @@ export const ExpandableAttribute = ({
       copy(secondary);
       setSnackbar('Value copied to clipboard');
     }
-    if (!expanded) {
+    if (!expanded && !!onExpansion) {
```
merlin-northern, line 60: ok, this is what we want, right, if onExpansion===true, just to double check.

merlin-northern (approved):

src/js/components/common/expandable-attribute.js
tranchitella: yes, we call it only if it is defined

PR #4268, opened 2023-12-27
fix: move cookie-based JWT tokens to the local storage

tranchitella:
If the log in happens through SSO (e.g., OAuth 2.0 or SAML 2) the backend sets the JWT token in a cookie. The UI is then responsible for moving it to the local storage, as happens with the JWT token on a standard email/password log in. This fixes the expiration issues with cookie-based JWT tokens.

Changelog: None
Ticket: MEN-6849

### DIFF
--- a/src/js/auth.js
+++ b/src/js/auth.js
@@ -32,7 +32,16 @@ export const getSessionInfo = () => {
     cleanUp();
     return { ...emptySession };
   }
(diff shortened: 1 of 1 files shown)


oldgiova (approved):
Not an expert but approving, and let's see for staging.

PR #4284, opened 2024-01-10
Hosted Mender rolling release

oldgiova:
Triggers mender helm version bump to
values-staging.yaml and values-prod.yaml in the mender-helm repository

Ticket: MC-6856
Changelog: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -6,6 +6,7 @@ stages:
   - build
   - e2e-test
   - publish
+  - version-bump
 
 variables:
   DOCKER_REPOSITORY: mendersoftware/gui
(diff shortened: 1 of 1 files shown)


mzedel (approved):
lgtm!
just to be sure: the staging trigger also causes a deployment right?
and if so: do you think this will be done within 15min (the trigger time for the staging tests to run)?

oldgiova:
> lgtm! just to be sure: the staging trigger also causes a deployment right? and if so: do you think this will be done within 15min (the trigger time for the staging tests to run)?

Yes, the staging trigger is causing an async deployment; the staging cluster is checking for a new version every ~5min, and the deployment takes ~2min, so we should be inside the 15min.
However, I think [this](https://northerntech.atlassian.net/browse/MC-6867) could be a better approach

PR #4315, opened 2024-02-13
chore: removed the map component due to the dependency license being non-compliant

aleksandrychev:
Ticket: MEN-6962


tranchitella (approved):

kacf (approved):
I think we agreed yesterday that this hasn't been exposed in the UI. Then we don't need a changelog, I think.

kjaskiewiczz (approved):

PR #4317, opened 2024-02-14
feat: explicitly install mender-client4 in bash snippets, for onboarding tutorial and 'Connect a new device'

tranchitella:
Changelog: Title
Ticket: MEN-6976


src/js/helpers.js
```diff
@@ -445,13 +445,15 @@ const getSetupArgs = ({ deviceType = 'generic-armv6', ipAddress, isDemoMode, ten
   return menderSetupArgs;
 };
 
+const installComponents = 'mender-client4 mender-connect mender-configure';
```
lluiscampos, line 448: This list is incomplete. There are two issues here:
* When in hosted, we also aim to installing the commercial components (see `getInstallScriptArgs`), which means that we should include `mender-monitor` in the list for that case
* We are unconditionally using the `--demo` flag for the script itself (see `getInstallScriptArgs`) which means that we would also need to add `mender-configure-demo` and `mender-configure-timezone` unconditionally and `mender-monitor-demo` when on hosted.

:point_up: These makes me think that it is too coupled to have here the list of packages (error prone and difficult to maintain). We need to rethink the acceptance criteria for https://github.com/mendersoftware/mender-dist-packages/pull/419 - Suggestion:
* Having in the script a new flag `--legacy-client` that by default is off, and the GUI can set to on.
* By default it will still work the same way
* We can eventually deprecate the flag whenever all our distros are installing v4 by default.

lluiscampos (changes requested):
This PR does what the JIRA tasks says, but I am afraid it is not good enough. See my comment below and let's discuss it a bit.

src/js/helpers.js
tranchitella: Proposal is `--force-mender-client4`

lluiscampos:
One more thing. This change modifies both onboarding and the "Connect a new device" flow in production. So the commit title (the changelog) should be reworked a bit. For example "feat: explicitly install mender-client4 in bash snippets, for onboarding tutorial and 'Connect a new device'", or just the former part if you want it shorter.

lluiscampos (approved):
:+1:

PR #4348, opened 2024-03-04
MEN-7029: Download log button failing to download file

mineralsfree:
Previous implementation didn't send authorization header 

Ticket: MEN-7029
Changelog: None

### DIFF
--- a/src/js/components/devices/device-details/deployments.js
+++ b/src/js/components/devices/device-details/deployments.js
@@ -19,8 +19,10 @@ import { Button, Table, TableBody, TableCell, TableHead, TableRow, buttonClasses
 import { makeStyles } from 'tss-react/mui';
 
 import { deploymentsApiUrl, getDeviceDeployments, resetDeviceDeployments } from '../../../actions/deploymentActions';
+import { getToken } from '../../../auth.js';
 import { deploymentDisplayStates, deploymentStatesToSubstates } from '../../../constants/deploymentConstants';
 import { DEVICE_LIST_DEFAULTS } from '../../../constants/deviceConstants';
+import { createDownload } from '../../../helpers.js';
 import Confirm from '../../common/confirm';
 import Pagination from '../../common/pagination';
 import { MaybeTime } from '../../common/time';
@@ -67,9 +69,17 @@ const columns = [
   {
     content: '',
     key: 'log',
-    Component: ({ deviceDeployment: { id, log, deviceId } }) =>
+    Component: ({ deviceDeployment: { id, log, deviceId }, token }) =>
       log && (
-        <Button component="a" href={`${window.location.origin}${deploymentsApiUrl}/deployments/${id}/devices/${deviceId}/log`} target="_blank">
+        <Button
+          onClick={() =>
+            createDownload(
+              ` ${window.location.origin}${deploymentsApiUrl}/deployments/${id}/devices/${deviceId}/log`,
+              `device_${deviceId}_deployment_${id}.log`,
+              token
+            )
+          }
+        >
           Log
         </Button>
       )
@@ -77,6 +87,7 @@ const columns = [
 ];
 
 const History = ({ className, items, page, perPage, setPage, setPerPage, total }) => {
+  const token = getToken();
   const onChangeRowsPerPage = perPage => {
     setPage(1);
     setPerPage(perPage);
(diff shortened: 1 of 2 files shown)


aleksandrychev:
@mineralsfree this change doesn't look like a proper fix. what's wrong here, when you click a log link then API endpoint authorization does not happen?

mineralsfree:
> @mineralsfree this change doesn't look like a proper fix. what's wrong here, when you click the log link then API endpoint authorization does not happen?

It opens a link in a new tab trying to fetch a file, but since JWT token is not passed in HTTP header, the request is not authorized

aleksandrychev:
> > @mineralsfree this change doesn't look like a proper fix. what's wrong here, when you click the log link then API endpoint authorization does not happen?
> 
> It opens a link in a new tab trying to fetch a file, but since JWT token is not passed in HTTP header, the request is not authorized

JWT also set in cookies, and API in this case should also recognise it. @tranchitella @alfrunes please confirm

alfrunes:
> JWT also set in cookies, and API in this case should also recognise it. @tranchitella @alfrunes please confirm

Correct, the backend uses the `JWT` cookie as the session token.

mineralsfree:
Then it seems the problem is that the cookie is not saved/sent on frontend, I cannot see anything similar to JWT or other auth cookie in dev tools after I log in. @aleksandrychev @alfrunes @tranchitella

aleksandrychev:
yes, right. now UI moves token from cookies to localStorage.  @mineralsfree you can use 
```cookies.set('JWT', token, { path: '/', secure: true, sameSite: 'strict', maxAge: 5 });``` before downloading, as we do for other downloads. maxAge will set cookies life time for a 5 seconds.

File: src/js/components/devices/device-details/deployments.js [hunk 1]
```diff
@@ -49,34 +53,58 @@ const EmptyState = ({ isFiltered }) => (
...
-        <Button component="a" href={`${window.location.origin}${deploymentsApiUrl}/deployments/${id}/devices/${deviceId}/log`} target="_blank">
-          Log
-        </Button>
-      )
-  }
-];
-
 const History = ({ className, items, page, perPage, setPage, setPerPage, total }) => {
+  const { token } = useSelector(getCurrentSession);
+  const columns = [
```
aleksandrychev, line 58: no need to move columns inside the history component, you can pass token to the component

```
{columns.map(({ key, Component }) => (
                <TableCell key={`${item.id}-${key}`}>
                  <Component deviceDeployment={item} token={token} />
                </TableCell>
              ))}
```

and get it later 
```Component: ({ deviceDeployment: { id, log, deviceId }, token }) =>```

File: src/js/components/devices/device-details/deployments.js [hunk 2]
```diff
@@ -49,34 +53,58 @@ const EmptyState = ({ isFiltered }) => (
...
-      log && (
-        <Button component="a" href={`${window.location.origin}${deploymentsApiUrl}/deployments/${id}/devices/${deviceId}/log`} target="_blank">
-          Log
-        </Button>
-      )
-  }
-];
-
 const History = ({ className, items, page, perPage, setPage, setPerPage, total }) => {
+  const { token } = useSelector(getCurrentSession);
```
aleksandrychev, line 57: please use getToken() from auth

src/js/components/devices/device-details/deployments.js
```diff
@@ -49,34 +53,58 @@ const EmptyState = ({ isFiltered }) => (
...
+    { content: 'Device status', key: 'status', Component: ({ deviceDeployment: { status } }) => status },
+    {
+      content: '',
+      key: 'log',
+      Component: ({ deviceDeployment: { id, log, deviceId } }) =>
+        log && (
+          <Button
+            component="a"
+            href={`${window.location.origin}${deploymentsApiUrl}/deployments/${id}/devices/${deviceId}/log`}
+            onClick={() =>
```
aleksandrychev, line 93: there is createDownload function in the helpers.js that makes cookie set

```
export const createDownload = (target, filename, token) => {
  let link = document.createElement('a');
  link.setAttribute('href', target);
  link.setAttribute('download', filename);
  link.style.display = 'none';
  document.body.appendChild(link);
  cookies.set('JWT', token, { path: '/', secure: true, sameSite: 'strict', maxAge: 5 });
  link.click();
  document.body.removeChild(link);
};
```

aleksandrychev (changes requested):

aleksandrychev:
@mineralsfree please update snapshot and then good to go

aleksandrychev (approved):

aleksandrychev:
@tranchitella backport?

tranchitella:
> @tranchitella backport?

no need to backport this fix.

tranchitella (approved):

PR #4354, opened 2024-03-12
MEN-7041: Annoying message on Basic and Pro plan when config add-on disabled

mineralsfree:
Ticket: MEN-7041
Changelog: None

In previous implementation, add-on was considered enabled whenever the plan is not Enterprise. Could not reproduce locally, so could not test it, but this fix should help


aleksandrychev (approved):

olehermanse:
@aleksandrychev said this should be merged.

PR #4381, opened 2024-04-25
feat: use check_in_time from inventory to calculate if the device is offline

kjaskiewiczz:
Ticket: MEN-7202


kjaskiewiczz:
The idea behind this change is to use check_in_time from inventory (propagated by the deviceauth once a day - https://github.com/mendersoftware/deviceauth/pull/712) to check which devices are offline in the dashboard view and in the device list view. In the device details view we want to use check_in_time from deviceauth directly.
I ended up with something working (at least when I was checking locally), but I have two issues:
- test are failing and I'm not sure how to fix them;
- I don't know if my implementation is acceptable;
Could someone with more frontend knowledge take it over? maybe @aleksandrychev or @mineralsfree ?
FYI @merlin-northern @alfrunes @tranchitella

tranchitella:
Superseded by https://github.com/mendersoftware/gui/pull/4401

PR #4382, opened 2024-04-30
MEN-7061: Accepted tooltip disappeared right after device is accepted

mineralsfree:
Ticket: MEN-7061
Changelog: None

### DIFF
--- a/src/js/components/devices/authorized-devices.js
+++ b/src/js/components/devices/authorized-devices.js
@@ -265,7 +265,6 @@ export const Authorized = ({
     if (!acceptedCount) {
       return;
     }
-    dispatch(advanceOnboarding(onboardingSteps.DEVICES_ACCEPTED_ONBOARDING));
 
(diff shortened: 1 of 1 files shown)


src/js/components/devices/authorized-devices.js
```diff
@@ -265,7 +265,6 @@ export const Authorized = ({
     if (!acceptedCount) {
       return;
     }
-    dispatch(advanceOnboarding(onboardingSteps.DEVICES_ACCEPTED_ONBOARDING));
```
aleksandrychev, line 268: did you check if there is another place doing dispatch(advanceOnboarding(onboardingSteps.DEVICES_ACCEPTED_ONBOARDING));?
otherwise step might not be changed

mineralsfree: Yes, in order to proceed, the user should click a button on the tooltip with a hint, that says "Yes, let's deploy!", or, alternatively click on "Dismiss the  tutorial".

aleksandrychev (approved):

mzedel (approved):

PR #4399, opened 2024-05-09
fix: use the two-steps log in on hosted

tranchitella:
Changelog: None
Ticket: MEN-6823

### DIFF
--- a/src/js/components/login/login.test.js
+++ b/src/js/components/login/login.test.js
@@ -47,6 +47,13 @@ describe('Login Component', () => {
     const { rerender } = render(ui, { preloadedState });
     await act(async () => {
       await user.type(screen.getByLabelText(/your email/i), 'something-2fa@example.com');
+    });
+    await act(async () => {
+      await user.click(screen.getByRole('button', { name: /Log in/i }));
+    });
+    expect(loginSpy).toHaveBeenCalled();
+    await waitFor(() => rerender(ui));
+    await act(async () => {
       await user.type(screen.getByLabelText(/password/i), 'mysecretpassword!123');
     });
     expect(await screen.findByLabelText(/Two Factor Authentication Code/i)).not.toBeVisible();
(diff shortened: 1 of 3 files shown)


src/js/components/login/login.test.js
```diff
@@ -47,6 +47,13 @@ describe('Login Component', () => {
     const { rerender } = render(ui, { preloadedState });
     await act(async () => {
       await user.type(screen.getByLabelText(/your email/i), 'something-2fa@example.com');
+    });
+    await act(async () => {
+      await user.click(screen.getByRole('button', { name: /Log in/i }));
+    });
```
mzedel, line 53:
```suggestion
    await act(async () => {
      await user.type(screen.getByLabelText(/your email/i), 'something-2fa@example.com');
      await user.click(screen.getByRole('button', { name: /Log in/i }));
    });
```

the call to `act` can take a bit of time to tear down and cause weird problems in test execution, thus the RTL recommendation is to bundle them similar to an active phase of a user interacting with a page

mzedel (approved):

src/js/components/login/login.test.js
tranchitella: This code fails with:

`
  ● Login Component › works as intended

    Unable to perform pointer interaction as the element has `pointer-events: none`:

    BUTTON(label=Log in)

      48 |     await act(async () => {
      49 |       await user.type(screen.getByLabelText(/your email/i), 'something-2fa@example.com');
    > 50 |       await user.click(screen.getByRole('button', { name: /Log in/i }));
         |       ^
      51 |     });
      52 |     expect(loginSpy).toHaveBeenCalled();
      53 |     await waitFor(() => rerender(ui));
`

tranchitella: the pipeline is passing and you approved the PR, @mzedel.
If I need to rework this particular section, let me know and I will follow up with a dedicated PR.

mzedel: Thanks for the explanation! The recommendation would be to be explicit and do a
`await waitFor(() =>  expect(screen.getByRole('button', { name: /log in/i })).not.toBeDisabled());`
or often an
`await waitFor(() => rerender(ui));`
also works...

but OTOH - if it works I won't mind... and absolutely: once approved always merge ahead!

PR #4414, opened 2024-05-15
MEN-7192 - enable password reset during user creation

mzedel:


mineralsfree (approved):
Changes look good. 

Side comment is that `FormCheckbox`'s prop naming is a bit misleading with `id` being later used as a input name

mzedel:
Good point, I'll create a follow up ticket to align these to `name` 👌

PR #4415, opened 2024-05-15
MEN-7261 - fix: cleaned up device deployment history to focus on the device instead of deployments

mzedel:
Ticket: MEN-7261
Changelog: None


mineralsfree (approved):
Legitimate

PR #4420, opened 2024-05-22
MEN-7030 - e2e tests for release tags RBAC

mzedel:


tests/e2e_tests/integration/08-rbac.spec.ts
```diff
@@ -51,60 +58,109 @@ test.describe('RBAC functionality', () => {
...
-    await page.getByPlaceholder(/email/i).fill(`limited-${username}`);
-    await page.getByPlaceholder(/Password/i).click();
-    await page.getByPlaceholder(/Password/i).fill(password);
-    if (['enterprise', 'staging'].includes(environment)) {
-      await page.getByRole('combobox', { name: /admin/i }).click();
-      // first we need to deselect the default admin role
-      await page.getByRole('option', { name: 'Admin' }).click();
-      await page.getByRole('option', { name: 'testRole' }).click();
+  test('allows role creation for release tags', async ({ environment, loggedInPage: page }) => {
+    test.skip(!['enterprise', 'staging'].includes(environment));
```
merlin-northern, line 66: does that mean I will not be able to run this test without modifications here on my local setup?

tests/e2e_tests/utils/commands.ts
```diff
@@ -288,3 +288,18 @@ export const compareImages = (expectedPath, actualPath, options = { threshold: 0
...
+  const request = await axios.put(`${baseUrl}api/management/v2/deployments/deployments/releases/${releaseName}/tags`, [tag], {
+    headers: {
+      Authorization: `Bearer ${token}`,
+      'Content-Type': 'application/json'
+    },
+    httpsAgent: new https.Agent({ rejectUnauthorized: false })
+  });
+
+  if (request.status >= 300) {
+    throw 'oh no';
```
merlin-northern, line 302: how about some logs to the console here?

tests/e2e_tests/utils/commands.ts
```diff
@@ -288,3 +288,18 @@ export const compareImages = (expectedPath, actualPath, options = { threshold: 0
   const pass = usePercentage ? (numDiffPixels / (width * height)) * 100 < threshold : numDiffPixels < threshold;
   return { pass, numDiffPixels };
 };
+
+export const tagRelease = async (releaseName: string, tag: string, baseUrl: string, token: string) => {
+  const request = await axios.put(`${baseUrl}api/management/v2/deployments/deployments/releases/${releaseName}/tags`, [tag], {
```
merlin-northern, line 293: this URL is not defined anywhere?

merlin-northern (approved):
looks good -- I honestly (and I am sorry) did not have the time to check the test logic in detail.

tests/e2e_tests/integration/08-rbac.spec.ts
mzedel: You should only need to pass a `--enterprise` to the run script or set `TEST_ENVIRONMENT=enterprise` if you want to run them with the playwright goodness against your local setup 👌
https://github.com/mendersoftware/gui/blob/master/.gitlab-ci.yml#L160

tests/e2e_tests/utils/commands.ts
mzedel: Well it is in the app code, but here the straight API calls should remain the exception - once I get to reuse the URL I would put it in a const 👌

PR #4422, opened 2024-05-23
MEN-7144 - configure SSO on user creation 

mzedel:
Not possible to add if a password related input has input:
Screenshot 2024-05-24 at 01 43 08

Screenshot 2024-05-24 at 15 34 43
Screenshot 2024-05-24 at 15 35 01

### DIFF
--- a/src/js/components/common/forms/formcheckbox.js
+++ b/src/js/components/common/forms/formcheckbox.js
@@ -16,7 +16,7 @@ import { Controller } from 'react-hook-form';
 
 import { Checkbox, FormControlLabel } from '@mui/material';
 
-export const FormCheckbox = ({ className, control, id, handleClick, style, label, required }) => (
+export const FormCheckbox = ({ className, control, disabled, id, handleClick, style, label, required }) => (
   <Controller
     name={id}
     rules={{ required }}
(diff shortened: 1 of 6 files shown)


mineralsfree (approved):
Legitimate

mzedel:
@mineralsfree changed this slightly to only disable the checkbox and not hide it after chatting with @michaelntech - could you take another look please?

src/js/components/common/forms/formcheckbox.js
```diff
@@ -25,7 +25,9 @@ export const FormCheckbox = ({ className, control, id, handleClick, style, label
       return (
         <FormControlLabel
           className={className}
-          control={<Checkbox name={id} onClick={handleClick} checked={value} style={style} color="primary" onChange={() => onChange(!value)} />}
+          control={
+            <Checkbox {...checkboxProps} name={id} onClick={handleClick} checked={value} style={style} color="primary" onChange={() => onChange(!value)} />
```
mineralsfree, line 29: I would avoid props spreading as it damages readability (taking into account we don't use typescript)

mzedel: The spread aligns with [upstream/ mui](https://github.com/mui/material-ui/blob/master/packages/mui-material/src/Checkbox/Checkbox.js) though 😉 - should now be just the `disabled` prop 👌

PR #4427, opened 2024-05-28
MEN-7083 - show user id & allow adding user by id to tenant

mzedel:
User ID visible in "My profile":
Screenshot 2024-05-28 at 21 31 29
- copiable on click similar to the organization info

Screenshot 2024-05-28 at 21 31 16
"Create new user" is now "Add new user"
Screenshot 2024-05-28 at 21 32 35
if a UUID is entered, the user creation form shrinks to focus on adding the user...
However:
If the value entered is not a valid UUID, the form will only focus on email entry:
Screenshot 2024-05-28 at 21 38 13


michaelntech (approved):

mzedel:
passing pipeline is [here](https://gitlab.com/Northern.tech/Mender/gui/-/pipelines/1317630858) - unfortunately the notification seems to be hanging, merging.

PR #4429, opened 2024-05-29
MEN-7221 - wider log viewer & friendlier log file names

mzedel:
With this an example deployment log would look like this:
`deployment-log-7799fe23-35a0-4494-bd3c-251f7821127f-mender-demo-artifact-2.7.0-2021-07-08T14_38_35.431Z.log`
and the deployment log viewer would look like:
Screenshot 2024-05-29 at 09 52 49


mineralsfree (approved):
:+1:

PR #4441, opened 2024-06-05
fix: Update grammar to be more inclusive

michaelntech:
Ticket: None
Changelog: None


mzedel (approved):

mzedel:
strange that the e2e test fails, while it passed [here](https://gitlab.com/Northern.tech/Mender/gui/-/jobs/7013180436) - I will take care of it 👌

PR #4443, opened 2024-06-05
MEN-7288 - feat/threshold migration

mzedel:


estenberg:
Looks good, just would clarify the "What's changing" part. In particular what does "detected" and "any contact with Mender" mean? I would perhaps make it a bit more technical so that it's more clear. Those people reading are pretty technical anyway. :)

mzedel:
@estenberg thank you for looking into this! Would the following be enough?:
> offline devices can only be filtered for after at least one full day without any API request to the Mender backend

estenberg:
Better, but API request can mean many things, such as the user or UI making an API request..

Perhaps "For filtering / listing which devices are offline, the data filtered on may be up to one day old. In other words, if the device has been offline for a while and only recently contacted the Mender server, it may still be listed as offline until the data filtered on is updated (within 1 day)."

estenberg:
Overall I feel this was a convoluted change (long time ago), that focused more on our infra performance than how it affected the user. It is what it is now, but we need to do better next time.

mzedel:
> Better, but API request can mean many things, such as the user or UI making an API request..

then maybe: _offline devices can only be filtered for after at least one full day without any API request by a device to the Mender backend_ 🙃 

> Perhaps "For filtering / listing which devices are offline, the data filtered on may be up to one day old. In other words, if the device has been offline for a while and only recently contacted the Mender server, it may still be listed as offline until the data filtered on is updated (within 1 day)."

Afaik it is a bit the other way around: if the device has been offline for a few hours it won't be listed as offline before 1 day has passed, whereas if it was offline for a while, the device coming back online will be visible right away. Here a device coming back online means: any request to the API by a device.

estenberg:
I am so confused by this. :-/

estenberg:
Okey, let me try again:

"If a device does not connect with the server, it may take up to 1 day until it is shown as an offline device in the UI."

src/js/components/common/dialogs/startupnotification.js
```diff
@@ -0,0 +1,95 @@
...
+import { DEVICE_ONLINE_CUTOFF } from '../../../constants/deviceConstants';
+import { isDarkMode } from '../../../helpers';
+import { getUserSettings } from '../../../selectors';
+import { useDebounce } from '../../../utils/debouncehook';
+
+const OfflineThresholdContent = () => (
+  <>
+    In our continuous efforts to enhance performance and to ensure the stability of our service, we have made adjustments to how granular device connectivity
+    can be checked.
+    <h4>What&apos;s Changing:</h4>
```
michaelntech, line 32: Change to lowercase "c"

michaelntech (approved):

PR #4447, opened 2024-06-12
MEN-7316 - made SAML tests rely on own SAML idp

mzedel:


tests/e2e_tests/utils/commands.ts
```diff
@@ -12,6 +12,7 @@
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
 import { BrowserContext, Page } from '@playwright/test';
+import { runServer } from '@sidewinder1138/saml-idp';
```
mineralsfree, line 15: Do we benefit from using fork instead of original library?

mzedel: yes - mostly because of [this commit](https://github.com/Sidewinder1138/saml-idp/commit/ba217b946d01eb857ac3dad8c154c887422f13e0)... I don't expect this to receive updates, the original hasn't seen any real movement either - but maintaining a private copy was not something I would want to take on just for the handful of tests 🤷‍♂️

mineralsfree: makes sense :+1:

mineralsfree (approved):

PR #4451, opened 2024-06-13
Revert: "revert: Revert "chore: introduced typescript support""

mzedel:
This reverts commit bbc23a48f68e70c91ff11c3088543fea14d86df3.


tsconfig-test.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "noEmit": true,
+    "skipLibCheck": true
+  },
+  "extends": "./tsconfig.json",
+  "exclude": ["src/js/api/types/Settings.ts"]
```
mineralsfree, line 7: why this file excluded?

tsconfig-test.json
mineralsfree, line 1: do we need this file?

mineralsfree (approved):

tsconfig-test.json
mzedel: Quite sure, see the other comment ☺️

tsconfig-test.json
mzedel: ah well it seems to be not to be working as I expected after all: https://github.com/microsoft/TypeScript/issues/7432#issuecomment-193891886

mzedel: hmm then I will remove the extra file and... adjust the type generation to rewrite the `2fa` name? or exclude the file from the barrel file? or... skip checking these types? all not exciting 😬

mzedel: now I ended up adjusting the type generation to ignore the include of the `Settings.ts` and not even try to resolve the import... So... what do you think about 73dc87ce98c316fac64600ea463c764c489fa5e2 ? tolerable or too risky?

mineralsfree: I think it's not perfect, but acceptable, maybe we should rename `2fa` to `twoFA`  at some point?

mzedel: Yeah, that would be good longer term! Unfortunately I think it will break the API spec/ require a version bump just for this - so likely little appetite in the backend 😬... I'll create a ticket 👌

PR #4454, opened 2024-06-17
MEN-7336 - feat: let device details remain open when adding the device to a group

mzedel:
Ticket: MEN-7336
Changelog: Title


mineralsfree (approved):
:+1:

PR #4456, opened 2024-06-17
MEN-7334 - fix: fixed an issue that would prevent promoting a device to a gateway device

mzedel:
Ticket: MEN-7334
Changelog: Title


mineralsfree (approved):
:+1:

PR #4460, opened 2024-06-20
fix staging tests

mzedel:
passing pipeline [here](https://gitlab.com/Northern.tech/Mender/gui/-/pipelines/1361601839).
@oldgiova pinged you for [the Hetzner changes](https://github.com/mendersoftware/gui/pull/4460/commits/4ded5ecbaed267ee055ec97ebb5010f0d3b7c73c).

Notable commits:
829d436f5ba514c7b4bd7a25a84df4a0dcea4b09 
eca2ec74d0d17e560dfa0650161792e41ec8d968

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -276,7 +276,7 @@ build:docker-multiplatform:
   # by default alvaldi-gui uses the unprivileged mode
   script:
     - if [[ ! $CI_COMMIT_REF_NAME =~ "(main|master|staging|production|feature-.+|\d+\.\d+\..+)" ]]; then
-    -   export MULTIPLATFORM_PLATFORMS='linux/amd64'
+    - export MULTIPLATFORM_PLATFORMS='linux/amd64'
     - fi
     - echo "building ${CI_PROJECT_NAME} for ${DOCKER_BUILD_SERVICE_IMAGE}"
     - docker context create builder
@@ -345,8 +345,6 @@ build:docker-multiplatform:
     - mkdir logs && cp /root/.npm/_logs/* logs/
   only:
     - staging
-  when: delayed
-  start_in: 15 minutes
   artifacts:
     expire_in: 2w
     paths:
@@ -360,16 +358,22 @@ test:staging-deployment:chrome:
   extends: .template:test:staging-deployment
   script:
     - npm run test-ci
+  when: delayed
+  start_in: 1 minutes
 
 test:staging-deployment:firefox:
   extends: .template:test:staging-deployment
   script:
     - npm run test-ci -- --browser=firefox
+  when: delayed
+  start_in: 25 minutes
 
 test:staging-deployment:webkit:
   extends: .template:test:staging-deployment
   script:
     - npm run test-ci -- --browser=webkit
+  when: delayed
+  start_in: 15 minutes
 
 .template:publish:tests:
   stage: publish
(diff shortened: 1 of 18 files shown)


.gitlab-ci.yml
```diff
@@ -316,8 +323,8 @@ build:docker-multiplatform:
       - baseImage.tar
 
 .template:test:staging-deployment:
-  image: mendersoftware/mender-test-containers:gui-e2e-testing
-  stage: .post
+  image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/mendersoftware/mender-test-containers:gui-e2e-testing
+  stage: test
   services:
     - docker:dind
```
oldgiova, line 329:
```suggestion
  services:
    - name: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/docker:dind
      alias: docker
```

.gitlab-ci.yml
```diff
@@ -138,6 +142,8 @@ test:licenses:
       junit:
         - junit/results.xml
     when: always
+  tags:
+    - hetzner-amd-beefy-privileged
```
oldgiova, line 146: The privileged runner is currently using `privileged=true` and the docker executor, hence is critical and should be used for internal jobs or protected branches... [See the references](https://docs.gitlab.com/runner/security/#usage-of-docker-executor).

Suggestion: keep using a standard GCP runner, and switch to Hetzner when we can find a suitable alternative (https://northerntech.atlassian.net/browse/QA-739)

```suggestion
    - mender-qa-worker-generic-light
```

oldgiova (commented):
Really nice! Just requesting a security change: the hetzner-amd-beefy-privileged is currently using a plain docker executor, and not the more secure sysbox executor, to woraround an issue in the mender-test-container pipeline.
My goal in the short term is to provide a suitable and working non-docker privileged and secure executor (sysbox/podman) for this kind of needs, but this is not yet ready.
As a suggestion, please keep using GCP instead of the `hetzner-amd-beefy-privileged`, if the `hetzner-amd-beefy` runner is failing.

.gitlab-ci.yml
mzedel: let me try this again, it didn't work a few pushes ago, but the tests have changed quite a bit 🙈

.gitlab-ci.yml
```diff
@@ -343,33 +351,37 @@ build:docker-multiplatform:
...
+  #   - staging
   artifacts:
     expire_in: 2w
     paths:
       - tests/e2e_tests/test-results
       - logs
     when: always
   tags:
-    - mender-qa-worker-generic-light
+    - hetzner-amd-beefy-privileged
```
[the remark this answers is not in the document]
mzedel: Oh ok, would it be ok to keep the staging tests on Hetzner (since they would only run on the staging branch) and just change the regular e2e tests?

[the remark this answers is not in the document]
oldgiova: > Oh ok, would it be ok to keep the staging tests on Hetzner (since they would only run on the staging branch) and just change the regular e2e tests? The main reason is basically the webkit flavour of the tests requiring more power than light has and if we can run them all on Hetzner that wouldn't need an override 🙃.

That would be great!

tranchitella (approved):

PR #4478, opened 2024-07-11
fix: Use same region default for signup as login

alfrunes:


src/js/components/login/signup.js
```diff
@@ -78,7 +78,12 @@ export const Signup = () => {
   const [redirectOnLogin, setRedirectOnLogin] = useState(false);
   const [captchaTimestamp, setCaptchaTimestamp] = useState(0);
   const [recaptcha, setRecaptcha] = useState('');
-  const [location, setLocation] = useState(locations.us.key);
+  const [location, setLocation] = useState(
+    Object.entries(locations).reduce(
+      (accu, [key, value]) => ([`staging.${value.location}`, value.location].includes(window.location.hostname) ? key : accu),
+      locations.us.key
+    )
+  );
```
alfrunes, line 86: FYI I copied this from https://github.com/alfrunes/gui/blob/8546ea0428965bcb363c5ef6cc4c4568e7d5f4a2/src/js/components/login/login.js#L102-L105

alfrunes: It's an exception for staging. It's just returning `eu` or `us` depending on the domain in the URL bar (falling back on `us`.

kjaskiewiczz (approved):

alfrunes:
I merged it as is, I'll leave it to @mzedel to clean up the code duplication.

PR #4480, opened 2024-07-11
test: Fix fragile end-to-end test by parsing downloaded artifact

alfrunes:
The test is sensitive to updates to the demo artifact (even if it's just the mtime of the tar headers that changed. This commit changes the assertion by parsing the archive and checking the artifact name in the header info.


kjaskiewiczz (approved):
nice!

mzedel:
This actually was brittle by design - as in the past the changes to the demo-artifact caused misalignment issues (iirc during either the update modules intro or the synchronized updates)...
https://github.com/mendersoftware/gui/pull/4494 - to restore the sensitivity and make it less aware of the Mender Artifact internals 👌

PR #4485, opened 2024-07-31
Fix/signup default

mzedel:


alfrunes (commented):
I find this behavior very confusing: If you sign up on one domain (`eu.hosted.mender.io`), you sign up to a different domain (`hosted.mender.io`) by default. This feels like something that requires explicit action. Maybe I'm the only one?

@michaelntech What do you think?

michaelntech:
> I find this behavior very confusing: If you sign up on one domain (`eu.hosted.mender.io`), you sign up to a different domain (`hosted.mender.io`) by default. This feels like something that requires explicit action. Maybe I'm the only one?
> 
> @michaelntech What do you think?

I think I agree, it would be a bit confusing. 

How about this suggestion – we allow sign ups only through `hosted.mender.io`? So `eu.hosted.mender.io/signup` redirects to `hosted.mender.io/signup`. If the user chooses EU region from the dropdown, they are redirected to the EU UI on completing signup.

mzedel:
https://www.figma.com/design/IPQOCLADVQgk4bfAR2wbsz?node-id=35-3855#303545681 - afair also had the intention to prevent adding extra users to EU as they might be billed differently/ capacity was not clear...
IMO the confusion is a valid concern, but we know our users are regularly confused by logging in into a different instance than they signed up for, but afaik haven't received any complaints about signing up to the wrong region.

alfrunes:
> [...] but we know our users are regularly confused by logging in into a different instance than they signed up for [...]

Could this be a symptom of the behavior I described above?

mzedel:
> > [...] but we know our users are regularly confused by logging in into a different instance than they signed up for [...]
> 
> Could this be a symptom of the behavior I described above?

The reports mention signing in to US but not landing in their EU tenant - similar to logging in to your AWS account and change the region.

PR #4505, opened 2024-08-16
MEN-7458 - fix: prevented commercial client components are only selected when plan/ addon accessible

mzedel:
Ticket: MEN-7458
Changelog: Title


src/js/helpers.js
```diff
@@ -424,10 +424,11 @@ export const standardizePhases = phases =>
     return standardizedPhase;
   });
 
-const getInstallScriptArgs = ({ isHosted, isPreRelease }) => {
+const getInstallScriptArgs = ({ isHosted, isPreRelease, hasMonitor }) => {
   let installScriptArgs = '--demo';
   installScriptArgs = isPreRelease ? `${installScriptArgs} -c experimental` : installScriptArgs;
-  installScriptArgs = isHosted ? `${installScriptArgs} --commercial --jwt-token $JWT_TOKEN` : installScriptArgs;
+  installScriptArgs = isHosted && hasMonitor ? `${installScriptArgs} --commercial` : installScriptArgs;
```
lluiscampos, line 430: This is okey, but I think that linking "hasMonitor" to "commercial flag" it is short-lived: what will happen when we have a different commercial add-on?

A non-beautiful alternative is to list the package(s) directly
```suggestion
  installScriptArgs = isHosted && hasMonitor ? `${installScriptArgs} mender-monitor` : installScriptArgs;
```

and when demo:

```suggestion
  installScriptArgs = isHosted && hasMonitor ? `${installScriptArgs} mender-monitor mender-monitor-demo` : installScriptArgs;
```

But I also see it quite messy.

As a further idea, we could rework the UX so that this is done in two different parts: you first are setting up the device, only with OS components, and later from another panel, tooltip, whatever you can add a commercial add-on like mender-monitor.

lluiscampos (approved):
It works as-is. But see my comment below for some brainstorming

src/js/helpers.js
mzedel: While I do agree that it could be short lived, so far it has lasted 3 years and no additional component was added.
If it was, I would add an abstraction in the UI that collected the "commercial" components a user has access to. Combining this with the demo package would IMO complicate things even more and for ease of use I would steer away from adding an additional step.

PR #4506, opened 2024-08-19
ci: Set correct dependenciy for staging-deployment tests

alfrunes:


mzedel (approved):
Also good, but not enough I think... probably needs #4507

alfrunes:
How about this https://github.com/mendersoftware/gui/pull/4506/commits/098195352eb4b37525ccce81b96bb9d0deffc21a ?
Adding the job to `.post` will not include these tests in the Github status.

PR #4537, opened 2025-05-07
:building_construction: Upgrade dependencies

alfrunes:
Not sure if the `node` major version bump is affecting us, but it was the closest supported version I could find.

### DIFF
--- a/tests/e2e_tests/integration/05-deviceDetails.spec.ts
+++ b/tests/e2e_tests/integration/05-deviceDetails.spec.ts
@@ -33,6 +33,8 @@ test.describe('Device details', () => {
     await page.click(`text=/inventory/i`);
     expect(await page.isVisible(`css=.expandedDevice >> text=Linux`)).toBeTruthy();
     expect(await page.isVisible(`css=.expandedDevice >> text=mac`)).toBeTruthy();
+    const expandedDevice = page.locator(`css=.expandedDevice`);
+    await expandedDevice.getByRole('tab', { name: /software/i }).click();
     expect(await page.isVisible(`css=.expandedDevice >> text=${demoDeviceName}`)).toBeTruthy();
   });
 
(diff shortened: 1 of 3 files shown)


mzedel (approved):

alfrunes:
@mzedel Could you double-check that what I'm doing with e2e tests make sense? Or am I fixing for a regression?

mzedel (approved):
Yep, that's 👍  @alfrunes (there was a similar [change in the monorepo](https://github.com/mendersoftware/mender-server/commit/96315790e4fd80639d964ec728e843d9efc41019#diff-4f7a5b7bad1eb81631a49c2182ff246baa96a4049160fba62b4fe491833001dbR41) albeit in a slightly different context).

lluiscampos (commented):
Here too please add a `Changelog` before merging

tests/e2e_tests/integration/05-deviceDetails.spec.ts
```diff
@@ -33,6 +33,9 @@ test.describe('Device details', () => {
     await page.click(`text=/inventory/i`);
     expect(await page.isVisible(`css=.expandedDevice >> text=Linux`)).toBeTruthy();
     expect(await page.isVisible(`css=.expandedDevice >> text=mac`)).toBeTruthy();
+    const expandedDevice = page.locator(`css=.expandedDevice`);
+    await expandedDevice.getByRole('tab', { name: /software/i }).click();
+    await page.click(`text=/software/i`);
```
mzedel: purely informative: the `page.click(`text=/software/i`);` does the same as the two lines right above (just with the risk of having a clickable text with `/software/i` in it) - but it shouldn't hurt the test here 🤞

mzedel (approved):

tests/e2e_tests/integration/05-deviceDetails.spec.ts
alfrunes: Hmm :thinking: It seems like (from the previous run) that the
```
await page.click(`text=/software/i`);
```
just `await`s indefinitely. I forgot to remove it when I was supposed to replace it with the other two lines (that I stole from your commit :disguised_face: )

mzedel (approved):

